PR #2, opened 2025-07-10
change name mangling

braulov:
Added a companion object to MangledName for registering used names. In ViperPoweredDeclarationChecker, implemented registration of all program names after creation using the extension function Program.registerAllNames()


jesyspa (changes requested):
Nice job getting up to speed so quickly!

PR #10, opened 2025-08-12
Detect partially shared arguments passed as unique

WeetHet:
Before, trying to pass `a` when `a.x` was passed as shared would succeed, even though the function would treat `a.x` as unique which is incorrect


formver.compiler-plugin/testData/diagnostics/uniqueness_checker/partially_shared.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.jetbrains.kotlin.formver.plugin.NeverVerify
+import org.jetbrains.kotlin.formver.plugin.Unique
+
+class B
+
+class A(
+    @Unique val data: B,
+)
+
+fun takesA(@Unique a: A) {}
```
jesyspa, line 13: Consistency: `consumeA`

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/ContextTrie.kt
```diff
@@ -3,14 +3,17 @@ package org.jetbrains.kotlin.formver.uniqueness
 import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
 
 class ContextTrie(
-    val parent: ContextTrie?, val children: MutableMap<FirBasedSymbol<*>, ContextTrie>, override var level: UniqueLevel
+    val parent: ContextTrie?,
+    val symbol: FirBasedSymbol<*>?,
```
jesyspa, line 7: I think it's a hack to make these optional; I'd rather see an `EmptyContextTrie` object that implements the interface.

jesyspa (approved):
Generally ok, but some suggestions

PR #16, opened 2025-08-18
Addition of uniqueness tests

TheDeveloper101:
Refactored uniqueness tests, marked that they are failing, added assertions so they do not fail and left comments as to why they fail


formver.compiler-plugin/testData/diagnostics/verification/uniqueness/failing-tests/binary_tree.kt
```diff
@@ -0,0 +1,21 @@
...
+import org.jetbrains.kotlin.formver.plugin.Unique
+import org.jetbrains.kotlin.formver.plugin.verify
+
+class Node(var data: Int, @Unique val left: Node?, @Unique val right: Node?)
+
+fun <!VIPER_TEXT!>get_left_val<!>(@Unique n: Node): Int? {
+    return n.left?.data
+}
+
+// these expressions should all verify - they currently do not for what i think is the same reason as the linked list
```
jesyspa, line 13: nit: please name reason explicitly, as other file may change

jesyspa (changes requested):
Could you split out the extra changes so that you only update the tests, please?
This isn't a refactor, it's an addition, by the way.

PR #17, opened 2025-08-22
Error handling

braulov:
Improve INTERNAL_ERROR diagnostics to show precise error location
Add internal.kt test demonstrating several scenarios


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -319,9 +317,16 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
 
     override fun visitProperty(property: FirProperty, data: StmtConversionContext): ExpEmbedding {
         val symbol = property.symbol
-        check(symbol.isLocal) {
-            "StmtConversionVisitor should not encounter non-local properties."
-        }
+        if (!symbol.isLocal) {
+            throw CheckException(property.source,"StmtConversionVisitor should not encounter non-local properties.")
+        }/*
```
jesyspa, line 322: What happened here?

formver.compiler-plugin/testData/diagnostics/conversion/internal.kt
```diff
@@ -0,0 +1,27 @@
...
+/*
+// Test class-related errors
+class TestClass
+fun testClassUsages() {
+    TestClass      // ResolvedQualifier
+    TestClass::class  // Class literal
+}*/
+
+// Test when expression
+fun <!VIPER_TEXT!>testWhen<!>(x: String) {
```
jesyspa, line 21: I'm confused, we shouldn't give an output here, right?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/CheckException.kt
```diff
@@ -0,0 +1,5 @@
+package org.jetbrains.kotlin.formver.core.conversion
+
+import org.jetbrains.kotlin.KtSourceElement
+
+class CheckException(val source: KtSourceElement?, override val message: String) : Exception()
```
jesyspa, line 5: What do you mean by `Check` here?

jesyspa (changes requested):
There are a lot of small issues, please clean them up.  In particular, please don't use Russian in the codebase.

formver.compiler-plugin/testData/diagnostics/conversion/internal.kt
jesyspa: That's true, but maybe putting it a different way: what are you testing here?

jesyspa: But why is this something we want to test?  We don't have an internal error here, so I think we're okay with any behaviour here

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ExactLocationException.kt
```diff
@@ -0,0 +1,5 @@
+package org.jetbrains.kotlin.formver.core.conversion
+
+import org.jetbrains.kotlin.KtSourceElement
+
+class ExactLocationException(val source: KtSourceElement?, override val message: String) : Exception()
```
jesyspa, line 5: Thinking about it some more, I'd recommend calling this `PluginInternalException`.  As it is, it sounds like this exception is thrown *because* there is an exact location, which isn't the case.  It is also not quite true, since `source` can be `null` (and besides, "exact" is a vague term).

jesyspa (changes requested):
I would like to discuss these changes over a call on Monday.

PR #19, opened 2025-09-24
Add graph based name resolver

braulov:


jesyspa (changes requested):
Not ready for review.

PR #24, opened 2025-10-06
Add infrastructure for graph resolver

braulov:
1. Rewrite logic with MangledName
- add class NameExpr - String (Lit(..)) or reference to string representation of another MangledName (SymbolVal(..)), or sequence of NameExpr (Join(..)).
- Now field mangledBaseName of MangledName have type NameExpr, mangledScope : NameScope?
2. Add GraphBasedNameResolver.
3. Add registration of domains and related objects.


jesyspa (changes requested):
Not ready for review.

PR #29, opened 2025-11-05
Introduce conversion into Viper functions

AlexFalter:
This PR introduces conversion from Kotlin functions annotated as Pure into Viper functions:

For this I introduced two new embeddings:
- PureFunctionEmbedding
- PureFunctionBodyEmbedding
which are similar to their non-pure counter-parts (i.e. FunctionEmbedding and FunctionBodyEmbedding), but embed an expression rather than a statement sequence.
I also introduced the UserFunction in the Viper AST to represent such pure user-defined functions.
Further, I introduced a function map (similar to the methods map) in the program of the ProgramConverter to store embedded functions.

Regarding conversion flow:

The distinction between pure vs. non-pure is maked in the ProgramConverter when embedding the provided function. The function is then embedded in the newly introduced pure embeddings and stored in the function map. Note, that currently the body of the function is being - regardless of input - converted into the integer literal 42.

Testing:

I have introduced a test case ensuring, that @Pure annotated functions are being converted into Viper functions rather than methods. All other test-cases not failing also shows, that non-annotated functions are not being wrongly converted

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/RichCallableEmbedding.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/RichCallableEmbedding.kt
@@ -14,7 +14,9 @@ import org.jetbrains.kotlin.formver.viper.ast.Method
  * This interface does not quite fit along any line that we usually draw, hence the unspecific name. This interface is really two things
  * rolled into one:
  * - A callable with full information about the signature.
- * - A callable which knows how to generate a method header for itself (if it needs one).
+ * - A callable which knows how to generate a method header and a function header for itself (if it needs one).
+ *
+ * TODO: The Function header will likely be removed as it makes no sense to emit a function header on its own - Consider this a placeholder
  *
  * These are not fundamentally related, but providing two interfaces would mean having to deal with diamond-shaped inheritance
  * in a number of classes that would be quite inconvenient. We thus force users to implement both at once.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -120,9 +125,23 @@ class ProgramConverter(
 
         // Note: it is important that `body` is only set after `embedUserFunction` is complete, as we need to
         // place the embedding in the map before processing the body.
-        embedUserFunction(declaration.symbol, signature).apply {
-            body = stmtCtx.convertMethodWithBody(declaration, signature, returnTarget)
+        if (declaration.symbol.isPure(session)) {
+            embedPureUserFunction(declaration.symbol, signature).apply {
+                body = PureFunctionBodyEmbedding(Exp.IntLit(42), returnTarget)
```
jesyspa: Please leave a TODO here to make it clear that this is a placeholder.  (Preferably, also make a variable that's called `placeholder` so that the extent of the placeholder is clear.)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
```diff
@@ -91,4 +94,17 @@ fun FullNamedFunctionSignature.toViperMethod(
...
+
+fun FullNamedFunctionSignature.toViperFunction(
+    body: Exp?,
+    returnVariable: VariableEmbedding,
+) = UserFunction(
+    name,
+    formalArgs.map { it.toLocalVarDecl() },
+    Type.Int,
+    getPreconditions().pureToViper(toBuiltin = true),
+    emptyList(),
```
jesyspa: Why no postconditions?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionEmbedding.kt
```diff
@@ -25,3 +31,16 @@ class UserFunctionEmbedding(private val callable: RichCallableEmbedding) : Funct
...
 }
+
+
+/**
+ * An embedding of a user-defined pure function
+ */
+class PureUserFunctionEmbedding(private val callable: RichCallableEmbedding) : PureFunctionEmbedding,
+        CallableEmbedding by callable {
+
+    var body: PureFunctionBodyEmbedding? = null
```
jesyspa: Until you need a PureFunctionBody embedding, perhaps just an `ExpEmbedding` would be best here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionEmbedding.kt
```diff
@@ -25,3 +31,16 @@ class UserFunctionEmbedding(private val callable: RichCallableEmbedding) : Funct
...
+
+    override val viperFunction: Function?
+        get() = body?.toViperFunction(callable) ?: callable.toViperFunctionHeader()
```
jesyspa: I think this is overly permissive - the idea of having a nullable bodyembedding for methods is that we only want to verify some methods.  But at the moment we want to translate all functions to ones with bodies, so that's quite a different story.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -120,9 +125,23 @@ class ProgramConverter(
 
         // Note: it is important that `body` is only set after `embedUserFunction` is complete, as we need to
         // place the embedding in the map before processing the body.
-        embedUserFunction(declaration.symbol, signature).apply {
-            body = stmtCtx.convertMethodWithBody(declaration, signature, returnTarget)
+        if (declaration.symbol.isPure(session)) {
```
jesyspa: I'm afraid this can't be the right place to do this.  Namely: this is only called for the "root" function that we are verifying, while you want to add bodies to all pure functions (those bodies are visible in methods that call them).  So you want to hook this up in a place that is called for all functions a user calls.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/NonInlineNamedFunction.kt
```diff
@@ -24,4 +25,10 @@ class NonInlineNamedFunction(val signature: FullNamedFunctionSignature) : RichCa
             null,
             PlaceholderVariableEmbedding(PlaceholderReturnVariableName, signature.callableType.returnType)
         )
+
+    override fun toViperFunctionHeader(): Function =
```
jesyspa: I don't think you need this

formver.compiler-plugin/testData/diagnostics/conversion/pure_function.kt
```diff
@@ -0,0 +1,7 @@
+import org.jetbrains.kotlin.formver.plugin.Pure
+import org.jetbrains.kotlin.formver.plugin.DumpExpEmbeddings
+
+fun <!VIPER_TEXT!>emptyFunction<!>() { }
+
+@Pure
+fun <!VIPER_TEXT!>emptyAnnotatedFunction<!>() { }
```
jesyspa: If you're going to put placeholders, I would put ones where the behaviours are as you expect them to be.  Now, the translation to Viper looks like it is doing something very strange.  (Or at least leave a comment, but given how easy it is to return 42 here, I would do that.)

jesyspa (changes requested):
This looks like the right direction, some minor comments.  The only big thing is that we need to find a different place to hook this into - give it a shot (add a test where you call the pure function and make sure it is generated), and if you can't find one we can call and look at it together).

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/FunctionBodyEmbedding.kt
```diff
@@ -20,3 +23,12 @@ data class FunctionBodyEmbedding(
     fun toViperMethod(signature: FullNamedFunctionSignature): Method =
         signature.toViperMethod(viperBody, returnTarget.variable)
 }
+
+data class PureFunctionBodyEmbedding(
```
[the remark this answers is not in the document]
jesyspa: I don't think you need that in a function? The return target is the variable to assign to in order to return a value, but IIRC in functions that is not allowed, the return value is just the result of the body expression.

AlexFalter:
Latest push introduced:

- FunctionCall embedding to represent a call to a function
- Added TODO markers
- Changed functions to be translated with return type Ref (see Slack for more details)
- Therefore functions now return null always (isntead of 42)
- Represent function call as an assign statement

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -281,18 +283,40 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
 
         when (val forAllLambda = functionCall.extractFormverFirBlock { isInvariantBuilderFunctionNamed("forAll") }) {
             null -> {
-                val callee = data.embedFunction(symbol)
-                return callee.insertCall(
-                    functionCall.functionCallArguments.withVarargsHandled(data, callee),
-                    data,
-                    data.embedType(functionCall.resolvedType),
-                )
+                if (data.isPureFunction(symbol)) {
```
jesyspa: I think this separation of responsibilities is not ideal.  The StmtConversionVisitor should not care whether the function it embeds is pure or not; it just needs to get a callable object from the ProgramConverter, and then put a call in it.  Deciding how to embed that function shouldn't be up to it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/NonInlineNamedFunction.kt
```diff
@@ -7,21 +7,34 @@ package org.jetbrains.kotlin.formver.core.embeddings.callables
...
 import org.jetbrains.kotlin.formver.viper.ast.Method
 
 class NonInlineNamedFunction(val signature: FullNamedFunctionSignature) : RichCallableEmbedding,
     FullNamedFunctionSignature by signature {
     override fun insertCall(
         args: List<ExpEmbedding>,
         ctx: StmtConversionContext,
-    ): ExpEmbedding = MethodCall(signature, args)
+    ): ExpEmbedding {
+        return if (ctx.isPureFunction(signature.symbol)) {
```
jesyspa: I think we should already be aware of this earlier, probably when we create this `NonInlineNamedFunction`, or perhaps even when we create the `signature`

formver.compiler-plugin/testData/diagnostics/conversion/pure_function.fir.diag.txt
```diff
@@ -0,0 +1,18 @@
+/pure_function.kt:(53,66): info: Generated Viper text for emptyFunction:
+function f$emptyAnnotatedFunction$TF$(): Ref
```
jesyspa: Something is going wrong here: you want to include the body.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -281,18 +283,40 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
             }
         }
 
+    private fun List<FirExpression>.withPureVarargsHandled(
```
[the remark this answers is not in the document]
AlexFalter: I thought of this as well, however with the viperMethod being a member of the Function embedding it feels strange to have PureFunctionEmbedding being a subclass of FunctionEmbedding

formver.compiler-plugin/testData/diagnostics/conversion/pure_function.fir.diag.txt
AlexFalter: This is just the header of the function called in the root function we are translating. If I understand correctly it is done for methods as well (i. e. when called only the header is being generated rather than the whole method again), so I kept this behaviour for now. It looks strange due to missing Pre/Post-Conditions...

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Function.kt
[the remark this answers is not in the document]
AlexFalter: I agree - however I think I have dug the hole to deep for me to untangle this PR before merge. I have understimated the amount of code changes I need to do for this a bit. So (assuming I get the rest above to work) if alright with you I would keep it in this PR and then be a more considerate of this with the next features

formver.compiler-plugin/testData/diagnostics/conversion/pure_function.fir.diag.txt
jesyspa: That is correct; for methods, we do not generate their bodies, since those are opaque and they are verified separately.  However, this isn't the case for functions: there, the bodies are important since Viper uses them, so a function should include its body.

That said, this can be left as a TODO for now

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -178,6 +195,29 @@ class ProgramConverter(
...
+                PureUserFunctionEmbedding(callable).also {
+                    functions[lookupName] = it
+                }
+            }
+
+            else -> existing
+        }
+    }
+
+    override fun embedAnyFunction(symbol: FirFunctionSymbol<*>): CallableEmbedding {
```
jesyspa: nit: instead of

```kotlin
fun f() {
  return ...
}
```

consider 

```kotlin
fun f() = ...
```

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
```diff
@@ -222,6 +223,23 @@ data class MethodCall(val method: NamedFunctionSignature, val args: List<ExpEmbe
...
+        get() = args
+
+    override fun toViper(ctx: LinearizationContext): Exp = function.toFuncApp(
+        args.map { it.toViper(ctx) },
+        ctx.source.asPosition
+    )
+
+    override fun <R> accept(v: ExpVisitor<R>): R =
+        v.visitFunctionCall(this) // or however your visitor is set up
+
```
jesyspa: nit: extra newline

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Function.kt
```diff
@@ -67,6 +67,21 @@ interface Function : IntoSilver<viper.silver.ast.Function>, Applicable {
...
+    override val pres: List<Exp>,
+    override val posts: List<Exp>,
+    override val body: Exp?,
+    override val pos: Position = Position.NoPosition,
+    override val info: Info = Info.NoInfo,
+    override val trafos: Trafos = Trafos.NoTrafos,
+) : Function {
+    override val includeInDumpPolicy: IncludeInDumpPolicy = IncludeInDumpPolicy.ALWAYS
+}
+
```
jesyspa: nit: extra newline

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
```diff
@@ -91,4 +90,17 @@ fun FullNamedFunctionSignature.toViperMethod(
...
-)
\ No newline at end of file
+)
+
+fun FullNamedFunctionSignature.toViperFunction(
+    body: Exp?,
+) = UserFunction(
+    name,
+    formalArgs.map { it.toLocalVarDecl() },
+    // TODO: Be explicit about the return types of functions instead of boxing them into a Ref
```
jesyspa: This is optimistic at this stage; `Ref` is our standard type to pass things around, so this TODO would have to be everywhere if we wanted to mark all the places.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
```diff
@@ -222,6 +223,23 @@ data class MethodCall(val method: NamedFunctionSignature, val args: List<ExpEmbe
...
+    override val subexpressions: List<ExpEmbedding>
+        get() = args
+
+    override fun toViper(ctx: LinearizationContext): Exp = function.toFuncApp(
+        args.map { it.toViper(ctx) },
+        ctx.source.asPosition
+    )
+
+    override fun <R> accept(v: ExpVisitor<R>): R =
+        v.visitFunctionCall(this) // or however your visitor is set up
```
jesyspa: I don't follow, what does this comment mean?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
jesyspa: Unfortunately no, and it's not always an error - Kotlin doesn't have a strict policy, but it's good to be consistent and to keep the code reasonably compact (as it tends to be quite lengthy already)

PR #31, opened 2025-11-18
Basic body translation of pure functions

AlexFalter:
This PR introduces basic body translation of pure functions. Mainly the following was done:

- Introduce PureBlock representing a block of pure expressions (i. e. the body of a pure function), which is then translated into a sequence of let-bindings to create a single viper expression
- StmtPureConversionVisitor, which traverses the Fir representation and translates into this newly introduced embedding
- convertFunctionWithBody function in the StmtConversionCtxt, which is used to convert the function making use of the PureLinearizer (rather than the default one)
- Test cases representing the current state of the conversion:

Currently only a sequence of variable assignments followed by a literal- or variable return can be translated.

**NOTE**:  We are currently not verifying the correctness of the purity annotation with this file. The PurityChecker is still to be hooked into this workflow

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtPureConversionVisitor.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtPureConversionVisitor.kt
@@ -16,7 +16,8 @@ import org.jetbrains.kotlin.text
 import org.jetbrains.kotlin.types.ConstantValueKind
 
 /**
- * A restrictive version of the StmtConversionVisitor, which is used to embed pure functions
+ * A restrictive version of the StmtConversionVisitor, which is used to embed functions annotated as pure
+ * into an ExpEmbedding representation that reflects their pure aspects better and makes linearization easier
  */
 object StmtPureConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>() {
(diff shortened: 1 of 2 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtPureConversionVisitor.kt
```diff
@@ -0,0 +1,92 @@
...
+            "Pure conversion not yet implemented for $element (${element.source.text}) - Consider removing the @Pure annotation of this function",
+            data
+        )
+
+    override fun visitReturnExpression(
+        returnExpression: FirReturnExpression,
+        data: StmtConversionContext,
+    ): ExpEmbedding {
+        val expr = when (returnExpression.result) {
+            is FirUnitExpression -> NullLit
```
jesyspa: This is wrong: there is (or should be) an object of type `Unit`.  If it's hard to get to work, throw an exception.

jesyspa (changes requested):
I think forking the StmtConversionVisitor isn't a viable solution, we should discuss what the current StmtConversionVisitor doesn't get right.

AlexFalter:
I think this PR is so far off that it is not worth keeping this open/adjusting. I will open a new PR for a different approach

PR #35, opened 2025-11-23
Hook in purity checker

AlexFalter:
This PR hooks in the existing purity checking into the translation workflow of pure functions to validate the pureness of their body. Namely the util function isPure() was introduced in ExprUtils to run the purity check on a provided embedding. I introduced a test-case to validate that there are no false negatives (the passing pure_function tests validate, that the purity checker has no false positives)

Changes to the existing workflow:

- Non-Pure function bodies, which have impure asserts, now are not being translated (i. e. an empty body is returned and the translation results in a header - see test-case)
- For clarity's sake I renamed .isValid() -> .satisfiesPurityRequirements() and .checkValidity() -> allNodesSatisfyPurityRequirements() - UPDATE: Dropped this point

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
@@ -254,10 +254,8 @@ fun StmtConversionContext.convertMethodWithBody(
     // as we may not encounter any `return` statement in the body
     returnTarget.variable.withIsUnitInvariantIfUnit().toViperUnusedResult(linearizer)
 
-    if (!body.checkValidity(declaration.source, errorCollector)) throw SnaktInternalException(
-        declaration.source,
-        "Function body is not valid"
-    )
+    // TODO: Stop translation if this check fails
+    body.checkValidity(declaration.source, errorCollector)
 
     return FunctionBodyEmbedding(seqnBuilder.block, returnTarget, bodyExp)
 }
--- a/formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
+++ b/formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
@@ -75,6 +75,9 @@ class ViperPoweredDeclarationChecker(private val session: FirSession, private va
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -253,7 +254,8 @@ fun StmtConversionContext.convertMethodWithBody(
     // as we may not encounter any `return` statement in the body
     returnTarget.variable.withIsUnitInvariantIfUnit().toViperUnusedResult(linearizer)
 
-    body.checkValidity(declaration.source, errorCollector)
+    if (!body.allNodesSatisfyPurityRequirements(declaration.source, errorCollector)) return null
```
jesyspa: This is not a reasonable implementation; if we consider this an error, we should stop conversion, not just generate a dummy.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -253,7 +254,11 @@ fun StmtConversionContext.convertMethodWithBody(
     // as we may not encounter any `return` statement in the body
     returnTarget.variable.withIsUnitInvariantIfUnit().toViperUnusedResult(linearizer)
 
-    body.checkValidity(declaration.source, errorCollector)
+    if (!body.checkValidity(declaration.source, errorCollector)) throw SnaktInternalException(
```
jesyspa: This is not an internal error, since it is caused by the user. 
I don't think this part needs to be changed at all for this PR, maybe best to just leave it as is?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -269,7 +272,10 @@ fun StmtConversionContext.convertFunctionWithBody(
     val body = extractReturnedExprFromPureFunctionBody(bodyWithPosition, declaration.source)
     val pureLinearizer = PureLinearizer(declaration.source)
 
-    // TODO: Hook in purity check here
+    if (!body.isPure()) throw SnaktInternalException(
```
jesyspa: Worth noting: this is also not an internal exception, it's a user error.  But it's fine to leave that as a TODO.

jesyspa (approved):
No action needed, comment FYI

PR #37, opened 2025-11-29
Delegate assignment to lineraizer

AlexFalter:
This PR delegates the translation work for assignments to the linearizer. This will later be used to emit Let-Expressions, rather than adding statements. Therefore, the Assign embedding returns an Expression and was converted to a "OnlyToViperExpEmbedding". As the tests are still running this transformation has not changed the translation behaviour

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
@@ -519,17 +519,12 @@ data class PredicateAccessPermissions(
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitPredicateAccessPermissions(this)
 }
 
-data class Assign(val lhs: ExpEmbedding, val rhs: ExpEmbedding) : UnitResultExpEmbedding {
+data class Assign(val lhs: ExpEmbedding, val rhs: ExpEmbedding) : OnlyToViperExpEmbedding {
     override val type: TypeEmbedding = lhs.type
 
-    override fun toViperSideEffects(ctx: LinearizationContext) {
-        val lhsViper = lhs.toViper(ctx)
-        if (lhsViper is Exp.LocalVar) {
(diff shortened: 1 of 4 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -519,17 +519,12 @@ data class PredicateAccessPermissions(
...
-        val lhsViper = lhs.toViper(ctx)
-        if (lhsViper is Exp.LocalVar) {
-            rhs.withType(lhs.type).toViperStoringIn(LinearizationVariableEmbedding(lhsViper.name, lhs.type), ctx)
-        } else {
-            val rhsViper = rhs.withType(lhs.type).toViper(ctx)
-            ctx.addStatement { Stmt.assign(lhsViper, rhsViper, ctx.source.asPosition) }
-        }
+    override fun toViper(ctx: LinearizationContext): Exp {
+        val result = ctx.emitAssignment(lhs, rhs)
+        return result ?: RuntimeTypeDomain.unitValue(pos = ctx.source.asPosition)
```
jesyspa: I'm a bit confused; the point of the `toViperSideEffects` function is that this type of expembedding doesn't give a meaningful expression as an expression.  Why get rid of that and always return `unit`?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationContext.kt
```diff
@@ -44,6 +46,8 @@ interface LinearizationContext {
     fun addDeclaration(decl: Declaration)
 
     fun addModifier(mod: StmtModifier)
+
+    fun emitAssignment(lhs: ExpEmbedding, rhs: ExpEmbedding): Exp?
```
jesyspa: To conform to the general pattern, please use `add` instead of `emit.

Also, what is the interpretation of the returned `Exp`?

jesyspa (changes requested):

PR #39, opened 2025-12-02
Alex/introduce let bindings

AlexFalter:
This PR introduces conversion of local declarations to a chain of let-bindings. I introduced the following:

- Let bindings in the Viper AST representation
- Declare embeddings now also call the linearizer to add an assignment if appropriate
- Declare embeddings are now pure, if they have a non-empty initializer
- A LetChainBuilder component keeping track of assignments and providing a "let-chain" of all tracked assignments
- Pure function bodies may now contain a block of declarations followed by a return statment
- return statements can only contain literals or a single variable embedding (to support builtin functions like the arithmetic functions I will open a separate PR)

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
@@ -6,7 +6,6 @@
 package org.jetbrains.kotlin.formver.core.linearization
 
 import org.jetbrains.kotlin.KtSourceElement
-import org.jetbrains.kotlin.formver.common.SnaktInternalException
 import org.jetbrains.kotlin.formver.core.asPosition
 import org.jetbrains.kotlin.formver.core.conversion.ReturnTarget
 import org.jetbrains.kotlin.formver.core.embeddings.expression.AnonymousVariableEmbedding
@@ -82,18 +81,6 @@ data class Linearizer(
         returnExp.withType(target.variable.type)
             .toViperStoringIn(target.variable, this)
         addStatement { target.label.toLink().toViperGoto(this) }
-    }
-
-    private fun convertReturnTargetToLinearizationVariableEmbedding(
-        target: ReturnTarget,
-        type: TypeEmbedding
-    ): LinearizationVariableEmbedding {
-        val retVarViper = target.variable.toViper(this)
-        if (retVarViper !is Exp.LocalVar) throw SnaktInternalException(
-            source,
-            "Translated return variable of function must be a local variable. Got: $retVarViper"
-        )
-        return LinearizationVariableEmbedding(retVarViper.name, type)
     }
 
     override fun addModifier(mod: StmtModifier) {


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -269,29 +266,39 @@ fun StmtConversionContext.convertFunctionWithBody(
     )
     val bodyWithPosition = convert(firBody)
     // TODO: Clean this up
-    val body = extractReturnedExprFromPureFunctionBody(bodyWithPosition, declaration.source)
-    val pureLinearizer = PureLinearizer(declaration.source)
+    val body = extractRelevantExpsFromPureFunctionBody(bodyWithPosition, declaration.source)
+    val letChainBuilder = LetChainBuilder(declaration.source)
+    val pureLinearizer = PureLinearizer(declaration.source, letChainBuilder)
```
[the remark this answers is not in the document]
AlexFalter: That indeed doesn't really make sense. I took inspiration from the SeqnBuilder above (in convertMethodWithBody). Just out of curiosity, is there a reason for this state to be external? Regardless, I'll pull it into the linearizer

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LetChainBuilder.kt
```diff
@@ -0,0 +1,34 @@
+package org.jetbrains.kotlin.formver.core.linearization
+
+import org.jetbrains.kotlin.KtSourceElement
+import org.jetbrains.kotlin.formver.common.SnaktInternalException
+import org.jetbrains.kotlin.formver.viper.ast.Declaration
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+
+class LetChainBuilder(val source: KtSourceElement? = null) : LetChainBuildContext {
+    private val assignments: MutableMap<Declaration.LocalVarDecl, Exp> = mutableMapOf()
```
jesyspa: I would use a list, as later assignments may depend on earlier ones.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureLinearizer.kt
```diff
@@ -46,17 +51,30 @@ class PureLinearizer(override val source: KtSourceElement?) : LinearizationConte
...
-    override fun addDeclaration(decl: Declaration) {
-        throw PureLinearizerMisuseException("addDeclaration")
-    }
+    // Nothing to do here, as an assignment is also added
+    override fun addDeclaration(decl: Declaration) {}
 
     override fun addAssignment(lhs: ExpEmbedding, rhs: ExpEmbedding) {
-        throw PureLinearizerMisuseException("addAssignment")
+        // TODO: Move this to purity check
+        if (lhs.ignoringMetaNodes() !is VariableEmbedding) throw SnaktInternalException(
```
jesyspa: The frequency with which you have to ignore meta nodes is a bit worrying.  Is there not a neater way of working with this?

jesyspa (changes requested):
Mostly looks good, some minor comments

jesyspa (approved):
Looks good, one question

PR #41, opened 2025-12-25
Add functionality for programmers to specify field access permissions

TheDeveloper101:


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.core.linearization.LinearizationContext
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+
+class AccEmbedding(
+    val field: Exp.FieldAccess,
+    val perm: PermExp,
+) : OnlyToBuiltinTypeExpEmbedding {
+
```
jesyspa: nit: extra line

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.jetbrains.kotlin.formver.core.embeddings.ExpVisitor
+import org.jetbrains.kotlin.formver.core.embeddings.asInfo
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.core.linearization.LinearizationContext
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+
+class AccEmbedding(
+    val field: Exp.FieldAccess,
+    val perm: PermExp,
```
jesyspa: Is this an enum?  If not, perhaps having one on our side would be good.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -0,0 +1,36 @@
...
+            info = sourceRole.asInfo,
+        )
+    }
+
+    override val subexpressions: List<ExpEmbedding> = TODO()
+
+    override val type: TypeEmbedding
+        get() = TODO()
+
+    override fun <R> accept(v: ExpVisitor<R>): R = TODO()
```
jesyspa: Modify the visitor to handle this please.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -0,0 +1,36 @@
...
+            perm = perm,
+            pos = ctx.source.asPosition,
+            info = sourceRole.asInfo,
+        )
+    }
+
+    override val subexpressions: List<ExpEmbedding> = TODO()
+
+    override val type: TypeEmbedding
+        get() = TODO()
```
jesyspa: Same thing as what forall does should be fine here.

jesyspa (changes requested):
Generally looks good, only issue is how to represent the field access best

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.jetbrains.kotlin.formver.core.asPosition
+import org.jetbrains.kotlin.formver.core.embeddings.ExpVisitor
+import org.jetbrains.kotlin.formver.core.embeddings.asInfo
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.core.linearization.LinearizationContext
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+
+class AccEmbedding(
+    val field: Exp.FieldAccess,
```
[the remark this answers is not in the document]
TheDeveloper101: I had this exact thought at first, but then I changed my mind as I figured we could translate to Viper ahead of using the embedding. But I think this is a cleaner and easier approach since it also makes the embedding easier to use.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
TheDeveloper101: No worries! I did create an enum to explicitly mention the permissions. Also, I added an extra data class to `PermExp` which is a wrapper around the `EpsilonPerm` to represent having read permissions

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
jesyspa: Sorry, to clarify: you had an empty line that could be removed (and now you do again here)

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Exp.kt
```diff
@@ -565,6 +565,27 @@ sealed interface Exp : IntoSilver<viper.silver.ast.Exp> {
...
     }
 
+    data class Acc(
+        val field: FieldAccess,
+        val perm: PermExp,
+        val pos: Position = Position.NoPosition,
+        val info: Info = Info.NoInfo,
+        val trafos: Trafos = Trafos.NoTrafos,
+    ) : Exp {
+        override val type = field.type
```
jesyspa: Is this true?  I think it should always be a boolean, since the access permission is a boolean.  (Or did we do something different for forall?)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/debug/Utils.kt
```diff
@@ -33,6 +33,7 @@ val PermExp.debugTreeView: TreeView
     get() = when (this) {
         is PermExp.WildcardPerm -> PlaintextLeaf("wildcard")
         is PermExp.FullPerm -> PlaintextLeaf("write")
+        is PermExp.EpsilonPerm -> PlaintextLeaf("read")
```
jesyspa: I don't think this is needed, wildcard does the same thing.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Exp.kt
TheDeveloper101: Right that makes sense, for `Forall` it was also a boolean, but I had modeled this after the field access

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/debug/Utils.kt
TheDeveloper101: I see, makes sense. I just wasn't sure how exactly wildcard permissions worked which is why I added the epsilon permission, but I removed it now

TheDeveloper101:
Sounds good :) yeah I can work on a separate PR to figure out how to create the embeddings and hook this up to the parser etc...

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.jetbrains.kotlin.formver.core.embeddings.types.FieldAccessTypeInvariantEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.types.buildType
+import org.jetbrains.kotlin.formver.core.linearization.LinearizationContext
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+
+enum class PermOption{
+    READ, WRITE
+}
```
jesyspa: Sorry, looking at the code again - I see we use PermExp pretty ubiquitously directly.  I think let's do that here as well.  (Maybe it's worth changing everywhere, I opened an issue about investigating this: https://github.com/jesyspa/SnaKt/issues/42)

jesyspa (changes requested):
Some minor stuff, generally look good.

jesyspa (approved):

PR #43, opened 2025-12-27
Introduce SSA-Block conversion

AlexFalter:
This PR introduces SSA conversion for blocks:

- The SSAConverter now holds a list of triples each containing a variable declaration, an SSA-Index and a defining expression for that declaration at this index.
- The SSAConverter holds a map from variable declarations to current ssa index of a declaration
- The linearizer resolves the actual name of a variable in Viper

**Note:** This PR does not yet support scope functions (constructs with run {} f. e.) as they are translated to a Function embedding that needs separate handling

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
@@ -36,10 +36,10 @@ class SsaConverter(
 
     /**
      * Resolves the symbolic name to its most recent SSA definition.
-     * If no local assignment is found, we assume the name refers to a function parameter
+     * If no local assignment is found, we assume the provided name is valid
      */
     fun resolveVariableName(name: SymbolicName): SymbolicName =
-        // TODO: Fall back to global value numbering before falling back to parameter
+        // TODO: Fall back to global value numbering before falling back to provided name
         assignments.lastOrNull { it.declaration.name == name }?.let { SsaVariableName(name, it.ssaIdx) } ?: name
 }
 


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -2,33 +2,39 @@ package org.jetbrains.kotlin.formver.core.linearization
...
 import org.jetbrains.kotlin.formver.viper.ast.Declaration
 import org.jetbrains.kotlin.formver.viper.ast.Exp
 
-class SsaConverter(val source: KtSourceElement? = null) {
-    private val assignments: MutableMap<Declaration.LocalVarDecl, Exp> = mutableMapOf()
+class SsaConverter(
+    val source: KtSourceElement? = null,
+    val variableIndex: MutableMap<SymbolicName, Int> = mutableMapOf()
+) {
+    private val assignments: MutableList<Triple<Declaration.LocalVarDecl, Int, Exp>> = mutableListOf()
```
jesyspa: I think a dedicated data class would be clearer than a triple.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -2,33 +2,39 @@ package org.jetbrains.kotlin.formver.core.linearization
...
     }
 
     fun addBody(body: Exp) {
         this.body = body
     }
+
+    fun resolveVariableName(name: SymbolicName): SymbolicName =
+    // Fall back to parameter if no assignment in map is found
+        // TODO: Fall back to global value numbering before falling back to parameter
+        assignments.lastOrNull { it.first.name == name }?.let { SsaVariableName(name, it.second) } ?: name
```
jesyspa: Could you be more specific about why this is the desired behaviour?  And why not use `variableIndex`?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -2,33 +2,39 @@ package org.jetbrains.kotlin.formver.core.linearization
...
 
     fun addAssignment(decl: Declaration.LocalVarDecl, varExp: Exp) {
-        // TODO: Allow this case using SSA transformation
-        if (assignments.containsKey(decl)) throw SnaktInternalException(
-            source,
-            "Found duplicate variable declaration while constructing a let-chain. Declaration is $decl"
-        )
-        assignments[decl] = varExp
+        variableIndex[decl.name] = variableIndex[decl.name]?.plus(1) ?: 0
+        assignments.add(Triple(decl, variableIndex[decl.name] as Int, varExp))
```
jesyspa: Please avoid the cast (e.g. store the value in a local).

jesyspa (commented):
Generally looks good, just some nits.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -2,33 +2,49 @@ package org.jetbrains.kotlin.formver.core.linearization
...
 
     fun addAssignment(decl: Declaration.LocalVarDecl, varExp: Exp) {
-        // TODO: Allow this case using SSA transformation
-        if (assignments.containsKey(decl)) throw SnaktInternalException(
-            source,
-            "Found duplicate variable declaration while constructing a let-chain. Declaration is $decl"
-        )
-        assignments[decl] = varExp
+        variableIndex[decl.name] = variableIndex[decl.name]?.plus(1) ?: 0
+        assignments.add(SsaAssignment(decl, variableIndex[decl.name] as Int, varExp))
```
jesyspa: Please avoid the cast here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -2,33 +2,49 @@ package org.jetbrains.kotlin.formver.core.linearization
...
 
     fun addBody(body: Exp) {
         this.body = body
     }
-}
\ No newline at end of file
+
+    /**
+     * Resolves a base variable name to the latest SSA variable name used in this or previous converters.
+     * The name is resolved as follows:
```
jesyspa: In general, the point of a comment is to explain why things are the way they are; what happens should be clear from the code.

jesyspa (changes requested):

jesyspa (approved):
Still some issues but should be fine once those are fixed.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -2,33 +2,44 @@ package org.jetbrains.kotlin.formver.core.linearization
...
 
     fun addBody(body: Exp) {
         this.body = body
     }
-}
\ No newline at end of file
+
+    /**
+     * Resolves the symbolic name to its most recent SSA definition.
+     * If no local assignment is found, we assume the name refers to a function parameter
```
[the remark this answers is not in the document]
AlexFalter: It is fine to implicitly assume this here as a fallback. If not correctly assumed the consistency check will catch this (and the embedding indeed would not be consistent)

PR #45, opened 2026-01-19
Introduce global value numbering for SSA conversion

AlexFalter:
This PR introduces the global value numbering algorithm for the existing SSA conversion.

As this is a quite big change, for all details please refer to: https://github.com/jesyspa/snakt-design/pull/4

In "short" the following is being done:
- When traversing an if embedding, a relationship between SSAConverters is established similar to the one of an AST
- Each SSAConverter holds a local and global branching condition, where the local condition holds a condition under which this SSAConverter is being traversed from their parent, and the global branching condition is the conjunction of these local branching conditions up to the top-level
- Assignments are being tracked locally in the scope of their AST block
- If for a variable usage no definition is found in local scope, it is being resolved globally
- If multiple global definitions are found, a new definition holding a ternary expression resolving to the correct definition based on the local branching condition of all the reaching definitions is inserted into the scope of this SSAConverter
- After traversal of the whole function body an overall return expression is constructed from all SSAConverters holding a return expression and their global branching condition as well as a let chain holding all assignments in dependency ordering

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
@@ -13,11 +13,11 @@ import org.jetbrains.kotlin.formver.viper.ast.Type
 class SsaConverter(
     val source: KtSourceElement? = null,
 ) {
-    private var head: SsaBlockNode = SsaBlockNode(SsaStartNode(), Exp.BoolLit(true), Exp.BoolLit(true))
+    private var head: SsaBlockNode = SsaBlockNode(SsaStartNode(), Exp.BoolLit(true))
     private val ssaAssignments: MutableList<Pair<SsaVariableName, Exp>> = mutableListOf()
     private val returnExpressions: MutableList<Pair<Exp, Exp>> = mutableListOf()
 
-    // Shared between nodes: Produce new ssa names
+    // Produce new ssa names for a source variable name
     private val ssaNameProducers: MutableMap<SymbolicName, FreshEntityProducer<SsaVariableName, SymbolicName>> =
         mutableMapOf()
 
@@ -35,9 +35,10 @@ class SsaConverter(
         val joinNode = SsaJoinNode(
             thenResultHead,
             head,
+            condition,
             this
         )
-        head = SsaBlockNode(joinNode, splitPoint.mostRecentBranchingCondition, splitPoint.fullBranchingCondition)
+        head = SsaBlockNode(joinNode, splitPoint.fullBranchingCondition)
     }
 
     fun constructExpression(): Exp {
@@ -64,7 +65,7 @@ class SsaConverter(
     }
 
     fun addAssignment(name: SymbolicName, varExp: Exp) {
-        val ssaName = head.updateLatestNameUsage(name)
+        val ssaName = head.updateLatestName(name)
         ssaAssignments.add(ssaName to varExp)
     }
 
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
@@ -9,41 +9,33 @@ import org.jetbrains.kotlin.formver.viper.ast.Exp
  * A node in a SSA-Graph
  */
 sealed interface SsaNode {
-
     /**
-     * Function used to resolve names to their SSAVariableName
+     * Function resolves source names to their SSAVariableName
      * Fallsback to provided name if no such name is found
      */
     fun resolveVariableName(name: SymbolicName): SymbolicName
 }
 
-class SsaStartNode() :
-    SsaNode {
+class SsaStartNode() : SsaNode {
     override fun resolveVariableName(name: SymbolicName): SymbolicName =
         name
 }
 
 class SsaBlockNode(
     private val predecessor: SsaNode,
-    val mostRecentBranchingCondition: Exp,
     val fullBranchingCondition: Exp
 ) : SsaNode {
-
     val latestName: MutableMap<SymbolicName, SsaVariableName> = mutableMapOf()
 
     fun generateBranchingBlockNodeFromThisNode(condition: Exp): SsaBlockNode =
         SsaBlockNode(
             this,
-            condition,
             if (fullBranchingCondition == Exp.BoolLit(true)) condition else Exp.And(condition, fullBranchingCondition),
         )
 
     context(ssaConverter: SsaConverter)
-    fun updateLatestNameUsage(name: SymbolicName): SsaVariableName {
-        val ssaName = ssaConverter.generateFreshSsaName(name);
-        latestName[name] = ssaName;
-        return ssaName
-    }
+    fun updateLatestName(name: SymbolicName): SsaVariableName =
+        ssaConverter.generateFreshSsaName(name).also { latestName[name] = it }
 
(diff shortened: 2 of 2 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureLinearizer.kt
```diff
@@ -68,6 +72,81 @@ data class PureLinearizer(
...
     }
 
+    override fun addBranch(
+        condition: ExpEmbedding,
+        thenBranch: ExpEmbedding,
+        elseBranch: ExpEmbedding,
+        type: TypeEmbedding,
+        result: VariableEmbedding?
+    ) {
+        val startNode = this.ssaConverter
```
jesyspa: Explicit `this` is not very common in Kotlin, any particular reason you use it here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureLinearizer.kt
```diff
@@ -29,9 +31,11 @@ class PureLinearizerMisuseException(val offendingFunction: String) : IllegalStat
  */
 data class PureLinearizer(
     override val source: KtSourceElement?,
-    val ssaConverter: SsaConverter
+    private val ssaConverterReference: SsaConverterReference = SsaConverterReference(SsaConverter())
```
jesyspa: Should this be left for the users to deal with?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -6,22 +6,86 @@ import org.jetbrains.kotlin.formver.core.names.SsaVariableName
...
+            !visited.add(current) -> false
+            else -> current.predecessors.all { pred ->
+                checkUpstream(pred, visited)
+            }
+        }
+
+    /**
+     * Returns a topological ordering of SSAConverters starting from this SSAConverter
+     */
+    private fun getTopologicallySortedBlocks(): List<SsaConverter> {
```
jesyspa: Is this really necessary?  It seems to me like you can build a topological order of SSAConverters from the order you create them in.  (Alternatively, it can be obtained on-demand when you ask the converter for its body.)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -6,22 +6,86 @@ import org.jetbrains.kotlin.formver.core.names.SsaVariableName
 import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Declaration
 import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.Exp.Companion.toDisjunction
+import org.jetbrains.kotlin.formver.viper.ast.Exp.TernaryExp
+import org.jetbrains.kotlin.formver.viper.ast.Type
+import org.jetbrains.kotlin.formver.viper.debugMangled
 
 class SsaConverter(
```
jesyspa: This is a class with a lot of mutable state, that looks like it's trying to be both a graph as a whole and a node within the graph.  I would recommend looking into how you could avoid mutable state as much as reasonable (especially non-monotonic mutable state, like what you do with predecessors/successors right now), and perhaps splitting the graph/node responsibilities.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureLinearizer.kt
```diff
@@ -59,13 +60,36 @@ data class PureLinearizer(
...
             rhs.toViper(this)
         )
     }
 
     override fun addReturn(returnExp: ExpEmbedding, target: ReturnTarget) {
-        ssaConverter.addBody(returnExp.toViper(this))
+        ssaConverter.addReturn(returnExp.toViper(this))
+    }
+
+    override fun addBranch(
```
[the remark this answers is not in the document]
jesyspa: nit: Please add a TODO here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
```diff
@@ -0,0 +1,111 @@
...
+    override val fullBranchingCondition: Exp
+) : SsaNode {
+
+    override val ssaNameProducer = predecessor.ssaNameProducer
+    val latestName: MutableMap<SymbolicName, SsaVariableName> = mutableMapOf()
+
+    override fun resolveVariableName(name: SymbolicName): SymbolicName =
+        latestName[name] ?: predecessor.resolveVariableName(name)
+
+    fun registerAndUpdateLatestNameUsage(name: SymbolicName): SsaVariableName {
```
jesyspa: I don't understand what the distinction between register and update are in this function name

jesyspa (changes requested):
The large-scale picture looks good to me, some more comments on style

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
```diff
@@ -0,0 +1,83 @@
...
+
+class SsaStartNode() :
+    SsaNode {
+    override fun resolveVariableName(name: SymbolicName): SymbolicName =
+        name
+}
+
+class SsaBlockNode(
+    private val predecessor: SsaNode,
+    val mostRecentBranchingCondition: Exp,
```
jesyspa: Is this still needed?  It seems like it would fit better in the SsaJoinNode, since that's where it's used.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
```diff
@@ -0,0 +1,83 @@
...
+            this,
+            condition,
+            if (fullBranchingCondition == Exp.BoolLit(true)) condition else Exp.And(condition, fullBranchingCondition),
+        )
+
+    context(ssaConverter: SsaConverter)
+    fun updateLatestNameUsage(name: SymbolicName): SsaVariableName {
+        val ssaName = ssaConverter.generateFreshSsaName(name);
+        latestName[name] = ssaName;
+        return ssaName
```
jesyspa: Now this is a good use for `return ssaConverter.generateFreshSsaName(name).also { latestName[name] = it }`
Also, we generally do not use semicolons in Kotlin

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    fun generateBranchingBlockNodeFromThisNode(condition: Exp): SsaBlockNode =
+        SsaBlockNode(
+            this,
+            condition,
+            if (fullBranchingCondition == Exp.BoolLit(true)) condition else Exp.And(condition, fullBranchingCondition),
+        )
+
+    context(ssaConverter: SsaConverter)
+    fun updateLatestNameUsage(name: SymbolicName): SsaVariableName {
```
jesyspa: This does not update a usage

Issue #46, opened 2026-02-23
Postcondition in Pure Function contains Permission


The-Ray-Man:
I think there is an issue with how pure functions are translated to viper. Here an example code snipped:

```kotlin
import org.jetbrains.kotlin.formver.plugin.NeverConvert
import org.jetbrains.kotlin.formver.plugin.Pure


class X(val a: Int, var b : Int) {}

@Pure
fun id(x : X) : X {
    return x
}

fun <!VIPER_TEXT!>testFieldAccessPureWithArgShared<!>(x : X) {
    id(x)
}
```

The following viper code for the `id` is generated:

```viper
function f$id$TF$T$X(p$x: Ref): Ref
  ensures df$rt$isSubtype(df$rt$typeOf(result), df$rt$c$X())
  ensures acc(p$c$X$shared(result), wildcard)
{
  p$x
}
```

The consistency checker complains:
```
/pure_example.kt:(224,256): error: An internal error has occurred.
Details: Consistency error: Function post-conditions must not contain permissions. (<no position>)
```

PR #48, opened 2026-02-24
Fix bug #46 - no access invariants for the special function result value

AlexFalter:
This PR fixes bug #46. We only add access invariants to the postcondition if the return variable is not the special 'result' return variable of functions. A corresponding test-case was added


jesyspa:
Hi @The-Ray-Man, could you take a look at this please?

The idea looks reasonable to me, but one thing that would be worth considering is whether checking the variable name is really the best way, or whether there's something more structured.

AlexFalter:
> Is the issue the result variable, though, or that this is a pure function and so can't return permissions? Maybe checking for whether it's a function would be better?

Yeah that makes sense - I've pushed an adjustment. If that is good to go in your opinion I'll merge

The-Ray-Man:
Just noted that below the fixed line we have:
```kotlin
if (subSignature.callableType.returnsUnique) {
    addIfNotNull(returnVariable.uniquePredicateAccessInvariant())
}
```
I think here we might run into a similar issue? For example a pure function which returns an unique object. 
(I did not test this though)

PR #50, opened 2026-02-25
Ramon/havoc-shared

The-Ray-Man:
## Goal
This pull request changes the way we handle read and writes to shared fields.
- writes: The writes are mostly removed. However some part of the statement needs to be kept, because of the possibility of side effects.
- reads: Instead of inhaling-read-exhale we now use the `havoc` method to get an unknown value.
- This should prepare us for once the uniqueness type system is ready as well as increase the readability of the viper code. 

## Havoc Method
There were several havoc methods added. This was necessary, since we want to ensure the access predicate. We generate for every type a havoc method.

The resulting method in viper looks like this:

```viper
// Class specific havoc method
method  havoc$T$Foo() returns (ret: Ref)
  ensures acc(p$c$Foo$shared(ret), wildcard)
  ensures df$rt$isSubtype(df$rt$typeOf(ret), df$rt$c$Foo())

// Class specific havoc method for nullable
method  havoc$NT$Foo() returns (ret: Ref)
  ensures ret != df$rt$nullValue() ==> acc(p$c$Foo$shared(ret), wildcard)
  ensures df$rt$isSubtype(df$rt$typeOf(ret), df$rt$nullable(df$rt$c$Foo()))

// Havoc methods for primitive types

method havoc$NT$Boolean() returns (ret: Ref)
  ensures df$rt$isSubtype(df$rt$typeOf(ret), df$rt$nullable(df$rt$boolType()))

method havoc$T$Boolean() returns (ret: Ref)
  ensures df$rt$isSubtype(df$rt$typeOf(ret), df$rt$boolType())

method havoc$NT$Char() returns (ret: Ref)
  ensures df$rt$isSubtype(df$rt$typeOf(ret), df$rt$nullable(df$rt$charType()))

method havoc$T$Char() returns (ret: Ref)
  ensures df$rt$isSubtype(df$rt$typeOf(ret), df$rt$charType())

// some are excluded
```

## Access Policies
Each field embedding has a access policy. It describes what needs to happen before and after the field access. The following policies exists:
- `BY_RECEIVER_UNIQUENESS`: Used for var fields.
- `ALWAYS_READABLE`: Used for val fields (since they are immutable it does not matter whether they are unique or shared)
- `ALWAYS_WRITEABLE`: This is used for special fields (e.g. length of a list). The full permission is required/ensured at the method boundary.
- `MANUAL`: Used when a field is annotated with `@Manual`.

## Field Read
If we read from a shared var field, the value we read can not be used to reason about the program, because it could change at any time. So the following transformation is performed:

```kotlin
class X(var a: Int)

fun foo(x: X) {
  var aLocal = x.a
}
```
```viper
aLocal := havoc$T$Int()
```

## Field Write
Similarly, when writing to a shared field, the information must be immediately discarded, since the field could change at any time.
```kotlin
class X(var a: Int)

fun bar() : Int

fun foo(x: X) {
  receiver.field = bar()
}
```
Gets translated to:
```viper
anon$0 := receiver
anon$1 := bar()
```
The first assignment is necessary, because the `receiver` could have side effects e.g. `getXwithSideEffects().a`
The second assignment is necessary, because the rhs could have side effects.

If there are no possible side conditions, then the statement is removed completely. 

## Notes
- Currently the necessary information about the uniqueness of the receiver of a field access is not available. Hence we assume that the receiver is never unique.

### DIFF
--- a/formver.compiler-plugin/testData/diagnostics/conversion/classes/field_getters.fir.diag.txt
+++ b/formver.compiler-plugin/testData/diagnostics/conversion/classes/field_getters.fir.diag.txt
@@ -13,10 +13,7 @@ method f$testPrimitiveFieldGetter$TF$T$PrimitiveFields$T$Unit(p$pf: Ref)
   inhale acc(p$c$PrimitiveFields$shared(p$pf), wildcard)
   unfold acc(p$c$PrimitiveFields$shared(p$pf), wildcard)
   l0$a := p$pf.bf$a
-  inhale acc(p$pf.bf$b, write)
-  l0$b := p$pf.bf$b
-  exhale acc(p$pf.bf$b, write)
-  inhale df$rt$isSubtype(df$rt$typeOf(l0$b), df$rt$intType())
+  l0$b := havoc$T$Int()
   label lbl$ret$0
   inhale df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
 }
@@ -44,23 +41,13 @@ method f$testReferenceFieldGetter$TF$T$ReferenceFields$T$Unit(p$rf: Ref)
   inhale acc(p$c$ReferenceFields$shared(p$rf), wildcard)
   unfold acc(p$c$ReferenceFields$shared(p$rf), wildcard)
   l0$f := p$rf.bf$f
-  inhale acc(p$rf.bf$g, write)
-  l0$g := p$rf.bf$g
-  exhale acc(p$rf.bf$g, write)
-  inhale df$rt$isSubtype(df$rt$typeOf(l0$g), df$rt$c$PrimitiveFields())
-  inhale acc(p$c$PrimitiveFields$shared(l0$g), wildcard)
+  l0$g := havoc$T$PrimitiveFields()
   unfold acc(p$c$PrimitiveFields$shared(l0$f), wildcard)
   l0$fa := l0$f.bf$a
-  inhale acc(l0$f.bf$b, write)
-  l0$fb := l0$f.bf$b
-  exhale acc(l0$f.bf$b, write)
-  inhale df$rt$isSubtype(df$rt$typeOf(l0$fb), df$rt$intType())
+  l0$fb := havoc$T$Int()
   unfold acc(p$c$PrimitiveFields$shared(l0$g), wildcard)
   l0$ga := l0$g.bf$a
-  inhale acc(l0$g.bf$b, write)
-  l0$gb := l0$g.bf$b
-  exhale acc(l0$g.bf$b, write)
-  inhale df$rt$isSubtype(df$rt$typeOf(l0$gb), df$rt$intType())
+  l0$gb := havoc$T$Int()
   label lbl$ret$0
   inhale df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
 }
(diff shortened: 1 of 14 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -445,13 +454,22 @@ data class FieldModification(val receiver: ExpEmbedding, val field: FieldEmbeddi
     override fun toViperSideEffects(ctx: LinearizationContext) {
         val receiverViper = receiver.toViper(ctx)
         val newValueViper = newValue.withType(field.type).toViper(ctx)
-        val invariant =
-            field.accessInvariantForAccess()?.fillHole(ExpWrapper(receiverViper, receiver.type))
-                ?.pureToViper(toBuiltin = true, ctx.source)
+        if (field.accessPolicy == AccessPolicy.ALWAYS_VOLATILE) {
+            // We write to a shared var field. Since this value could change at any time, we just evaluate the receiver and the rhs and assign them to throw-away-variables
+            // This must be done, because the receiver as well as the rhs could have side effects.
+            val throwAwayReceiver = ctx.freshAnonVar(receiver.type)
```
The-Ray-Man: move receiver and newValue into the body and call toViperUnused result

jesyspa:
Alex, please take a look as well as this can impact your work.

The-Ray-Man:
(there are some more changes on the way. Will ping you both when I pushed all of them)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/AccessPolicy.kt
```diff
@@ -6,5 +6,18 @@
 package org.jetbrains.kotlin.formver.core.conversion
 
 enum class AccessPolicy {
-    ALWAYS_INHALE_EXHALE, ALWAYS_READABLE, ALWAYS_WRITEABLE, MANUAL;
+    /** var fields unique **/
+    UNIQUE_BASE,
+
+    /** var fields shared **/
+    ALWAYS_VOLATILE,
```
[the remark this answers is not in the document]
The-Ray-Man: I see. I was under the impression that the uniqueness checker would run before it and could annotate the field accesses with the necessary information such that we can make the correct decision. 

Nevertheless, when translating to Viper, we need that information. I am not sure where it will come from. From the ``LinearizationContext``?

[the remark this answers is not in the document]
jesyspa: Looking at this in more details, it seems I was wrong and it is per access.  My bad, in that case perhaps we want "WRITEABLE" and "DROP_ACCESSES" as names?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/HavocMethodCallBuilder.kt
```diff
@@ -0,0 +1,49 @@
...
+        return this
+    }
+
+    abstract fun build(): Stmt.MethodCall
+}
+
+
+class PrimitiveHavocMethodCallBuilder(private val preType: PretypeEmbedding) : HavocMethodCallBuilder() {
+    override fun build(): Stmt.MethodCall =
+        Stmt.MethodCall(SpecialName("havoc"), listOf(flags.adjustRuntimeType(preType.runtimeType)), listOf(target))
```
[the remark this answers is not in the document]
The-Ray-Man: I think this is safe for the following reasons: 
- For the primitive havoc (with primitive I mean types for which we do not need to ensure a predicate, like int, string etc.). Since we do not need to ensure a predicate, we can have one havoc function for all the primitive types and supply the runtime type as an argument.
- We do not collide with other ``havoc`` methods, because the other have another prefix, as well added their class name to it.
- We do not collide with other ``SpecialName``s. Special names are used in the following locations:
  - ``InjectionImageFunction``: there is no function named "havoc"
  - ``SpecialFields``: Currently there is no special field called "havoc".
  - ``ListSizeFieldEmbedding``: There the name is "size".

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Havoc.kt
```diff
@@ -0,0 +1,154 @@
...
+import org.jetbrains.kotlin.formver.viper.ast.Method
+import org.jetbrains.kotlin.formver.viper.ast.Stmt
+import org.jetbrains.kotlin.formver.viper.ast.Type
+
+/**
+ * The ``HavocMethodCallBuilder`` is used to create havoc calls.
+ * They can be created using the `Havoc` object.
+ * First the flags and the target (the variable to which the result should be written to) must be set, before calling `.build()`
+ */
+abstract class HavocMethodCallBuilder {
```
jesyspa: By the way, does having the `Method` not give you an easy way to call it in general?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
```diff
@@ -33,12 +35,23 @@ class ClassEmbeddingDetails(
...
     val sharedPredicate: Predicate
         get() = _sharedPredicate ?: error("Predicate of ${type.name} has not been initialised yet.")
     val uniquePredicate: Predicate
         get() = _uniquePredicate ?: error("Unique Predicate of ${type.name} has not been initialised yet.")
+    val havocMethod: Method
+        get() = _havocMethod ?: error("Havoc method of ${type.name} has not been initialised yet.")
+    val havocMethodNullable: Method
+        get() = _havocMethodNullable ?: error("Havoc method of ${type.name} has not been initialised yet.")
+
+    fun initHavocMethods() {
```
jesyspa: Is there any need for this init step to be separate from the rest?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -59,6 +60,7 @@ class ProgramConverter(
     private val classes: MutableMap<SymbolicName, ClassTypeEmbedding> = mutableMapOf()
     private val properties: MutableMap<SymbolicName, PropertyEmbedding> = mutableMapOf()
     private val fields: MutableSet<FieldEmbedding> = mutableSetOf()
+    private val havocMethods: MutableMap<SymbolicName, Method> = mutableMapOf()
```
jesyspa: Why put these here, rather in methods above?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/SpecialMethods.kt
```diff
@@ -5,8 +5,8 @@
 
 package org.jetbrains.kotlin.formver.core.embeddings.callables
 
-import org.jetbrains.kotlin.formver.viper.ast.BuiltInMethod
+import org.jetbrains.kotlin.formver.viper.ast.Method
 
 object SpecialMethods {
-    val all: List<BuiltInMethod> = listOf()
+    val all: List<Method> = listOf()
```
jesyspa: Is this still necessary?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -385,18 +379,25 @@ data class FieldAccess(val receiver: ExpEmbedding, val field: FieldEmbedding) :
     override fun toViperStoringIn(result: VariableEmbedding, ctx: LinearizationContext) {
         val receiverViper = receiver.toViper(ctx)
         val receiverWrapper = ExpWrapper(receiverViper, receiver.type)
```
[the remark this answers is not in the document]
jesyspa: By the way, doing this lower (inside if (field.unfoldToAccess)) would probably reduce the code needed a little.

jesyspa (approved):
Some minor suggestions, but generally LGTM!

jesyspa (changes requested):
Hang on, why are the tests failing?

PR #51, opened 2026-03-01
Alex/purity checker

AlexFalter:
This PR cleans up the purity checker. Namely:

- Missing checks have been implemented
- Postconditions of pure functions are now being checked for permissions
- Translation is aborted if a method body is invalid

### DIFF
--- a/formver.common/src/org/jetbrains/kotlin/formver/common/ErrorCollector.kt
+++ b/formver.common/src/org/jetbrains/kotlin/formver/common/ErrorCollector.kt
@@ -15,7 +15,7 @@ import org.jetbrains.kotlin.KtSourceElement
  */
 class ErrorCollector {
     private val minorErrors = mutableListOf<String>()
-    private val purityErrors = mutableListOf<Pair<KtSourceElement, String>>()
+    private val purityErrors = mutableListOf<Pair<KtSourceElement?, String>>()
 
     fun addMinorError(error: String) {
         minorErrors.add(error)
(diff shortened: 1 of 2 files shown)


AlexFalter:
As I am not sure if I missed something I will open this up as a draft for now and set it to open after a first review

jesyspa (changes requested):
Generally looks good

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/FunctionBodyEmbedding.kt
```diff
@@ -5,18 +5,28 @@
...
 
+import org.jetbrains.kotlin.KtSourceElement
 import org.jetbrains.kotlin.formver.core.conversion.ReturnTarget
 import org.jetbrains.kotlin.formver.core.embeddings.callables.FullNamedFunctionSignature
 import org.jetbrains.kotlin.formver.core.embeddings.callables.toViperMethod
 import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
 import org.jetbrains.kotlin.formver.viper.ast.Method
 import org.jetbrains.kotlin.formver.viper.ast.Stmt
 
+sealed class MethodBodyConversionResult {
```
jesyspa: This looks like it should be an interface

jesyspa (approved):
Please resolve the comment, then LGTM

PR #52, opened 2026-03-02
Remark: Z3_EXE captured by publishToMavenLocal

pawinkler:
When `publishToMavenLocal` is run, the current value of the `Z3_EXE` environmental variable is captured in the local Maven repository. Changes to the variable thus have no effect until the task is re-run.

This PR adds a small remark to the README.md which explains this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,9 @@ You need to (additionally) set `Z3_EXE` in `~/.xprofile` and/or
 `~/.bash_profile` depending on your shell, window manager, display
 manager, operating system, etc.
 
+**Note**: The `Z3_EXE` environmental variable is captured in the local Maven repository when `publishToMavenLocal`
+is run. Changing the environment variable later has no effect until the task is executed again.
+
 ## Contact
 
 Reach out to komi.golov@jetbrains.com if you'd like to use or contribute to the plugin!


jesyspa (commented):
Just to check: in slack, you mentioned that the JVM was capturing something.  Is that a separate point from this one?

pawinkler:
Some additional information about this bug for future debugging:

- The bug was confirmed using commit bd2b199 in April 2026 using Java 24 on MacOS 26.4 
- The error message is:
> ```
> An internal error has occurred.
> Details: Cannot run prover at location '[captured z3 path]': not a file.
> Please report this at https://github.com/jesyspa/kotlin
> ```

The bug can be recreate locally using these steps:
1. Clone the main SnaKt and the Snakt-Usage-Example repositories.
3. (SnaKt) Reset Gradle demons: `./gradlew --stop `.
4. (SnaKt) Set z3 path to invalid location locally: `export Z3_EXE=/usr/local/bin/z4 `.
5. (SnaKt) Build and locally publish SnaKt: `./gradlew publishToMavenLocal --rerun-tasks`.
6. (Snakt-Usage-Example) Compile Kotlin with SnaKt included: `./gradlew compileKotlin --rerun-tasks`. This results in the above mentioned error.
7. (Snakt-Usage-Example) Optionally, verify that the global `Z3_EXE` was unaffected: `echo $Z3_EXE`.

pawinkler:
This is actually the expected behavior of Gradle's Kotlin demon, which caches environment variables. Changing these requires restarting the demon.

PR #63, opened 2026-03-11
Stop Silicon verifier after use to prevent Z3 process leak

jesyspa:
## Summary

- **Z3 process leak fix**: `ViperPoweredDeclarationChecker.check()` creates a new `Verifier` (and `DefaultMainVerifier`) per declaration but never calls `stop()`. This causes Z3 processes to accumulate across verifications within a single compilation, eventually exhausting memory.
- **`Verifier.stop()`**: New method that delegates to `DefaultMainVerifier.stop()`, called in a `try/finally` to ensure cleanup on all code paths (including early returns and exceptions).
- **`SILICON_PARALLEL_VERIFIERS` env var**: Optionally limits Silicon's internal Z3 parallelism via `--numberOfParallelVerifiers`. Defaults to Silicon's existing behavior (all cores) when unset, useful for memory-constrained machines.

## Measurements (12GB machine, `--max-workers=2`)

Running full `*Verification*` test suite:

| | Z3 processes (peak) | RAM (peak) | Result |
|---|---|---|---|
| Before | ~283 | 8GB+ (OOM killed) | FAIL |
| After (stop + `SILICON_PARALLEL_VERIFIERS=2`) | ~16 | ~3.5GB | PASS (2m43s) |

## Test plan

- [x] Full `*Verification*` test suite passes with `--rerun`
- [x] No golden file changes
- [x] All other test suites (Conversion, Purity_checker, Uniqueness_checker) pass

### DIFF
--- a/formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
+++ b/formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
@@ -80,16 +80,20 @@ class ViperPoweredDeclarationChecker(private val session: FirSession, private va
                 }
             }
             val verifier = Verifier()
-            val onFailure = { err: VerifierError ->
-                val source = err.position.unwrapOr { declaration.source }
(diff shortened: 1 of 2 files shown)


jesyspa:
Thanks!  I thought about that, but I kind of doubt it - this isn't part of the JVM, it's separate processes.  But who knows :)

PR #70, opened 2026-03-12
Alex/heap dependent expressions

AlexFalter:
This PR introduces heap-dependent expressoins in pure functions. For a detailed description of the implemented algorithm, please refer to:
https://github.com/jesyspa/snakt-design/pull/5

Note: Handling of pure functions in specifications is omitted in this PR and will be added with a bigger verification example.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationContext.kt
```diff
@@ -56,6 +57,8 @@ interface LinearizationContext {
         result: VariableEmbedding?
     )
 
+    fun handleFieldAccess(field: FieldEmbedding, receiver: ExpEmbedding, result: VariableEmbedding)
```
jesyspa: What does "handle" mean?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -274,7 +276,7 @@ fun StmtConversionContext.convertFunctionWithBody(
         errorCollector.addPurityError(declaration.source, "Impure function body detected in pure function")
         return null
     }
-    val pureLinearizer = PureLinearizer(declaration.source, SharedLinearizationState(anonVarProducer))
+    val pureLinearizer = PureLinearizer(declaration.source, SharedLinearizationState(anonVarProducer), SsaConverter(declaration.source), UnfoldPolicy.UNFOLD)
```
jesyspa: Why should this pure linearizer have UNFOLD as unfold policy?

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/TypeEmbedding.kt
```diff
@@ -98,6 +99,9 @@ data class TypeEmbedding(val pretype: PretypeEmbedding, val flags: TypeEmbedding
     context(nameResolver: NameResolver)
     override val debugTreeView: TreeView
         get() = PlaintextLeaf(name.mangled)
+
+    fun hierarchyUnfoldPath(field: FieldEmbedding): Sequence<ClassTypeEmbedding>? =
+        (pretype as? ClassTypeEmbedding)?.details?.hierarchyUnfoldPath(field)
```
jesyspa: At least leave a TODO that this should be implemented properly instead of via a cast, please.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureLinearizer.kt
```diff
@@ -105,6 +103,24 @@ data class PureLinearizer(
         }
     }
 
+    override fun addFieldAccess(
+        field: FieldEmbedding,
+        receiver: ExpEmbedding,
+        result: VariableEmbedding
+    ) {
```
jesyspa: Any particular reason for this order?  Typically, if the order in the code is `x = y.z`, then that is also the best order to pass the parameters in.

jesyspa (changes requested):
You do not have any verification tests.  Is this intentional?  From what I understood, it's not easy to tell whether this code will verify.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -376,56 +373,19 @@ data class FieldAccess(val receiver: ExpEmbedding, val field: FieldEmbedding) :
     }
```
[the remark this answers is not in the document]
jesyspa: Your logic relies on heap accesses informing the SSA converter, but here there are accesses that bypass that.  Why is that correct?  Can it be avoided, so that later changes don't have to discover this by seeing something break?

jesyspa (changes requested):

jesyspa (commented):
Draft review of the heap-dependent expressions PR. Overall the approach looks reasonable — moving field access handling into the linearization contexts and tracking access invariants through the SSA converter is a clean separation. A few concerns below (posted as individual comments).

jesyspa (commented):
Draft review of the heap-dependent expressions PR. Overall the approach looks reasonable — moving field access handling into the linearization contexts and tracking access invariants through the SSA converter is a clean separation. A few inline concerns below.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationContext.kt
```diff
@@ -56,6 +54,20 @@ interface LinearizationContext {
         result: VariableEmbedding?
     )
 
+    /**
+     * This function is used to delegate the responsibility of deciding how to translate
+     * a field access to Viper to the LinearizationContext rather than the FieldAccess
+     * querying the context for some policy on how to deal with itself.
```
jesyspa: When I see a function in an interface, I should know:
- when can I call a function? 
- what does calling the function do?
- what does the function return?

In particular, "how it is used" is not a property of the function, but rather of the usages, so it should at most be a note.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationContext.kt
```diff
@@ -56,6 +54,20 @@ interface LinearizationContext {
...
+     */
+    fun translateFieldAccess(access: FieldAccess): Exp
+
+    /**
+     * This function is used to store a FieldAccess. Namely, if it is already decided that
```
jesyspa: It is not clear what it means to store a FieldAccess.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureLinearizer.kt
```diff
@@ -32,10 +33,10 @@ class PureLinearizerMisuseException(val offendingFunction: String) : IllegalStat
 data class PureLinearizer(
     override val source: KtSourceElement?,
     val state: SharedLinearizationState = SharedLinearizationState(FreshEntityProducer(::AnonymousVariableEmbedding)),
-    private val ssaConverter: SsaConverter = SsaConverter(source)
+    private val ssaConverter: SsaConverter = SsaConverter(source),
+    // If true, field access are let-bound to an anonymous variable, otherwise they are returned as an 'unfolding' expression.
+    private val storeFieldAccesses: Boolean = false
```
jesyspa: It doesn't seem like you meaningfully changed the approach; rather, you've just replaced a meaningful enum with a boolean that is harder to understand.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/purity/ExpPurityVisitor.kt
```diff
@@ -42,17 +42,17 @@ internal class ExprPurityVisitor(val declaredVariables: MutableSet<VariableEmbed
     override fun visitFieldAccess(e: FieldAccess): Boolean = e.allChildrenPure(this)
     override fun visitPrimitiveFieldAccess(e: PrimitiveFieldAccess): Boolean = e.allChildrenPure(this)
     override fun visitIs(e: Is) = e.allChildrenPure(this)
+    override fun visitCast(e: Cast): Boolean = e.allChildrenPure(this)
```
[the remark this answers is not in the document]
jesyspa: I think it depends on the direction of the cast, no?  Downcasts can require extra permissions.

jesyspa (changes requested):
Two suggestions:
- You are sending this PR, that touches many parts of the code, for the so manyeth time now.  You would have an easier time splitting off a part that you are confident in and getting that merged, so that only the issues that need to be iterated on keep getting brought up.
- You've again implemented some changes without presenting a design; you asked some questions over slack, but you did not explain your approach.  I would strongly recommend that you do so instead of trying to tweak a few things until it works.

AlexFalter:
Latest push introduces changes as discussed in Slack. Main one:

Split PureLinearizer into PureFunBodyLinearizer and PureExpLinearizer

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/purity/ExpPurityVisitor.kt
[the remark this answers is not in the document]
AlexFalter: As discussed over Slack: Such downcasts produce a InhaleInvariants embedding and will be rejected by the purity check

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -274,9 +271,13 @@ fun StmtConversionContext.convertFunctionWithBody(
         errorCollector.addPurityError(declaration.source, "Impure function body detected in pure function")
         return null
     }
-    val pureLinearizer = PureLinearizer(declaration.source, SharedLinearizationState(anonVarProducer))
-    body.toViperUnusedResult(pureLinearizer)
-    return pureLinearizer.constructExpression()
+    val pureFunBodyLinearizer = PureFunBodyLinearizer(
+        declaration.source,
+        SharedLinearizationState(anonVarProducer),
+        SsaConverter(declaration.source),
```
jesyspa, line 277: Does this need to be exposed like this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/TypeEmbedding.kt
jesyspa: reminder to leave a todo

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
```diff
@@ -93,9 +92,53 @@ data class Linearizer(
...
+                    }
+                    Stmt.assign(
+                        result.toLocalVarUse(), Exp.FieldAccess(receiverViper, field.toViper(), source.asPosition)
+                    )
+                }
+            }
+        }
+    }
+
+    private fun Sequence<ClassTypeEmbedding>?.unfoldHierarchy(
```
jesyspa: Duplicate with hierarchyUnfoldPath?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureFunBodyLinearizer.kt
```diff
@@ -105,28 +95,34 @@ data class PureLinearizer(
...
+        if (viperReceiver !is Exp.LocalVar) throw SnaktInternalException(
+            source,
+            "Invalid receiver encountered in pure function"
+        )
+        val receiverWrapper = ExpWrapper(viperReceiver, receiver.type)
+        val hierarchyPath = receiver.type.hierarchyUnfoldPath(field)
+        val accessInvariants =
+            hierarchyPath?.map { it.predicateAccess(receiverWrapper, source) }?.toList() ?: emptyList()
+        val primitiveAccess: Exp = Exp.FieldAccess(viperReceiver, field.toViper(), source.asPosition)
+        ssaConverter.addAssignment(result.name, primitiveAccess, accessInvariants)
```
jesyspa, line 111: Not clear why this is addAssignment, not store

jesyspa (approved):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureFunBodyLinearizer.kt
```diff
@@ -105,28 +95,34 @@ data class PureLinearizer(
...
+        val field = access.field
+        val viperReceiver = receiver.toViper(this)
+        if (viperReceiver !is Exp.LocalVar) throw SnaktInternalException(
+            source,
+            "Invalid receiver encountered in pure function"
+        )
+        val receiverWrapper = ExpWrapper(viperReceiver, receiver.type)
+        val hierarchyPath = receiver.type.hierarchyUnfoldPath(field)
+        val accessInvariants =
+            hierarchyPath?.map { it.predicateAccess(receiverWrapper, source) }?.toList() ?: emptyList()
```
jesyspa, line 109: Doesn't the map already return a list?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
AlexFalter: I think so, as we need the pureFunBodyLinearizer later in the function

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
AlexFalter: Not really if I see it correctly - I believe this is poor wording.
hierarchyUnfoldPath returns the path of ClassTypeEmbeddings as a sequence that needs to be unfolded
unfoldHierarchy unfolds this path in context. I will adjust the naming of this one

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureFunBodyLinearizer.kt
AlexFalter: store calls toViper on the RHS of an embedding. The expressions here are already translated, hence the SSAConverter is called directly

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/PureFunBodyLinearizer.kt
AlexFalter: hierarchyPath is a sequence, not a list. Hence, to use it as a list in the SSAConverter, I convert it here directly

PR #74, opened 2026-03-17
Fix: pure function body lost in modular verification

pawinkler:
**Description**:

When a @Pure function B is called from within another function A that is being verified, its Viper function body was missing from the generated program. The verifier could only reason about B via its pre- and postconditions, which limited the usage of pure functions. 

**Suspected Cause**:

The `PureUserFunctionEmbedding` can be generated in two different ways, either by calling `registerForVerification` (triggered when the function itself was verified; generated the body) or by calling `embedPureFunction` (triggered when the function was called in the specifications or body of another function/method; did not generate the body). 

Without the body, Viper's function inlining did not work, leading to the verification limitations.

**Fix**:

Changes in `ProgramConverter.kt`:

- New `convertAndSetPureBody` helper: converts and registers the body of a pure function
- `embedPureFunction`: triggers the new helper function
- `registerForVerification`: also triggers the new helper function

**Note**: I am not entirely sure if the implementation of the helper function satisfies our code requirements, please verify this.


jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -124,7 +121,11 @@ class ProgramConverter(
         // place the embedding in the map before processing the body.
         if (declaration.symbol.isPure(session)) {
             embedPureUserFunction(declaration.symbol, signature).apply {
```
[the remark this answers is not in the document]
pawinkler: I've improved the overall structure a bit and tried to align it with the two branches. Is this better?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -559,14 +554,40 @@ class ProgramConverter(
...
+            signature.symbol.valueParameterSymbols,
+            signature.labelName,
+            returnTarget,
+        )
+        val stmtCtx = MethodConverter(
+            this@ProgramConverter,
+            signature,
+            paramResolver,
+            scopeIndexProducer.getFresh(),
+        ).statementCtxt()
```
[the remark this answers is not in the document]
pawinkler: It's now included in the individual functions

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -103,43 +100,61 @@ class ProgramConverter(
...
-        // Note: it is important that `body` is only set after `embedUserFunction` is complete, as we need to
+        // Note: it is important that `body` is only set after embedding is complete, as we need to
         // place the embedding in the map before processing the body.
         if (declaration.symbol.isPure(session)) {
-            embedPureUserFunction(declaration.symbol, signature).apply {
-                body = stmtCtx.convertFunctionWithBody(declaration)
-            }
+            embedPureUserFunction(declaration.symbol, signature)
         } else {
+            val returnTarget = returnTargetProducer.getFresh(signature.callableType.returnType)
```
jesyspa: This code fragment looks duplicated, am I missing something?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -103,43 +100,61 @@ class ProgramConverter(
...
+                    scopeIndexProducer.getFresh(),
+                ).statementCtxt()
             embedUserFunction(declaration.symbol, signature).apply {
                 body = stmtCtx.convertMethodWithBody(declaration, signature, returnTarget)
             }
         }
     }
 
+    @OptIn(SymbolInternals::class)
     fun embedPureUserFunction(
```
jesyspa: Would it make sense for this to be private?

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: I've merged your PR and resolved it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -104,46 +104,54 @@ class ProgramConverter(
...
         }
     }
 
-    fun embedPureUserFunction(
+    @OptIn(SymbolInternals::class)
+    private fun embedPureUserFunction(
         symbol: FirFunctionSymbol<*>,
         signature: FullNamedFunctionSignature
     ): PureUserFunctionEmbedding {
+        // Return if already registered
```
jesyspa: This comment is not helpful, please remove it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -104,46 +104,54 @@ class ProgramConverter(
...
         (functions[signature.name] as? PureUserFunctionEmbedding)?.also { return it }
         val new = PureUserFunctionEmbedding(processCallable(symbol, signature))
+        // Insert into the map before processing the body, so recursive calls can find the embedding.
         functions[signature.name] = new
+        val declaration = symbol.fir as? FirSimpleFunction
```
jesyspa, line 129: What is the desired behaviour if this cast fails?

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: The cast would fail when `symbol` is not a FirSimpleFunction, but we allow the `@Pure` annotation only on normal functions. So if I understand correctly, this should never occur, but it could fail silently. Should I throw an error here for safety?

jesyspa: Yes, if the case cannot happen please throw an exception.

jesyspa (approved):

Issue #72, opened 2026-03-13
Permission Management


The-Ray-Man:
## Tracking: Permission Management System (Fold/Unfold)

This issue tracks the implementation of the permission management strategy as defined in the [folding-unfolding.md](https://github.com/jesyspa/snakt-design/blob/ramon/unfold-fold/Design/folding-unfolding.md) design document. The core goal is to integrate Uniqueness Analysis results into the translation process to automate predicate folding and unfolding.

### Phase 1: Data Extraction & Association

* [x] Uniqueness Analysis Integration:
  * Execute uniqueness analysis for every method to extract flow facts.  
  * **Location:** Hook into `registerForVerification` within `ProgramConverter`.  
* [x] FIR-to-CFG Mapping:
  * Implement a DFS traversal on `CFGNodes` to build a mapping from `FirElement` to its corresponding `CFGNodes`.  
  * **Requirement:** Store only the first and last nodes for each element.  
  * **Note:** Pay close attention to edge cases where FIR elements may not map 1:1 to CFG structures.

### Phase 2: Core Infrastructure Updates

* [x] State Persistence in `MethodConverter`:
  * Store the `firToNodes` map and `flowFacts` within `MethodConverter` to make them accessible during the `FirElement` -> `ExpEmbedding` translation.  
* [x] ExpEmbedding Interface Changes:  
  * Add properties to represent the **Uniqueness state** both before and after expression execution.  
  * Add a property to track **used paths**, implemented via a recursive lookup.

### **Phase 3: Translation & Logic Implementation**

* [x] Update `StmtConversionVisitor`:
  * Modify the visitor to inject the Uniqueness state into generated `ExpEmbeddings` where applicable.  
* [x] Permission Management Logic: 
  * [x] Create a `PermissionManagement` object/service.  
  * [x] **Responsibility 1:** Identify necessary fold/unfold operations for a given `ExpEmbedding`.  
  * [x] **Responsibility 2:** Extract necessary invariants.  
* [ ] Unique Predicate Logic:
  * Update the in the design repository mentioned `ExpEmbedding`s to insert folds and unfolds of unique predicates where  necessary. 
    * [x] `Declare`
    * [x] `Assign`
    * [x] `FieldModification`
    * [x] `MethodCall`
    * [ ] `Return`. Partially Implemented - missing support from uniqueness checker.
    * [x] `If`
    * [x] `While`
    * [ ] `Elvis` missing support from uniqueness checker.
[...]

PR #75, opened 2026-03-17
Permission Management

The-Ray-Man:
This PR targets #72 

# Overview of Changes

## Uniqueness Information
When a program is registered for verification, we call the uniqueness checker. This also means that the verification plugin is now dependent on the uniqueness plugin.
The uniqueness information is collected in a `UniquenessInformation` object, which using dfs maps the in/out flows from the cfg nodes to the corresponding firEmbeddings.
The uniqueness information is added to the `MethodConverter` for the next step.

### New Properties on ExpEmbedding
For every `ExpEmbedding` we want to have access to the uniqueness information before and after the execution of it. Two mutable fields were added on the `ExpEmbedding`. Generally having immutable fields would make more sense, however this would require to change all the classes and add them to the constructor, which would make the construction less clean.
We use the abstract class `DefaultUniqueness` to initialize both fields to null. 

>Question: Is this the best way to do it? Should we change it to immutable properties and add the values during construction?

### Fir to ExpEmbedding translation
The `StmtConversionVisitor` was updated to also set the uniqueness information for each generated `ExpEmbedding`.

## Permission Manager
The permission management can be divided into three parts: The management of unique predicates, shared predicates and the invariant extraction.  For each of those parts a separate object was created.
The task of the permission manager is to figure out what needs to be unfolded/folded.

All the permission managers have a private constructor, they can be created via the `create` method. The creation can fail, if we do not have enough information available. Since currently there are some code structures that are not yet supported by the uniqueness checker, we do not always have the necessary information (flow before/after) available. With the failing construction we can "soft fail". 

>Question: We could also remove the PermissionManager objects and move the logic into the `ExpEmbedding`. I think with the `PermissionManager` we have the logic a bit more separated and the `ExpEmbedding` remain rather simple. But I am not sure, if that is the best way to do it?

>Question: Also about handling cases where there is not enough information available. In an ideal world this would never happen. There are two points of failure: missing uniqueness information, missing path information (see below).  What should be de default for such situation behaviour there?  Just unfold nothing (likely result in a viper error, but should not introduce unsoundness)?

### Path Abstraction
The added `Path` classes are used to associate the `ExpEmbedding` paths with the fir Paths and add some utilities that can be used by the `PermissionManager`.  On each `ExpEmbedding` there were two new properties added:
- `endingPath`: Is the path that ends in the current `ExpEmbedding`. This path may be extended if the current `ExpEmbedding` becomes the receiver of a `FieldAccess`. 
- `containingPaths`: This is used to collect paths for larger statements such as assignments. All the paths contained in the assignment are collected there.  Usually only the final path is added here and not all the prefixes. So if the statement is `a.b = x.y.z + 1`, the collection would consist of `Path(a.b), Path(x.y.z)`. This does not mean however that two paths never share a prefix. For the statement `x.y = x.y.z + 1` both `Path(x.y)` and `Path(x.y.z)` are included.

>Question: At the moment these are implemented as lazy fields. I am not sure if this is the best way to do it. I used lazy fields, because we do not always need them, so calculating them might be wasteful. We could also implement a visitor that extracts the paths, but this visitor might make unnecessary passes. E.g. extract the paths on the statement level (for unique predicates) and afterwards again on field access level (for shared predicates). 

### Linearizer
On the `Linearizer` I added a new method `addBranchWithFolding`. This method should add a branch, and does the necessary folding on the `condition`. This was necessary, because:
- The unique predicates must be unfolded on "statement" level. Because for example for an assignment, a used field only becomes moved once the read value is assigned. For example: `b = tree.left + tree.right` , on the `FieldAccess` level both the access to `left` and `right` look the same and would conclude that `unique(tree)` must be unfolded. However, we can only unfold `unique(tree)` once.
- For branches this means that we need to look at the whole condition and do the necessary uniqueness folding before and after the whole condition. We can not do this on the `If` `ExpEmbedding` level, because this would involve using the linearizer to add statements, which we can not do, in pure settings. Therefore we moved the unfolding into the linearizer and created a new method (`addBranchWithGolding`). 

>Question: I think a better way to do this, would be to use `unfoldPolicy` to decided if we should add unfold statements or not. Is that correct? 

Because of the added uniqueness unfolds on the branch level it was necessary to update `SequentialLogicOperatorEmbedding`. When translation conditions that could short-circuit, a `If` `ExpEmbedding` was created and translated to viper. However if we do it like this, some unique predicates might become unfolded twice. Because multiple nested `If`s are created. I changed it, such that the `SequentialLogicOperatorEmbedding`, use the linearization context to add their branches.

## Limitations
Currently there are quite some parts that are not supported:
- Generally all member functions. It is unclear, if `this` should be considered unique of shared. Currently, the permission Managers just ignore those statements. Hence, nothing is unfolded.
> Question: Is this a reasonable default, or should we assume that is uniqe/shared? Or just leave it for now until we figure out how the uniqueness checker should handle this situations?
- Selecting the correct predicate for classes that inherit. At the moment just the predicate of the actual type is unfolded/folded.
- Return, Casts, Safe Calls, Elvis Operator: the uniqueness checker provides no information on these statements.
- The `Unfolding Strategy` of the `Linearizer` is not taken into account.
- The Invariant extraction needs some more work.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
@@ -1,0 +1,74 @@
+package org.jetbrains.kotlin.formver.core.conversion
+
+import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
+import org.jetbrains.kotlin.formver.core.embeddings.SourceRole
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.PrimitiveFieldAccess
+import org.jetbrains.kotlin.formver.core.embeddings.expression.VariableEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.withType
+import org.jetbrains.kotlin.formver.core.embeddings.properties.FieldEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+
+/**
+ * Path abstraction to handle paths. The core functionality is to associate the `ExpEmbedding` paths with the `firPaths`.
+ * This is necessary, because the uniqueness checker works on the fir level, whereas the permission management works
+ * on the `ExpEmbedding` level.
+ */
+sealed interface Path {
+    fun addField(field: FieldEmbedding): Path {
+        return PathElement(this, field)
+    }
+
+    val length: Int
+
+    val firPath: List<FirBasedSymbol<*>>
+
+    fun pathWithoutLast(): Path?
+
+    val type: TypeEmbedding
+
+    val expEmbedding: ExpEmbedding
+
+    /**
+     * Returns a list of all prefixes from the path, ordered ascending by the length.
+     */
+    fun traverse(): List<Path>
+
+}
+
+
+data class PathRoot(val base: VariableEmbedding) : Path {
+    override val length: Int
+        get() = 1
+
+    override val firPath = listOf((base.sourceRole!! as SourceRole.FirSymbolHolder).firSymbol)
+    override val type = base.type
+    override val expEmbedding = base
+
+    override fun pathWithoutLast(): Path? = null
+    override fun traverse() = listOf(this)
+}
+
+data class PathElement(val base: Path, val field: FieldEmbedding) : Path {
+    override val length: Int
+        get() = base.length + 1
+
+    override val firPath = base.firPath + field.symbol!!
+    override val type = field.type
+    override val expEmbedding = PrimitiveFieldAccess(base.expEmbedding, field)
+
+    override fun pathWithoutLast(): Path = base
+
+    override fun traverse() = base.traverse() + this
+
+}
+
+data class PathCast(val base: Path, val newType: TypeEmbedding) : Path {
+
+    override val length: Int = base.length
+    override val firPath: List<FirBasedSymbol<*>> = base.firPath
+    override fun pathWithoutLast(): Path? = base.pathWithoutLast()
+    override val type: TypeEmbedding = newType
+    override val expEmbedding: ExpEmbedding = base.expEmbedding.withType(newType)
+    override fun traverse(): List<Path> = base.traverse().dropLast(1) + this
+}
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
@@ -1,13 +1,13 @@
 package org.jetbrains.kotlin.formver.core.conversion
 
 import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
-import org.jetbrains.kotlin.formver.core.embeddings.SourceRole
 import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.FirVariableEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.expression.PrimitiveFieldAccess
-import org.jetbrains.kotlin.formver.core.embeddings.expression.VariableEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.expression.withType
-import org.jetbrains.kotlin.formver.core.embeddings.properties.FieldEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.properties.UserFieldEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.uniqueness.Path as FirPath
 
 /**
  * Path abstraction to handle paths. The core functionality is to associate the `ExpEmbedding` paths with the `firPaths`.
@@ -15,13 +15,13 @@ import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
  * on the `ExpEmbedding` level.
  */
 sealed interface Path {
-    fun addField(field: FieldEmbedding): Path {
+    fun addField(field: UserFieldEmbedding): Path {
         return PathElement(this, field)
     }
 
     val length: Int
 
-    val firPath: List<FirBasedSymbol<*>>
+    val firPath: FirPath
 
     fun pathWithoutLast(): Path?
 
@@ -37,11 +37,11 @@ sealed interface Path {
 }
 
 
-data class PathRoot(val base: VariableEmbedding) : Path {
+data class PathRoot(val base: FirVariableEmbedding) : Path {
     override val length: Int
         get() = 1
 
-    override val firPath = listOf((base.sourceRole!! as SourceRole.FirSymbolHolder).firSymbol)
+    override val firPath = listOf(base.symbol)
     override val type = base.type
     override val expEmbedding = base
 
@@ -49,11 +49,11 @@ data class PathRoot(val base: VariableEmbedding) : Path {
     override fun traverse() = listOf(this)
 }
 
-data class PathElement(val base: Path, val field: FieldEmbedding) : Path {
+data class PathElement(val base: Path, val field: UserFieldEmbedding) : Path {
     override val length: Int
         get() = base.length + 1
 
-    override val firPath = base.firPath + field.symbol!!
+    override val firPath = base.firPath + field.symbol
     override val type = field.type
     override val expEmbedding = PrimitiveFieldAccess(base.expEmbedding, field)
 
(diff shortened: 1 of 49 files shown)


jesyspa (commented):
Sorry for the delay on this.  Let's discuss how it can be split up.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.jetbrains.kotlin.formver.core.embeddings.properties.UserFieldEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.uniqueness.Path as FirPath
+
+/**
+ * Path abstraction to handle paths. The core functionality is to associate the `ExpEmbedding` paths with the `firPaths`.
+ * This is necessary, because the uniqueness checker works on the fir level, whereas the permission management works
+ * on the `ExpEmbedding` level.
+ */
+sealed interface Path {
```
ligee, line 17: I think with this naming we're looking for trouble. Having `uniqueness.Path` was likely a mistake, but renaming it and introducing a new 'Path' here is doubly so. Please use some relatively unique compound name.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
```diff
@@ -0,0 +1,74 @@
...
+     * Returns a list of all prefixes from the path, ordered ascending by the length.
+     */
+    fun traverse(): List<Path>
+
+}
+
+
+data class PathRoot(val base: FirVariableEmbedding) : Path {
+    override val length: Int
+        get() = 1
```
ligee, line 42: why do you need a getter here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
```diff
@@ -0,0 +1,74 @@
...
+    override val type = base.type
+    override val expEmbedding = base
+
+    override fun pathWithoutLast(): Path? = null
+    override fun traverse() = listOf(this)
+}
+
+data class PathElement(val base: Path, val field: UserFieldEmbedding) : Path {
+    override val length: Int
+        get() = base.length + 1
```
ligee, line 54: and here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/Path.kt
```diff
@@ -0,0 +1,74 @@
...
+ * Path abstraction to handle paths. The core functionality is to associate the `ExpEmbedding` paths with the `firPaths`.
+ * This is necessary, because the uniqueness checker works on the fir level, whereas the permission management works
+ * on the `ExpEmbedding` level.
+ */
+sealed interface Path {
+    fun addField(field: UserFieldEmbedding): Path {
+        return PathElement(this, field)
+    }
+
+    val length: Int
```
ligee, line 22: It will help the readability if you'll have the order consistent (unless you have a good reason otherwise). E.g., properties then functions.
And maybe keep the same order in the implementations too.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/UniquenessInformation.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.jetbrains.kotlin.fir.FirElement
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.CFGNode
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.EnterNodeMarker
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.ExitNodeMarker
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.FunctionCallExitNode
+import org.jetbrains.kotlin.formver.uniqueness.FlowFacts
+import org.jetbrains.kotlin.formver.uniqueness.UniquenessTrie
+
+
+class UniquenessInformation(val root: CFGNode<*>, val flowFacts: FlowFacts<UniquenessTrie>) {
```
ligee, line 17: Please add a kdoc comment here with the intended purpose of this class.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/UniquenessInformation.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    fun flowAfter(firElement: FirElement): UniquenessTrie? {
+        return nodeCollectionMap[firElement]?.exit?.let { flowFacts.flowAfter(it) }
+    }
+
+    class NodeCollection {
+        private var _entry: CFGNode<*>? = null
+        private var _exit: CFGNode<*>? = null
+        private val _all: MutableList<CFGNode<*>> = mutableListOf()
```
ligee, line 32: Please, do not use `_`-prefixed fields, unless you have a public prop with the same name. Even this case will be obsolete as soon as we release the feature about explicit baking fields.

formver.compiler-plugin/testData/diagnostics/conversion/operators/safe_call.kt
```diff
@@ -1,5 +1,4 @@
 // NEVER_VALIDATE
-
```
ligee, line 2: This change should probably be reverted, along with the resulting diag

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -39,6 +41,28 @@ sealed interface ExpEmbedding : DebugPrintable {
...
+    /**
+     * Uniqueness tree after the `ExpEmbedding` is executed.
+     */
+    var uniquenessAfter: UniquenessTrie?
+
+    /**
+     * All paths that appear inside the `ExpEmbedding`.
+     */
+    val containingPaths: Lazy<Set<Path>>
+        get() = lazy { emptySet() }
```
ligee, line 58: This is quite fishy. You need a very good reason to have a lazy value that is returned from getter, which is lazy by itself. 
You may want to have a lazily cached value for it in some cases, but `Lazy` should not be a part of the interface in general.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -90,6 +114,16 @@ sealed interface ExpEmbedding : DebugPrintable {
     fun isValid(ctx: PurityContext): Boolean = true
 }
 
+/**
+ * Used to set default behavior for the permission related fields.
+ */
+abstract class DefaultUniqueness : ExpEmbedding {
```
ligee, line 120: I do not understand, why do you need it in a separate abstract class, instead of placing the default implementations right into `ExpEmbedding`. Do you have a reason for it?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -347,10 +392,22 @@ fun List<ExpEmbedding>.toViper(ctx: LinearizationContext): List<Exp> = map { it.
...
 
+    override val endingPath: Lazy<Path?> = lazy {
+        // only UserFieldEmbeddings are support, because it must have a backlink to the corresponding fir element.
+        when (field) {
+            is UserFieldEmbedding -> inner.endingPath.value?.addField(field)
+            else -> null
+        }
+    }
+
+    override val containingPaths: Lazy<Set<Path>> = lazy {
```
ligee, line 407: Do I understand correctly that this should be the default implementation anyway?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ExpEmbedding.kt
```diff
@@ -39,6 +41,28 @@ sealed interface ExpEmbedding : DebugPrintable {
     val sourceRole: SourceRole?
         get() = null
 
+    /**
+     * Uniqueness tree before the `ExpEmbedding` is executed.
+     */
+    var uniquenessBefore: UniquenessTrie?
```
ligee, line 47: Having vars here is a potential source of problems. As you've mentioned in the intro, it makes sense to convert then into vals and pass to the constructor.
I've checked some creation sites and it doesn't seem very difficult to me.

ligee (changes requested):

PR #81, opened 2026-03-19
Enhance CFG renderer with flow addition

The-Ray-Man:
This pull request enhances the rendering and debugging utilities for uniqueness analysis within the Kotlin Formver compiler plugin. These improvements are designed to make the internal state of the analysis more transparent and easier to verify during development.

## Key Changes
- CFG Rendering: Introduced a new utility to render Control Flow Graphs (CFG) alongside their associated uniqueness facts.
- Trie Visualization: Enhanced the UniquenessTrie class with a custom render method to improve symbol display and structural clarity.

### From the Compiler
To visualize the analysis state, use a debugger to set a breakpoint at a location where both the CFG and flow facts are in scope.

Execute the following command in the evaluator:
`graph.render(facts)`

This will output a Graphviz (DOT) string that can be rendered into a visual diagram.

### From Testing
Annotated the file you want to dump with `// DUMP_UNIQUENESS_CFG`
The graph is added to the corresponding `.diag` file.

## Technical Notes
Type Representation: The types displayed correspond to the nodes within the Trie. Note that "in-head" joining may be required to resolve the actual resulting type at a specific point.

Delta-based Display: To reduce visual noise, types are only displayed if there is a difference between `flowBefore` and `flowAfter`.

Function Boundaries: For clarity, types are always displayed at function boundaries, regardless of whether the flow facts have changed.
## Example
```kotlin
class A {
    @Unique var x = Object()
    @Unique var w = Object()
}

class B {
    @Unique var y = A()
}

fun test(@Unique b1: B, @Unique b2: B) {
    if (b1.y == b2.y) {
        @Unique var b1tmp = b1.y
        b2.y = b1.y
        b1.y = b1tmp
    } else {
        @Unique var stored = b1.y.w
        borrow(b1.y.x)
        b1.y.w = stored
    }
}



fun borrow(@Borrowed a: Object) {}
```
graphviz (1)

### DIFF
--- a/formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
+++ b/formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
@@ -44,7 +44,6 @@ fun render(node: CFGNode<*>, facts: FlowFacts<UniquenessTrie>): String {
     }
 }
 
-@Suppress("unused") // Can be used from the debugger
 fun ControlFlowGraph.render(facts: FlowFacts<UniquenessTrie>): String {
     val options = ControlFlowGraphRenderOptions(
         data = { data: CFGNode<*> -> render(data, facts) },
--- a/formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
+++ b/formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
@@ -1,23 +1,47 @@
 package org.jetbrains.kotlin.formver.uniqueness.utils
 
-import org.jetbrains.kotlin.fir.resolve.dfa.cfg.CFGNode
-import org.jetbrains.kotlin.fir.resolve.dfa.cfg.ControlFlowGraph
-import org.jetbrains.kotlin.fir.resolve.dfa.cfg.ControlFlowGraphRenderOptions
-import org.jetbrains.kotlin.fir.resolve.dfa.cfg.renderTo
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.*
 import org.jetbrains.kotlin.formver.uniqueness.FlowFacts
 import org.jetbrains.kotlin.formver.uniqueness.UniquenessTrie
 
 
+/**
+ * This method creates the uniqueness-related content for each CFG Node. To not overload the graph with data, not everything is displayed.
+ * In these cases some flow information is added:
+ * - The flow changed when executing the node: In and Out flow are displayed
+ * - On the border of the function the flow is displayed
+ * - On nodes that join multiple flows the flow is displayed
+ */
 fun render(node: CFGNode<*>, facts: FlowFacts<UniquenessTrie>): String {
     val flowBefore = facts.flowBefore(node)
     val flowAfter = facts.flowAfter(node)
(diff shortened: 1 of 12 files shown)


jesyspa:
Hi, could you send this to Ilya please?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
```diff
@@ -0,0 +1,31 @@
...
+    val builder = StringBuilder()
+
+    builder.appendLine("Before:")
+    builder.appendLine(flowBefore.render())
+    builder.appendLine("After:")
+    builder.appendLine(flowAfter.render())
+    return builder.toString()
+}
+
+@Suppress("unused") // Can be used from the debugger
```
ligee: It would be good to add a simple test for it. And it will also allow you to remove the suppress.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
```diff
@@ -1,23 +1,36 @@
...
     val flowBefore = facts.flowBefore(node)
     val flowAfter = facts.flowAfter(node)
     val builder = StringBuilder()
 
-    builder.appendLine("Before:")
-    builder.appendLine(flowBefore.render())
-    builder.appendLine("After:")
-    builder.appendLine(flowAfter.render())
-    return builder.toString()
+    if (node is FunctionEnterNode) {
```
ligee: Please add comments explaining the intended layout. Your comments to the PR are good, you can just copy some of them to the code comments.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
```diff
@@ -1,23 +1,36 @@
...
+        builder.appendLine("Initial:")
```
ligee: BTW, you can use `buildString` instead; it will make the code a bit less noisy.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
```diff
@@ -22,6 +22,12 @@ fun render(node: CFGNode<*>, facts: FlowFacts<UniquenessTrie>): String? {
         return builder.toString()
     }
 
+    if (node.previousNodes.count() > 1) {
```
ligee: I would recommend to fixup it to the previous commit.

ligee (changes requested):

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/utils/CFGRenderer.kt
The-Ray-Man: I added the `// DUMP_UNIQUENESS_CFG` flag, which can be included in a file to dump the uniqueness CFG for each method. Currently, this output is sent to the `.diag` file. I am not sure if that is the ideal location or if a separate file should be created. however, since this is primarily for debugging, I don't think it matters much.

ligee (approved):

PR #82, opened 2026-03-20
Support casts in uniqueness checker

### LINKED ISSUE #79: Support of Casts In Uniqueness Checker
The following test is accepted by the Uniqueness Checker. However the `consume` should not succeed because `current` is partially moved.

```kotlin
class A(
    @Unique val next : A?
)

fun <!VIPER_TEXT!>consume<!>(@Unique a : A) {}

fun <!VIPER_TEXT!>test<!>(@Unique a: A) {
    val current = a.next
    if (current != null) {
        val local = current.next
        consume(current)

    }
}
```

I think the problem is, that the CFG node for the access of `current` is followed by a `SmartCastExpressionExitNode` which seems to lose some information.

paganma:
This PR adds support for extracting paths from cast and smartcast expressions.
In addition to this, it also adds the following fixes:
- Refactor the logic for resolving function declarations from `UniquenessGraphAnalyzer` and `UniquenessGraphChecker`.
- Fixes a bug where the whole flow UniquenessTrie was checked for invariance as opposed to just the subtree of the leaked variable.

### DIFF
--- a/formver.compiler-plugin/testData/diagnostics/uniqueness_checker/consume_local.fir.diag.txt
+++ b/formver.compiler-plugin/testData/diagnostics/uniqueness_checker/consume_local.fir.diag.txt
@@ -11,3 +11,5 @@
 /consume_local.kt:(996,997): error: Argument uniqueness mismatch: expected 'unique global', actual 'unique local'.
 
 /consume_local.kt:(1122,1123): error: Argument uniqueness mismatch: expected 'unique global', actual 'unique local'.
+
+/consume_local.kt:(1321,1322): error: Argument uniqueness mismatch: expected 'unique global', actual 'moved'.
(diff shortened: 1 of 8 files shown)


jesyspa:
I am going to be busy until Wednesday at the earliest, so I think sending this to @ligee for review would make more sense.  That said, it looks like you're still making changes to it - please request a review only when you're done.

jesyspa (approved):
Cool special cases, agreed that this is the intended behaviour.

komiputer:
Rebased so this branch contains only its own change (the size postcondition, its fixture, and its goldens); no longer stacked on the other Set PRs. Also fixed the postcondition to be sound: size only grows by one when `add` returns `true`; unchanged when it returns `false`.

PR #86, opened 2026-03-25
Propagate access dependencies for ternary expressions in pure functions

AlexFalter:
In this PR we introduce conditional heap-dependent expressions in pure functions. Namely, we extend the already existing model, which passed access dependencies unconditionally, and extend it by a condition. On field access or function usage, instead of encoding an expression as an unconditional access, we create a ternary expression unfolding predicates depending on the condition under which what predicates have to be unfolded. To validate our implementation we have created various test cases

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
@@ -72,14 +72,16 @@ class SsaConverter(
     ) {
         val ssaName = head.updateLatestName(name)
         varExp.propagateAccessInvariants(ssaName)
+        val propagatedInvariants = accessInvariants[ssaName]
         if (newVarAccessInvariants.isNotEmpty()) {
-            val propagatedInvariants = accessInvariants[ssaName]
             if (propagatedInvariants == null || propagatedInvariants.isEmpty()) {
                 accessInvariants[ssaName] = mapOf(Exp.BoolLit(true) to newVarAccessInvariants)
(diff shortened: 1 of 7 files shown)


jesyspa (commented):
I am still quite concerned about this case:

```kotlin
class Box(inner: Box?)
fun f(x: Box?, y: Box?): Box? = (x ?: y)?.inner?.inner
```

It seems to me like the current approach will not correctly unfold permissions, but maybe I'm missing something.

I'll test it out and get this merged if it's okay - no action needed on your part.

komiputer:
Taking ownership of this PR. I've rebased onto current main and opened a fresh PR from my fork: see follow-up link below. Closing this one as superseded.

Summary of changes on top of Alex's commits:
- Rebased on current main (5 commits replayed; the testdata-only `Adjust tests` commit became empty after upstream's test reorganization and was dropped).
- Simplified `withAccessDependencies` to collapse the ternary when all branches share the same unfoldings (fixes a verbosity regression in `local_variables` and `branching` goldens).
- Moved `pure_function_with_heap_dependent_expressions` out of `negative/` since `getSafeNextNextValue` now verifies with Alex's fix.
- Added `elvis_receiver_chain.kt` covering `(x ?: y)?.inner?.inner` to confirm the elvis-then-safe-call lowering propagates access dependencies correctly through the SSA TernaryExp path.

`testSequentialBranches` and `testMultipleVarMerge` still don't fully verify and are now marked with `<!VIPER_VERIFICATION_ERROR!>`; the fix is incremental, not complete.

komiputer:
Couldn't close from the CLI (insufficient permissions on this repo). Will need to be closed manually.

PR #87, opened 2026-03-25
Add early returns

AlexFalter:
This PR introduces the proper handling of early returns in pure functions by adding a returns value to the SSABlockNode, updating it to be true once a return is encountered and infering the full branching condition of a BlockNode after a join accordlingly.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
@@ -7,6 +7,7 @@ import org.jetbrains.kotlin.formver.core.names.SsaVariableName
 import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Declaration
 import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.Exp.Companion.toDisjunction
 import org.jetbrains.kotlin.formver.viper.ast.Type
 
 class SsaConverter(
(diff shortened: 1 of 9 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -38,7 +39,13 @@ class SsaConverter(
             condition,
             this
         )
-        head = SsaBlockNode(joinNode, splitPoint.fullBranchingCondition)
+        val branchCondition = when {
```
jesyspa, line 42: Need to explain reasoning

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaNode.kt
```diff
@@ -23,9 +23,10 @@ class SsaStartNode() : SsaNode {
 
 class SsaBlockNode(
     private val predecessor: SsaNode,
-    val fullBranchingCondition: Exp
+    val fullBranchingCondition: Exp,
 ) : SsaNode {
     val latestName: MutableMap<SymbolicName, SsaVariableName> = mutableMapOf()
+    var returns: Boolean = false
```
jesyspa, line 29: Not sure mutability is a good idea here, need to look into it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/SsaConverter.kt
```diff
@@ -38,7 +39,13 @@ class SsaConverter(
...
+            thenResultHead.returns && head.returns -> Exp.BoolLit(false)
+            thenResultHead.returns -> head.fullBranchingCondition
+            head.returns -> thenResultHead.fullBranchingCondition
+            else -> listOf(thenResultHead.fullBranchingCondition, head.fullBranchingCondition).toDisjunction()
```
jesyspa, line 46: This is always a list of two, so it can just be an Or

jesyspa (changes requested):
Will get back to this later, needs some cleanup.

Alex, no action needed on your part, just notes to myself.

komiputer:
Closing in favor of #162, which rebases this onto current main, regenerates the golden files, and addresses the inline review notes.

komiputer:
Picked this up and pushed a rebased version at #162 — current main, golden files regenerated, and your inline notes (Or vs toDisjunction, when-comment, mutability of `returns`) addressed. Closing this one would be cleaner; I lacked the permission to do so from my side.

PR #89, opened 2026-03-29
Alex/loosen axiom trigger

AlexFalter:
This PR loosens the trigger of the axiom ensuring references of underlying primitives are equivalent, to ensure verification of examples not inferring the previous trigger condition.


The-Ray-Man:
Can we merge this? In #137 I had the same issue with the trigger.

The-Ray-Man (approved):

PR #91, opened 2026-03-31
Short Name Resolver - Reworked Name Mangeling

The-Ray-Man:
# Short Name Resolver
The core of the PR is the `ShortNameResolver`, it's goal is to create more human readable viper code. The ShortNameResolver is able to reduce the number of characters in the diagnostics files by about 25%.

## How does it wok?
Every name has a ordered list of name candidates. To build this candidates a custom DSL exists. The earlier candidates are usually shorter and the last candidate should be the most specific name.

During the registration the `shortNameResolver` collects all names and builds a dependency graph with it. It also infers if a name could appear in the Viper Program. This graph can be viewed form the debugger by calling the `render()` method. 

Once the `mangle` function is called the resolver will use the graph to find unique and short names for all the names that can appear in viper. 
First it will calculated the represented by relation on the names. This is used for things that have different names but we want to merge them. This can happen for public fields with the same name. In Viper we can allow them to be shared. It can happen that the public fields have structurally a different name, therefore they are not deduplicated when registering the names. This usually happens for getters and setters of public fields with the same name, when they are defined in inner classes.

This works in a iterative progress:
1. Identify the conflicting names and group them
2. Choose one group to resolve
3. Select one name and update it's name to the next candidate. To be more precise: If we want to move the name of `nameA`, we look at the current candidate `canA` of `nameA`. If `canA` consists of other names, we will check if we can move one of those, if there are no "parentNames" that can be moved, or it just consists of strings we will select the next candidate of `nameA`. We do this that way, because many "leaf names" are `ScopedNames` and quite early they have the name of the scope as part of their candidate. Since scopes make the code bloaty very fast, we first try to not use them.
5. Restart at 1

Note: In each iteration, we only move one single name. We do it like this, because the moving of the first name might also solve other conflicts. So by this approach we will not perform unnecessary moves.

Afterwards the resolver generated a mapping from the names to their string representation, which is then used to resolve the names during the translation to silver. Before translating them to silver, we need to use the string names to remove the duplicated methods which can be shared. 

### Separator Selector
To separate the names we need some kind of separator. Beforehand the dollar symbol was used. Personally, I think that the names are much harder to read if the dollar symbol is used. The new resolver tries to use `_` as separator. However if the program submitted by the user already contains `_` it will use the dollar symbol.

### Backwards Combability 
[Here](https://github.com/JetBrains/SnaKt/pull/91/changes/39b7b28b81391046bfff86acdcae0ca273986cb2..1792ec372cb4cc2e75e2bdfc2597679540a277c9#diff-caa71f10c2bdbece93970c07f23ade04840ee5042e817d640f780dc560b804ad) it can be seen that using the old resolver the modifications on the resulting viper code are minimal.
commit e31bace shows the passing tests using the new resolver.
commit 5492245 shows the passing tests generated by codes (I plan to drop this commit before merging)

# Additional Changes
The organization of the name system was changed a bit. Most of the changes are more organizational than behaviour changes.

## Name Registration
Beforehand not all names were registered during the `check` method. There were names missing in pre/postconditions as well as invariants etc.. This names have to be known, because Viper does not allow to shadow local variables with quantifiers. 

## NameType
Before each `SymbolicName` has the optional property `mangledType` which specify the type (function, predicate, domain etc.) of the associated entity. These names should be unique, but they were distributed in many files.

We introduced a class (sealed) `NameType` to collect all these names in one location.

## NamedEntity
Since the new resolver wants to collect many names in one place we need a common type. `NamedEntity` was created to be a supertype of `NameScope`, `NameType` and `SymbolicName`.

## SymbolicName Hierarchy
The type hierarchy of the SymbolicNames was refined. 
- All the names which are not in the kotlin program and are unique, are collected under `FreshName`. 
- All the Names which are associated with a type (`TypeName`, `FunctionTypeName` etc.) are collected under `NameOfType`. The `ClassKotlinName` is not a subtype of `TypeName`. 
- `ViperKeyword` is also a `NamedEntity`. It is used to check that there are no collisions with the keywords.

## SpecialName
Before `SpecialName` served multiple purposes. It was used for special fields but also for functions generated by the domain. Since these two things have basically nothing in common, `SpecialName` was split into: `SpecialFieldName` (used for special fields) and `RelatedDomainFuncName` used for the functions which are not part of the domain but are closely related.

## FunctionTypeName
Beforehand the function type embedding just had a list of type names as their name.  `FunctionTypeName` was created to collect the type names of the argument as well as the return type name.

## HavocKotlinName and PredicateKotlinName
These two names inherited `KotlinName`. This is not a good fit, because the idea behind `KotlinName` are names that appear in the kotlin code. They were moved only inherit `SymbolicName`. As a consequence the field for the `KotlinName` in `ScopedKotlinName` was also updated to `SymbolicName`. Since the name is now no longer fitting, it was renamend to `ScopedName`. 

## TypeDomain
Before the functions used camelCase but the axioms used underscores. This was changed, such that all of them use camelCase.

# Various
- The commit message is a bit messy, I will squash them when merging.
- The implementation of `NamedEntityVisitor` uses a lot of extension methods. This is like this because the `SymbolicName` (as well as the `NamedEntity`) interface is defined in the `viper` package. And `plugin` package depends on the `viper` package. We might be able to resolve this, is it worth it?
- I asked codex to find programs that generate a name conflict. It came up with some tricky programs, but all of the issues could be identified and fixed. I added the tests in this branch, but I think we should remove them once we merge (there are many and they do not really bring a benefit for other situations).


jesyspa (commented):
This is again a very large PR, and I would prefer to split it.  Two thoughts:
- Do the new tests really add enough to be worth their maintenance burden?  It isn't quite clear to me why so many are needed.
- Is it possible to avoid flipping the logic directly?  It's very hard to check that this didn't break anything when it's changing so much.

PR #92, opened 2026-04-01
Moved shared-access predicate invariants to preconditions

pawinkler:
**Summary**:

This PR modifies the handling of shared-access predicate invariants on types. These are now being enforced by the preconditions of the surrounding method instead of being inhaled in it. Additionally, this required some changes to the unfolding policies.

**Context**:

Shared-access predicate invariants ensure that an access to an object on the heap is permitted. Previously, these were only inhaled before the access occurred and it was not verified that the individual function actually has enough permissions to access the object. Implementing this change demonstrated that functions expecting a super type of the provided type did not have enough permissions to access it, leading to failures in the test cases. 

**Changes**:

- We change this by reusing the existing `sharedPredicateAccessInvs` and wrapping them into preconditions instead of inhaling them. 
- We mitigate the permission issue by applying unfolding in the reverse subtyping order before making the function call. More details about this can be seen in the test case `multiple_interfaces`.
- To test if our linearizer handles this correctly, we added a new test case with subtyping, unfolding and assignments. Again, the previous policy was not strong enough. We mitigated this by unfolding the subtyping order again when we assigned a subtype to a location which expected a super type. 

**Limitations**:

A limitation of this PR is that it does not cover type casts. Currently, when we explicitly or implicitly cast to a subtype, the permissions of this subtype are assumed. Our current access predicate logic does not support an easy fix for this. 

**Tests**:

The expected output files were updated to expect the new permission handling and unfolding. No other changes were detected.

One test case was created to test subtyping with conditional blocks and assignments.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
@@ -13,8 +13,10 @@ import org.jetbrains.kotlin.formver.core.embeddings.callables.NamedFunctionSigna
 import org.jetbrains.kotlin.formver.core.embeddings.callables.toFuncApp
 import org.jetbrains.kotlin.formver.core.embeddings.callables.toMethodCall
 import org.jetbrains.kotlin.formver.core.embeddings.expression.debug.*
+import org.jetbrains.kotlin.formver.core.embeddings.types.ClassTypeEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.types.buildType
+import org.jetbrains.kotlin.formver.core.embeddings.types.predicateAccess
 import org.jetbrains.kotlin.formver.core.linearization.LinearizationContext
 import org.jetbrains.kotlin.formver.core.linearization.addLabel
 import org.jetbrains.kotlin.formver.core.linearization.freshAnonVar
@@ -191,11 +193,31 @@ data class NonDeterministically(val exp: ExpEmbedding) : UnitResultExpEmbedding,
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitNonDeterministically(this)
 }
 
+/** Unfolds every predicate in the subtyping hierarchy between the actual type and the expected supertype */
+private fun unfoldSubtypePredicatesForArgs(
+    args: List<ExpEmbedding>,
+    formalArgs: List<VariableEmbedding>,
+    ctx: LinearizationContext,
+) {
+    args.zip(formalArgs).forEach { (arg, param) ->
+        val innerPretype = arg.ignoringCastsAndMetaNodes().type.pretype
+        val paramPretype = param.type.pretype
+        if (innerPretype is ClassTypeEmbedding && paramPretype is ClassTypeEmbedding) {
+            val path = innerPretype.details.supertypePathTo(paramPretype) ?: return@forEach
+            for (step in path) { // unfold for each type in hierarchy
+                val predAcc = step.predicateAccess(arg, ctx.source)
+                ctx.addStatement { Stmt.Unfold(predAcc, ctx.source.asPosition) }
+            }
+        }
+    }
+}
+
 // Note: this is always a *real* Viper method call.
 data class MethodCall(val method: NamedFunctionSignature, val args: List<ExpEmbedding>) : StoredResultExpEmbedding {
     override val type: TypeEmbedding = method.callableType.returnType
 
     override fun toViperStoringIn(result: VariableEmbedding, ctx: LinearizationContext) {
+        unfoldSubtypePredicatesForArgs(args, method.formalArgs, ctx)
         ctx.addStatement {
             method.toMethodCall(
                 args.map { it.toViper(ctx) },
@@ -224,10 +246,13 @@ data class FunctionCall(val function: NamedFunctionSignature, val args: List<Exp
     override val subexpressions: List<ExpEmbedding>
         get() = args
 
-    override fun toViper(ctx: LinearizationContext): Exp = function.toFuncApp(
-        args.map { it.toViper(ctx) },
-        ctx.source.asPosition
-    )
+    override fun toViper(ctx: LinearizationContext): Exp {
+        unfoldSubtypePredicatesForArgs(args, function.formalArgs, ctx) // unfold before function call
+        return function.toFuncApp(
+            args.map { it.toViper(ctx) },
+            ctx.source.asPosition
+        )
+    }
 
     override fun <R> accept(v: ExpVisitor<R>): R =
         v.visitFunctionCall(this)
@@ -276,13 +301,6 @@ data class FunctionExp(
     override val type: TypeEmbedding = body.type
 
     override fun toViperMaybeStoringIn(result: VariableEmbedding?, ctx: LinearizationContext) {
-        signature?.formalArgs?.forEach { arg ->
-            // Ideally, we would want to assume these rather than inhale them to prevent inconsistencies with
-            // permissions. Unfortunately, Silicon for some reason does not allow Assumes.
-            listOfNotNull(arg.sharedPredicateAccessInvariant()).forEach { invariant ->
-                ctx.addStatement { Stmt.Inhale(invariant.toViperBuiltinType(ctx), ctx.source.asPosition) }
-            }
-        }
         body.toViperMaybeStoringIn(result, ctx)
         ctx.addLabel(returnLabel.toViper(ctx))
     }
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
@@ -19,7 +19,6 @@ import org.jetbrains.kotlin.formver.core.linearization.LinearizationContext
 import org.jetbrains.kotlin.formver.core.linearization.addLabel
 import org.jetbrains.kotlin.formver.core.linearization.freshAnonVar
 import org.jetbrains.kotlin.formver.core.linearization.pureToViper
-import org.jetbrains.kotlin.formver.core.linearization.unfoldSubtypePredicates
 import org.jetbrains.kotlin.formver.viper.NameResolver
 import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Exp
@@ -192,19 +191,11 @@ data class NonDeterministically(val exp: ExpEmbedding) : UnitResultExpEmbedding,
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitNonDeterministically(this)
 }
 
-private fun LinearizationContext.unfoldSubtypePredicatesForArgs(
-    args: List<ExpEmbedding>,
-    formalArgs: List<VariableEmbedding>,
-) {
-    args.zip(formalArgs).forEach { (arg, param) -> unfoldSubtypePredicates(arg, param.type, this) }
-}
-
 // Note: this is always a *real* Viper method call.
 data class MethodCall(val method: NamedFunctionSignature, val args: List<ExpEmbedding>) : StoredResultExpEmbedding {
     override val type: TypeEmbedding = method.callableType.returnType
 
     override fun toViperStoringIn(result: VariableEmbedding, ctx: LinearizationContext) {
-        ctx.unfoldSubtypePredicatesForArgs(args, method.formalArgs)
         ctx.addStatement {
             method.toMethodCall(
                 args.map { it.toViper(ctx) },
@@ -234,7 +225,6 @@ data class FunctionCall(val function: NamedFunctionSignature, val args: List<Exp
         get() = args
 
     override fun toViper(ctx: LinearizationContext): Exp {
-        ctx.unfoldSubtypePredicatesForArgs(args, function.formalArgs) // unfold before function call
         return function.toFuncApp(
             args.map { it.toViper(ctx) },
             ctx.source.asPosition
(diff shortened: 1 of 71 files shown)


jesyspa:
> Shared-access predicate invariants ensure that an access to an object on the heap is permitted. Previously, these were only inhaled before the access occurred and it was not verified that the individual function actually has enough permissions to access the object.

This is not true.  Shared predicates were always inhaled at the start of the function.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
```diff
@@ -191,11 +193,31 @@ data class NonDeterministically(val exp: ExpEmbedding) : UnitResultExpEmbedding,
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitNonDeterministically(this)
 }
 
+/** Unfolds every predicate in the subtyping hierarchy between the actual type and the expected supertype */
```
jesyspa: A function like this should already exist for accessing fields; please use the existing one.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
```diff
@@ -110,6 +110,16 @@ class ClassEmbeddingDetails(
 
     override fun subTypeInvariant(): TypeInvariantEmbedding = type.subTypeInvariant()
 
+    /** Constructs a list resembling the full subtyping relations between the current class and some supertype */
+    fun supertypePathTo(target: ClassTypeEmbedding): List<ClassTypeEmbedding>? {
+        if (type == target) return emptyList()
```
jesyspa, line 126: How is this different from hierarchyPathTo below?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
```diff
@@ -70,6 +70,17 @@ data class Linearizer(
         val lhsViper = lhs.toViper(this)
         val rhsViper = rhs.withType(lhs.type).toViper(this)
         addStatement { Stmt.assign(lhsViper, rhsViper, source.asPosition) }
+        // Unfold the subtype's shared predicate hierarchy so the supertype's shared predicate becomes available
```
jesyspa: This looks very similar to the function you have above.

formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/FirLightTreeFormVerPluginNoVerificationDiagnosticsTestGenerated.java
```diff
@@ -56,6 +56,12 @@ public void testAllFilesPresentInClasses() {
         KtTestUtil.assertAllTestsPresentByMetadataWithExcluded(this.getClass(), new File("formver.compiler-plugin/testData/diagnostics/verification/classes"), Pattern.compile("^(.+)\\.kt$"), null, true);
       }
 
+      @Test
+      @TestMetadata("conditional_subtype_passing.kt")
```
jesyspa, line 60: I don't see such a test in the list.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
pawinkler: After some investigation, they're not really different. I thought this was easier than modifying it. I've merged them now.

pawinkler:
> > Shared-access predicate invariants ensure that an access to an object on the heap is permitted. Previously, these were only inhaled before the access occurred and it was not verified that the individual function actually has enough permissions to access the object.
> 
> This is not true. Shared predicates were always inhaled at the start of the function.

True, this is a bit obscured. I meant that we inhale the permissions as soon as type information becomes available, e.g. by declaring it as an argument or receiver or by type casting them. This is also the reason why we have the limitations with the casts. If all predicates were available in the beginning of the function, handling them would be easy, but some inhales related to casts only occur later in the function body.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
pawinkler: I unified everything now (hopefully in a nice way).

formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/FirLightTreeFormVerPluginNoVerificationDiagnosticsTestGenerated.java
pawinkler: Sorry, I forgot to add it to git. It should now be available.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
pawinkler: I've updated this to use a shared function

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
```diff
@@ -191,11 +192,20 @@ data class NonDeterministically(val exp: ExpEmbedding) : UnitResultExpEmbedding,
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitNonDeterministically(this)
 }
 
+private fun unfoldSubtypePredicatesForArgs(
```
jesyspa: Please make this an extension method of `LinearizationContext` to make it clear that it's modifying the context.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
```diff
@@ -191,11 +192,20 @@ data class NonDeterministically(val exp: ExpEmbedding) : UnitResultExpEmbedding,
...
+    args: List<ExpEmbedding>,
+    formalArgs: List<VariableEmbedding>,
+    ctx: LinearizationContext,
+) {
+    args.zip(formalArgs).forEach { (arg, param) -> unfoldSubtypePredicates(arg, param.type, ctx) }
```
jesyspa: I am not a fan of this approach, where the use-site of an expression has to figure out what to do with it.  Is there no way to have the `arg` know what type it is being used as?  I would expect us to have this information in the ExpEmbedding, though maybe not for upcasts.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
```diff
@@ -224,10 +234,13 @@ data class FunctionCall(val function: NamedFunctionSignature, val args: List<Exp
     override val subexpressions: List<ExpEmbedding>
         get() = args
 
-    override fun toViper(ctx: LinearizationContext): Exp = function.toFuncApp(
-        args.map { it.toViper(ctx) },
-        ctx.source.asPosition
-    )
+    override fun toViper(ctx: LinearizationContext): Exp {
+        unfoldSubtypePredicatesForArgs(args, function.formalArgs, ctx) // unfold before function call
```
jesyspa: nit: comment is not helpful

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
```diff
@@ -16,7 +16,6 @@ import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 
 class ClassEmbeddingDetails(
     val type: ClassTypeEmbedding,
-    val isInterface: Boolean,
```
jesyspa, line 19: Why is this no longer needed?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
```diff
@@ -110,19 +109,48 @@ class ClassEmbeddingDetails(
...
-        if (className == type.name) {
-            yield(this@ClassEmbeddingDetails.type)
-        } else {
-            val sup = classSuperTypes.firstOrNull { !it.details.isInterface }
-                ?: throw IllegalArgumentException("Reached top of the hierarchy without finding the field")
-
-            yield(this@ClassEmbeddingDetails.type)
-            yieldAll(sup.details.hierarchyPathTo(field))
+    /**
+     * Returns the sequence of types to a class with access to [field] from the current type.
```
jesyspa: I don't understand what this sentence is supposed to say.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
```diff
@@ -110,19 +109,48 @@ class ClassEmbeddingDetails(
...
+    }
+
+    /**
+     * Returns a sequence of all types (including multiple hierarchies) to [target] from the current type.
+     */
+    fun hierarchyForTypeCast(target: ClassTypeEmbedding) = hierarchyPathTo(target)
+
+    /**
+     * Returns the sequence of types to unfold to reach [target] from the current type.
+     * When [includeTarget] is true (used for field access), the target class is appended.
```
jesyspa: Why is this the responsibility of `hierarchyPathTo`?  Shouldn't the caller know whether they want to unfold the class itself?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassEmbeddingDetails.kt
```diff
@@ -110,19 +109,48 @@ class ClassEmbeddingDetails(
...
+     * Returns an empty sequence if the current type already is [target].
+     * Throws if [throwOnFailure] is true and [target] is unreachable; otherwise returns empty.
+     */
+    private fun hierarchyPathTo(
+        target: ClassTypeEmbedding,
+        includeTarget: Boolean = false,
+        throwOnFailure: Boolean = false,
+    ): Sequence<ClassTypeEmbedding> {
+        val path = pathToOrNull(target)
+            ?: if (throwOnFailure) throw IllegalArgumentException("Reached top of the hierarchy without finding ${target.name}")
```
jesyspa: Why is throwing on failure not always the right choice?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
```diff
@@ -70,6 +70,9 @@ data class Linearizer(
...
+        val lhsWrapper = ExpWrapper(lhsViper, rhs.ignoringCastsAndMetaNodes().type)
+        unfoldSubtypePredicates(lhsWrapper, lhs.type, this)
```
jesyspa: It seems a bit strange to make a wrapper just to pass it into your own function.  Why not just pass in the type?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationContext.kt
```diff
@@ -68,4 +71,17 @@ fun LinearizationContext.freshAnonVar(init: TypeBuilder.() -> PretypeBuilder): A
 fun LinearizationContext.addLabel(label: Label) {
     addDeclaration(label.toDecl())
     addStatement { label.toStmt() }
+}
+
+/** Unfolds every predicate in the subtyping hierarchy between [actual]'s concrete type and [formalType]. */
+fun unfoldSubtypePredicates(actual: ExpEmbedding, formalType: TypeEmbedding, ctx: LinearizationContext) {
```
jesyspa: Similarly, please make ctx a receiver.

Though I really wonder: is cleaner than having separate functions that find the path and unfold it?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationContext.kt
```diff
@@ -68,4 +71,17 @@ fun LinearizationContext.freshAnonVar(init: TypeBuilder.() -> PretypeBuilder): A
...
+    val innerPretype = actual.ignoringCastsAndMetaNodes().type.pretype
+    val formalPretype = formalType.pretype
+    if (innerPretype is ClassTypeEmbedding && formalPretype is ClassTypeEmbedding) {
```
jesyspa: You ignore everything except the pretype here; will this work in the presence of nullability?

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
pawinkler: I've changed it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/ControlFlow.kt
pawinkler: I think this should be possible. If I understand correctly, were doing a similar thing already with inline functions. I can try to recreate it for all functions.

pawinkler:
I've realized that this approach is actually quite flawed. It does not properly work with nullability, as you pointed out above, and pure expressions, and fixing it with this approach is cumbersome and requires many small modifications. 

That's why I tried to implement your suggestion from above regarding the information in the embedding level, and I introduced a new embedding called Upcast, which is basically an inverted cast. We can detect this while encoding the program and then insert it into our AST. Then, we can nicely detect it in the different Linearizers and then either generate the unfoldings directly or re-use the SSA to place them in the correct location.

Note: This is not yet fully finished, but what do you think of this approach? And do you have a suggestion on how I could reuse the behavior of the PureFunBodyLinearizer in the PureExpLinearizer which doesn't include an SSA yet?

jesyspa:
Let's discuss during the meeting tomorrow.

pawinkler:
This PR was closed and transferred to three new drafts, which will implement the same goal in a more modular approach.

PR #93, opened 2026-04-02
Added support for Kotlin objects

pawinkler:
**Summary**:

This PR adds support for Kotlin objects to SnaKt and introduces new associated test cases. The translation is identical to the existing encoding of classes with an additional uniqueness constraint on object references.

**Context**:

SnaKt supports reasoning about classes, but not about objects. However, the reasoning about objects is very similar. Objects are singletons with global access, but the interactions with them are identical to the interactions with individual class instances. Thus, under the assumption that these interactions are correctly implemented for classes, we introduce additional uniqueness encoding for the reference of each objects and reuse the the shared access predicates from classes.

**Changes**:

- We add supporting elements to handle reasoning about objects, such as encodings for object literals.
- We extend the current translation to handle object reference and create the associates axioms.
- Shared access predicates for objects used in a function are now required by preconditions.
- We add a new test case verifying the implementation.

**Note**: The current handling for objects mirrors the handling (non-unique) class references passed to a function, which require the shared-access predicate. We thus treat used object the same as classes which are passed to a function as an argument and for which we do not have any additional information. This allows read, but no write operations.

**Tests**:

A new test case objects.kt was added.

**Note**: This PR depends on #92.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
@@ -415,6 +415,7 @@ class ProgramConverter(
                     addAll(it.pureInvariants())
                     addAll(it.accessInvariants())
                     addAll(it.provenInvariants())
+                    addIfNotNull(it.sharedPredicateAccessInvariant())
                     if (it.isUnique) {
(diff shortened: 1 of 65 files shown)


jesyspa (changes requested):
Not ready for review since PR it depends on hasn't been merged yet.

pawinkler:
This PR is being put on hold until a proper analysis on how we want to support Kotlin objects is completed.

PR #94, opened 2026-04-02
Adding support for explicit permission management in pre/post-conditions

### LINKED ISSUE #34: Add support for explicit permissions in pre-/postconditions
Until we have uniqueness/for power users who want more explicit control, it would be great to be able to write the following to control permissions explicitly:

```kotlin
class X(var a: Any)
fun f(x: X) {
  precondition { acc(x, write) }
  ...
}
```

The main difficulty is how this will work with automated permission management, we want to disable any kind of predicate-based access or automatic in-/exhaling for `a.x`.  Plus some design questions: Should we also disabling automation for other members of X? What about other base classes?  Maybe require marking the whole function with `@ManualPermissions`?

TheDeveloper101:
This PR adds the syntax for `acc` statements in preconditions, therefore allowing for explicit permissions management as mentioned in #34.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -235,6 +271,27 @@ fun StmtConversionContext.insertForAllFunctionCall(
...
+fun StmtConversionContext.insertAccFunctionCall(
+    field: FirPropertyAccessExpression,
+    perm: FirExpression,
+): ExpEmbedding {
+    val symbol = field.calleeReference.symbol as? FirPropertySymbol ?: throw SnaktInternalException(
+            field.source,
+            "acc requires a property access like x.a"
+        )
+    val fieldAccess = embedPropertyAccess(field)
+    val field = ((fieldAccess as ClassPropertyAccess).property.getter as BackingFieldGetter).field
```
The-Ray-Man: I think we should either add a comment why this cast will never fail or use safe casts and throw an error if the cast fails.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -307,8 +300,22 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
         }
+
+        functionCall.extractFormverFirStmt {  isInvariantBuilderFunctionNamed("acc") }?.let {
+            accLambda ->
+            //acc is only allowed to be used in the same context as forAll
+            if (!data.isValidForForAllBlock) throw SnaktInternalException(
+                accLambda.source,
+                "`acc` scope is only allowed inside one of the `loopInvariants`, `preconditions` or `postconditions`."
+            )
+            val fieldExpr = accLambda.arguments[0] as FirPropertyAccessExpression
```
The-Ray-Man: Let's use safe casts here as well. If a user types ``acc(1,write)`` they would just get a failed cast exception.
And we could also throw a good error when the user fails to use two arguments in the function. Or if they just write ``acc(x.f)`` then we use full permission for it (that is the default in viper)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -6,30 +6,66 @@
...
 import org.jetbrains.kotlin.formver.core.purity.checkValidity
 import org.jetbrains.kotlin.formver.core.purity.isPure
 import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+import org.jetbrains.kotlin.name.StandardClassIds
+import org.jetbrains.kotlin.text
 import org.jetbrains.kotlin.utils.addIfNotNull
 import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 import org.jetbrains.kotlin.utils.filterIsInstanceAnd
```
The-Ray-Man, line 34: There are many unused imports here

The-Ray-Man (changes requested):
I think that goes into a good direction. Made some remarks about error handling.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
TheDeveloper101: I like the idea of `acc(x.f)` defaulting to full permissions, so I will add that in

TheDeveloper101:
I was able to add in the error handling pretty easily. However, when I am trying to add the default value of `acc(x.f)` to be full permissions, I am getting a no value for parameter exception even when im able to generate the correct viper. any help would be appreciated

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
```diff
@@ -27,6 +27,14 @@ fun <T> postconditions(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T)
  * This class is designed as a receiver for lambda blocks of `loopInvariants`, `preconditions` and `postconditions`.
  */
 class InvariantBuilder {
+
+    class Permission
+
+    val read : Permission
+        get() = throw FormverFunctionCalledInRuntimeException("read.get")
+    val write : Permission
+        get() = throw FormverFunctionCalledInRuntimeException("write.get")
```
The-Ray-Man: I did not yet work with the `Builtins` logic yet, but I do not understand why we have the `read` and `write`?

TheDeveloper101: that was because we wanted to be able to embed the read and write parameters, so we needed to add them to the builtins since they’re not functions

TheDeveloper101 (commented):
This looks good :) - regarding the changes for the receiver and permissions

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -307,8 +301,29 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
+                accLambda.source,
+                "`acc` scope is only allowed inside one of the `loopInvariants`, `preconditions` or `postconditions`."
+            )
+            val fieldExpr = accLambda.arguments[0] as? FirPropertyAccessExpression ?: throw SnaktInternalException(
+                accLambda.source,
+                "must be a property access expression."
+            )
+            val permExpr = when(accLambda.arguments.getOrNull(1)?.source?.text?.toString() ?: "") {
+                "write" -> PermExp.WildcardPerm()
+                "read", "" -> PermExp.FullPerm()
```
The-Ray-Man: This is the wrong way around.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -235,6 +237,28 @@ fun StmtConversionContext.insertForAllFunctionCall(
     }
 }
 
+fun StmtConversionContext.insertAccFunctionCall(
+    field: FirPropertyAccessExpression,
+    perm: PermExp,
+): ExpEmbedding {
+    val symbol = field.calleeReference.symbol as? FirPropertySymbol ?: throw SnaktInternalException(
```
The-Ray-Man: We could remove the ``val symbol`` as it is unused. Do we even need this first check?

The-Ray-Man (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
TheDeveloper101: I dont think its needed anymore, its a remnant of the old way I had of embedding everything

TheDeveloper101:
I just pulled all of the recent changes and updated my copy, as well as made the changes requested, let me know if theres anything else to do for this to be merged :)

The-Ray-Man (changes requested):
Made some minor fixes. I think the code is good, but there is a remaining issue. 

Some tests are failing. I think they all failed because of the same reason. For some reason new viper methods are created. 

`heap_dependent_specifications.fir.diag.txt`
```viper
method g$public$read(this$dispatch: Ref) returns (ret: Ref)

method g$public$write(this$dispatch: Ref) returns (ret: Ref)
```

I am not sure why they are created, but I think they should not be.

TheDeveloper101 (commented):
this all looks good :) ill investigate the failing changes

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -241,6 +243,24 @@ fun StmtConversionContext.insertForAllFunctionCall(
...
+    perm: PermExp,
+): ExpEmbedding {
+    val fieldAccess = embedPropertyAccess(field)
+    val field = ((fieldAccess as? ClassPropertyAccess ?: throw SnaktInternalException(
+        field.source,
+        "could not embed as class property"
+    )).property.getter as? BackingFieldGetter ?: throw SnaktInternalException(
+        field.source,
+        "acc requires property access like x.a"
+    )).field
```
jesyspa: Make this a helper function please

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/FormverFirBlock.kt
```diff
@@ -22,6 +22,13 @@ fun FirStatement.extractFormverFirBlock(predicate: FirFunctionSymbol<*>.() -> Bo
     return formverInvariantsArgument.anonymousFunction
 }
 
+fun FirStatement.extractFormverFirStmt(predicate: FirFunctionSymbol<*>.() -> Boolean): FirFunctionCall? {
+    if (this !is FirFunctionCall) return null
+    val firFunction = toResolvedCallableSymbol() as? FirFunctionSymbol<*> ?: return null
+    if (!predicate(firFunction)) return null
```
jesyspa: There's some duplicate code with the above, can it be done in a cleaner way?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -307,8 +298,30 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
+            //acc is only allowed to be used in the same context as forAll
+            if (!data.isValidForForAllBlock) throw SnaktInternalException(
+                accLambda.source,
+                "`acc` scope is only allowed inside one of the `loopInvariants`, `preconditions` or `postconditions`."
+            )
+            val fieldExpr = accLambda.arguments[0] as? FirPropertyAccessExpression ?: throw SnaktInternalException(
+                accLambda.source,
+                "must be a property access expression."
+            )
+            val permExpr = when (val perm = accLambda.arguments.getOrNull(1)?.source?.text?.toString()) {
```
jesyspa: This access into the source is kinda hacky.  Can  `isFormVerFunctionNamed` be used here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -307,8 +298,30 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
                     forAllLambda.body?.source, "Lambda body should be accessible in `forAll` function call."
                 )
                 return data.insertForAllFunctionCall(forAllArg.symbol, forAllBody)
+        }
+
+        functionCall.extractFormverFirStmt {  isInvariantBuilderFunctionNamed("acc") }?.let {
+            accLambda ->
+            //acc is only allowed to be used in the same context as forAll
+            if (!data.isValidForForAllBlock) throw SnaktInternalException(
```
jesyspa: Probably better to rename the property

jesyspa (changes requested):
Some thoughts, generally looks reasonable. :) The new functions are expected, I suspect when you regenerate the goldens for your test they'll appear there

TheDeveloper101:
so sorry for getting back to this late! ive been really busy with school lately :( unfortunately due to issues with my viper install, im not able to run and check which tests are failing, so if someone can update the relevant tests so that they no longer fail I would be extremely grateful

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
TheDeveloper101: I dont think that quite works because we are indexing into the argument list of a FirFunctionCall so each argument is a FirExpression which can't be cast into a FirFunctionSymbol

The-Ray-Man:
Hi
I updated the expected results. I also rebased the branch on the up to date main

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -37,6 +38,17 @@ val SpecialKotlinFunction.callableId: CallableId
 
 fun SpecialKotlinFunction.embedName(): ScopedName = callableId.embedFunctionName(callableType)
 
+/**
+ * Strips type casts, meta nodes, and type-invariant wrappers to recover the core embedding
+ * an expression ultimately produces. Arguments of a special function call are wrapped with
+ * the parameter's type invariants, so we have to peel those away to inspect the real payload.
+ */
+private fun ExpEmbedding.unwrapToCore(): ExpEmbedding =
```
jesyspa: InhaleInvariants is really a Cast node, so should also be stripped by `ignoringCastsAndMetaNodes`

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -211,6 +223,38 @@ object SpecialKotlinFunctions {
             ctx.insertForAllFunctionCall(param.symbol, body)
         }
 
+        val permissionCallableType = buildFunctionPretype {
+            withReturnType { nullableAny() }
+        }
+        withCallableType(permissionCallableType) {
+            addFunction(SpecialPackages.formver, name = "read") { _, _ -> PermissionLit(PermExp.WildcardPerm()) }
+            addFunction(SpecialPackages.formver, name = "write") { _, _ -> PermissionLit(PermExp.FullPerm()) }
```
jesyspa, line 220: Why not just construct a permission?

jesyspa (changes requested):

jesyspa (approved):
Looks good to me, test churn is from the update to InhaleInvariants.

jesyspa:
@The-Ray-Man I'm biased since I was involved in the development, could you also review please?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -13,12 +13,20 @@ import org.jetbrains.kotlin.formver.viper.ast.PermExp
...
 ) : ExpEmbedding {
     override val type: TypeEmbedding
         get() = buildType { boolean() }
 
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitAccEmbedding(this)
-    override fun children(): Sequence<ExpEmbedding> = sequenceOf(access)
+    override fun children(): Sequence<ExpEmbedding> = sequenceOf(receiver)
+}
+
+data class PermissionLit(val perm: PermExp) : ExpEmbedding {
```
The-Ray-Man, line 26: I think it is not ideal to mix ``ExpEmbeddings`` together with `Viper` structures. But since it is a 1 to 1 mapping it's not a big issue.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
```diff
@@ -13,12 +13,20 @@ import org.jetbrains.kotlin.formver.viper.ast.PermExp
 
 data class AccEmbedding(
     val field: FieldEmbedding,
-    val access: ExpEmbedding,
+    val receiver: ExpEmbedding,
```
The-Ray-Man: nit: can we change the order such that ``receiver`` comes before `field`? This is inconsistent with the other field access classes and feels unintuitive.

The-Ray-Man (approved):
Just minor things. I think I would have done it very similar.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
jesyspa: Yeah, I looked at the options here but I do think this is the most natural way to express this.  We want the arguments to be translated to ExpEmbedding.

The-Ray-Man: I think we might have a misunderstanding. I think the data class `PermissionLit` is fine, but it's property the `PermExp` which imo is not ideal. Because the `PermExp` comes from the `Viper` IR. I think generally we don't want to have our IRs mixed.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
jesyspa: Done, swapped to `receiver, field, perm`.

TheDeveloper101:
i think this is good to merge now right?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
TheDeveloper101: thank you!

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/AccEmbedding.kt
jesyspa: I think in general that's true, but here I think it's okay, since the goal here is to represent a literal (and the set of valid ones comes from Viper)

PR #95, opened 2026-04-07
Upgrading Viper to the newest version as of April 2026

pawinkler:
**Summary**:

This PR introduces new Gradle options to allow users to switch between their used version of the Viper back-end and to select a local M2 version of Viper. Additionally, it updates the code base to use the newest version available. This required the adaption of three breaking changes, see below.

**Changes**:

- Added support for two new Gradle options allowing the selection of a Viper snapshot and to use the local M2 repository.
- Implemented a new method for Viper's position which was required by the updated version.
- Changed the type of the perm argument in AccessPredicates from `Exp` to `Option<Exp>`, as required by the updated version.
- Updated the test cases to the new symbol for integer division `\`.

**Note**: This PR updates the used Viper version. Our internal version number is `2026-04-10_b005beb`, which consists of the published date and used Viper commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,15 +138,6 @@ You need to (additionally) set `Z3_EXE` in `~/.xprofile` and/or
 `~/.bash_profile` depending on your shell, window manager, display
 manager, operating system, etc.
 
-## Gradle Properties
-
-Two properties in `gradle.properties` control which Silicon artifact is used:
-
-- `viperVersion`: the version of `viper:silicon_2.13` to depend on (e.g. `1.1-SNAPSHOT`).
-- `useLocalViper`: set to `true` to resolve Silicon from the local Maven repository (`~/.m2`)
-  instead of the JetBrains Space Maven repository. Useful when testing a locally built Silicon
-  via `sbt publishM2`.
-
 ## Running tests
 
 The test pipeline for each function does conversion (Kotlin FIR → Viper AST),
--- a/README.md
+++ b/README.md
@@ -138,6 +138,15 @@ You need to (additionally) set `Z3_EXE` in `~/.xprofile` and/or
 `~/.bash_profile` depending on your shell, window manager, display
 manager, operating system, etc.
 
+## Gradle Properties
+
+Two properties in `gradle.properties` control which Silicon artifact is used:
+
+- `viperVersion`: the version of `viper:silicon_2.13` to depend on (e.g. `1.1-SNAPSHOT`).
+- `useLocalViper`: set to `true` to resolve Silicon from the local Maven repository (`~/.m2`)
+  instead of the JetBrains Space Maven repository. Useful when testing a locally built Silicon
+  via `sbt publishM2`.
+
 ## Running tests
 
 The test pipeline for each function does conversion (Kotlin FIR → Viper AST),
--- a/formver.compiler-plugin/core/build.gradle.kts
+++ b/formver.compiler-plugin/core/build.gradle.kts
@@ -8,7 +8,7 @@ dependencies {
     compileOnly(kotlin("compiler"))
 
     // TODO: figure out how to avoid this dependency
-    compileOnly("viper:silicon_2.13:1.3-SNAPSHOT")
+    compileOnly(ViperVersions.silicon)
 }
 
 sourceSets {
--- a/formver.compiler-plugin/core/build.gradle.kts
+++ b/formver.compiler-plugin/core/build.gradle.kts
@@ -8,7 +8,7 @@ dependencies {
     compileOnly(kotlin("compiler"))
 
     // TODO: figure out how to avoid this dependency
-    compileOnly("viper:silicon_2.13:${project.property("viperVersion")}")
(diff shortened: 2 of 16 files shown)


The-Ray-Man (commented):
I think this looks good. Let’s merge this once Viper 1.3 is available through maven.

pawinkler:
Thanks! Everything should be ready now.

gradle.properties
```diff
@@ -1,3 +1,4 @@
+viperVersion=1.3-SNAPSHOT
```
jesyspa: AFAIK this is usually stored in `buildSrc/settings.gradle.kts`, please `gradle init` a fresh project and take a look at how it's set up: https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html

README.md
```diff
@@ -138,6 +138,15 @@ You need to (additionally) set `Z3_EXE` in `~/.xprofile` and/or
 `~/.bash_profile` depending on your shell, window manager, display
 manager, operating system, etc.
 
+## Gradle Properties
+
+Two properties in `gradle.properties` control which Silicon artifact is used:
+
+- `viperVersion`: the version of `viper:silicon_2.13` to depend on (e.g. `1.1-SNAPSHOT`).
+- `useLocalViper`: set to `true` to resolve Silicon from the local Maven repository (`~/.m2`)
```
jesyspa: I don't really get the motivation for this.  If you want to use maven local, you would generally just add mavenLocal and bump the viper version.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Position.kt
```diff
@@ -13,6 +13,7 @@ sealed class Position : IntoSilver<viper.silver.ast.Position> {
 
     internal class Wrapper<P>(val wrappedValue: P) : viper.silver.ast.Position {
         override fun toString(): String = "<wrapped value>"
+        override fun deltaColumn(delta: Int): viper.silver.ast.Position = this
```
jesyspa, line 21: Please add a comment on what this is.

jesyspa (changes requested):
Thanks for updating, some comments.

formver.compiler-plugin/core/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ dependencies {
     compileOnly(kotlin("compiler"))
 
     // TODO: figure out how to avoid this dependency
-    compileOnly("viper:silicon_2.13:1.2-SNAPSHOT")
+    compileOnly("viper:silicon_2.13:${project.property("viperVersion")}")
```
jesyspa: By the way, is there a reason to stay on silicon_2.13?

gradle.properties
pawinkler: I've reverted to the default setup now.

README.md
pawinkler: We needed to update the version at three different locations in Gradle. That's why I thought it would be nicer to have a general management for it.

formver.compiler-plugin/core/build.gradle.kts
pawinkler: Afaik, Silicon doesn't have version numbers. 2.13 describes the Scala version used for Silicon and no never version is available.

README.md
jesyspa: But `mavenLocal()` only needs to be updated in one place, plus one update to `buildSrc`, no?

pawinkler: True, but usually the local version numbers don't align with the remote ones. So we need to update the `mavenLocal()` and the Viper version in the three `build.gradle.kts` files.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Position.kt
pawinkler: I've added a comment.

README.md
jesyspa: Sorry, I don't follow.  I would expect the change to require changing two places:
- Add `mavenLocal()` to `settings.gradle.kts`
- Modify the viper version in `buildSrc/settings.gradle.kts`

I think the consolidation of the Viper version is a good idea, just not the `useLocalViper` flag.

gradle.properties
jesyspa: Is there something about the recommended approach with `buildSrc` that doesn't work?

README.md
pawinkler: You're right, the local change itself affects only these two, the remaining ones were only necessary because the local versions are not aligned with the remote ones. If we keep the consolidated Viper version, this is resolved.

jesyspa (changes requested):
As discussed in the comments, please move version configuration to `buildSrc`.

pawinkler:
I've updated the version number to date+commit, updated the description and published it again.

jesyspa (approved):

PR #96, opened 2026-04-08
Migrating to Phased Test Infrastructure

The-Ray-Man:
# Migrating to Phased Test Infrastructure
The goal of this PR is to improve the development experience.  For this the conversion and verification was split. 

| Gradle Task | Mode | Conversion | Verification | Primary Use Case |
| :--- | :--- | :--- | :--- | :--- |
| `test` | **`FULL`** | Runs for every test | Runs for every test | **CI/CD:** Ensures total consistency in automated pipelines. |
| `update` | **`UPDATE`** | Runs for every test | Runs only if conversion changed | **Manual Review:** Validates that code changes produced the expected output. |
| `untilConversion` | **`CHECK_CONVERSION`** | Runs for every test | Never runs | **Refactoring:** Confirms that logic changes didn't accidentally break the output. |

## Summary
To be able to compare the conversion and verification results the diagnostics  were split into two golden files. For verification only errors are added to the golden files. Meaning that, if no golden file exists then either the corresponding tests has no verification targets or everything successfully verifies.

The testing pipeline has two new services:
- `diagnosticsCollector`: This service collects the diagnostics and separates the verification and conversions. It can assert the equality to each golden file.
- `tagCollector`: This service collects the tags, which are added to the `.kt` file. It can assert the equality to the golden file, as well as assert the equality of only a category (e.g. conversion) of tags. 

Execution of the pipeline:
1. The plugin is registered as a checker. It generates the viper program and attaches them to the function declaration. Also it attaches if the function should be verified according to the directives and annotations. This is done, because this information is not available afterwards.
2. The `AfterConversionHandler`, reports the diagnostics and if necessary asserts the equality of the conversion (in `CHECK_CONVERSION` mode)
3. The verification facade extracts the converted programs and if necessary verifies them. The result is then checked agains the golden files.

# Todos
- [x] If we decide this is a good approach, we should remove the old testing infrastructure. And update the Github action.
- [x] The gradle tasks at the moment are a bit a mess. Especially, how I add the `testMode`.
- [x] The modifications might break the plugin when we actually want to use it.


formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
```diff
@@ -79,24 +82,22 @@ class ViperPoweredDeclarationChecker(private val session: FirSession, private va
...
+
             val viperProgram = with(programConversionContext.nameResolver) { program.toSilver() }
+            declaration.viperProgram = viperProgram
+            declaration.shouldVerify = config.shouldVerify(declaration)
+
             val onFailure = { err: VerifierError ->
-                val source = err.position.unwrapOr { declaration.source }
-                reporter.reportVerifierError(source, err, config.errorStyle)
+                val source = err.position.unwrapOr { declaration.source }!!
+                // We report consistency errors in viper as internal errors, because if this happens our plugin is wrong.
```
ligee: The former comment is more understandable.

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/runners/PhasedTests.kt
```diff
@@ -0,0 +1,331 @@
...
+import org.jetbrains.kotlin.test.frontend.fir.FirOutputArtifact
+import org.jetbrains.kotlin.test.frontend.fir.handlers.*
+import org.jetbrains.kotlin.test.model.*
+import org.jetbrains.kotlin.test.runners.AbstractPhasedJvmDiagnosticLightTreeTest
+import org.jetbrains.kotlin.test.services.*
+import org.jetbrains.kotlin.test.utils.MultiModuleInfoDumper
+import org.jetbrains.kotlin.util.capitalizeDecapitalize.toLowerCaseAsciiOnly
+
+
+abstract class AbstractPhasedDiagnosticTest() : AbstractPhasedJvmDiagnosticLightTreeTest() {
```
ligee: Seems you're not using anything from `AbstractPhasedJvmDiagnosticLightTreeTest` besides the `parser` parameter.
You shoul probably inherit the `AbstractKotlinCompilerWithTargetBackendTest(TargetBackend.JVM_IR)` instead.

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/runners/PhasedTests.kt
```diff
@@ -0,0 +1,331 @@
...
+    override fun configure(builder: TestConfigurationBuilder) = with(builder) {
+        defaultDirectives {
+            LATEST_PHASE_IN_PIPELINE with TestPhase.FRONTEND
+            +ENABLE_PLUGIN_PHASES
+            +RENDER_DIAGNOSTICS_FULL_TEXT
+            LANGUAGE with "+PropertyParamAnnotationDefaultTargetMode"
+            +JvmEnvironmentConfigurationDirectives.FULL_JDK
```
ligee: Is `FULL_JDK` needed for all tests? If not, I'd suggest to remove it and add directives to the tests that do need it. The tests with the full jdk is quite a bit heavier.

ligee (changes requested):

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/runners/PhasedTests.kt
The-Ray-Man: Yes that is true. Only 58 of 150 tests need it.

formver.compiler-plugin/testData/diagnostics/conversion/classes/inheritance.kt
```diff
@@ -1,3 +1,4 @@
+// FULL_JDK
 // NEVER_VALIDATE
```
ligee, line 1: I skimmed through the files where you put the  `FULL_SDK` directive, and in most of them, I do not understand, why the full sdk is required. If you know a generic answer, please put it into the appropriate commit message. If not, i.e. there are some errors when you try to run the tests without the directive, please create an YT issue to review this directive use in the future.

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/GenerateTests.kt
```diff
@@ -5,21 +5,23 @@
...
-import org.jetbrains.kotlin.formver.plugin.runners.AbstractFirLightTreeFormVerPluginNoVerificationDiagnosticsTest
-import org.jetbrains.kotlin.formver.plugin.runners.AbstractFirLightTreeFormVerPluginDiagnosticsTest
+import org.jetbrains.kotlin.formver.plugin.runners.AbstractPhasedDiagnosticTest
 import org.jetbrains.kotlin.generators.dsl.junit5.generateTestGroupSuiteWithJUnit5
 
 fun main() {
     generateTestGroupSuiteWithJUnit5 {
         testGroup(testDataRoot = "formver.compiler-plugin/testData", testsRoot = "formver.compiler-plugin/test-gen") {
-            testClass<AbstractFirLightTreeFormVerPluginDiagnosticsTest> {
+//            testClass<AbstractFirLightTreeFormVerPluginDiagnosticsTest> {
```
ligee: Is it the old ifra you want to delete, if this approach works? Then please put a TODO comment on top of the commentexd code.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
```diff
@@ -131,7 +127,7 @@ class ViperPoweredDeclarationChecker(private val session: FirSession, private va
     }
 
     private fun PluginConfiguration.shouldVerify(declaration: FirSimpleFunction): Boolean = when {
-        verificationSelection == TargetsSelection.FORCE_DISABLE -> false
+//        verificationSelection == TargetsSelection.FORCE_DISABLE -> false
```
ligee: Please, do not leave commented out code without an explanation comment.

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/services/TagCollector.kt
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        val actualText = builder.toString()
+        return actualText
+    }
+
+    fun sourceFileWithoutTags(file: TestFile): String {
+        val content = testServices.sourceFileProvider.getContentOfSourceFile(file)
+        return content.replace(
+            Regex("\"<![A-Z_]*!>(.*?)<!>\"")
```
ligee: I guess you probably want to use `clearTextFromDiagnosticMarkup` here.

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/services/VerificationFacade.kt
```diff
@@ -0,0 +1,200 @@
...
+ * If necessary, asserts the verification result as well as the conversion result.
+ */
+class ViperResultHandler(testServices: TestServices) :
+    AnalysisHandler<FirOutputArtifact>(testServices, false, false) {
+    override val artifactKind: TestArtifactKind<FirOutputArtifact> get() = FrontendKinds.FIR
+
+    override fun processModule(module: TestModule, info: FirOutputArtifact) {
+        val shouldAssert = when (getTestMode()) {
+            TestMode.CHECK_CONVERSION -> false // in this mode we never want to verify.
+            TestMode.UPDATE -> testServices.diagnosticsCollector.conversionHasChanged()
```
ligee: It's better to extract this `when` into a separate utility fn, and replace both places. Otherwise, it will be too easy to change it in one place and forget in another.

formver.compiler-plugin/build.gradle.kts
```diff
@@ -123,6 +127,36 @@ tasks.register<Test>("testBothModes") {
     systemProperty("formver.conversionOnly", "false")
 }
 
+tasks.register<Test>("PhasedFULL") {
+    description = "Runs until verification"
```
ligee: Unconventional task naming. Maybe it is better to stick to regular snakeCase ones.

ligee (changes requested):
You also state in TODO part that the actual verification could be broken with these changes. What do you mean?

formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/services/TagCollector.kt
The-Ray-Man: yes, thanks :)

The-Ray-Man:
> You also state in TODO part that the actual verification could be broken with these changes. What do you mean?

My changes might break the actual usability of the plugin. So when someone (a user of the plugin, not developer) wants to verify their code, then the plugin will be called somehow (I don't know how exactly). And if just the `check` method of the plugin is executed, then nothing will be actually verified because the verification happens now in the verification facade.

formver.compiler-plugin/testData/diagnostics/conversion/classes/inheritance.kt
The-Ray-Man: I am not really sure what the reason is. But I think it has something to do with function arguments. I think if we use a type already defined (like Int) and not one of our own (like Baz) we get this issue.

For example, in this file, when removing the ``FULL_JDK``, I get this error:

``warning: Cannot access 'Serializable' which is a supertype of 'Int' or one of its arguments. This may be forbidden soon. Check the module classpath for missing or conflicting dependencies.``

Is this intended behavior? What do you think?

ligee: I find it quite puzzling. I expect serializable to be present in the emulated JDK as well. So, I suggest creating an issue to investigate it at some point.

The-Ray-Man: I dived a bit deeper and I think the issue was not the function arguments but the performed addition. I could come up with a very simple example where the issue also appears. I created a [YT issue](https://youtrack.jetbrains.com/issue/KT-85928/FULLJDK-Directive-Unexpected-Behaviour).

ligee (approved):

PR #99, opened 2026-04-09
Type Domain Name Style

The-Ray-Man:
This PR updates the axiom names to use camelCase like the domain functions. We do this, because it will allow us to have a better short name resolver.

Before:
```
// function uses camelCase
function df$rt$stringFromRef(r: Ref): Seq[Int]

// axiom does not
axiom rt$subtype_reflexive {
    (forall t: d$rt ::df$rt$isSubtype(t, t))
}
```

Now:
```
function df$rt$stringFromRef(r: Ref): Seq[Int]

axiom rt$subtypeReflexive {
    (forall t: d$rt ::df$rt$isSubtype(t, t))
}
```


pawinkler (approved):
Looks good!

PR #101, opened 2026-04-09
Name Registration

The-Ray-Man:
This PR should not be merged before #100 

Previously, not all names appearing in the program were registered with the resolver. This was not an issue for the `SimpleResolver` because it operates independently of registered names. However, future resolver implementations will require every name to be registered, and this PR prepares the codebase for that requirement.

Some registration passes included here may be redundant. For example, when registering a while loop, the process also traverses the condition. It is generally impossible for a condition to contain a fresh name, making that specific pass unnecessary.

I suggest retaining these passes regardless. The computational overhead is minimal, and keeping them ensures a more robust guarantee that all names are registered without exception.


jesyspa (approved):
Agreed that it's better to register these things explicitly even if it's not possible to have new names.

PR #103, opened 2026-04-09
Name Hierarchy Update

The-Ray-Man:
This PR refactors some part of the name hierarchy. This are the changes and the consequences:

**Change**: 
`SpecialName` was split into `DomainAssociatedFuncName` and `SpecialFieldName`. 
**Consequence**: 
- The domain associated functions (functions that are closely related to the domain but not actually part of the domain e.g. addInt) are now prefixed with `df` (domain-function) instead of `sp` (**sp**ecial)
- Special fields (at the moment just size) is prefixed with `p` (property) instead of `sp`

**Change**: 
`PredicateKotlinName` was renamed to `PredicateName` and does no longer inherit `KotlinName` but `FreshName`. The same was done for `HavocKotlinName`.
**Consequence**: 
The `ScopedKotlinName` does now hold a `SymbolicName` instead of a `KotlinName`. Also it was renamed to just `ScopedName`

**Change**: 
The file `FreshNames` was renamed to `FreshName` to be consistent with `KotlinName`
**Consequence**:  
None

**Change**:  
`NamedEntity` was introduced as a top level interface for `SymbolicName`, `FreshName`, `ScopeName`, and `NameType`. This is also for preparation such that a future name resolver can collect all these things in a single collection.
**Consequence**:  
None, all subtype relation which hold previously still hold.

**Change**:
`SetOfNames` was changed to `ListOfNames` and made generic. 
**Consequence**:
None

**Change**:
Instead of giving function types just a list of names as name we introduce `FunctionTypeName`  to give functions an actual name
**Consequence**:
None


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/KotlinName.kt
```diff
@@ -117,7 +101,17 @@ data class SetOfNames(val names: List<SymbolicName>) : KotlinName {
...
+    context(nameResolver: NameResolver)
+    override val mangledBaseName: String
+        get() = buildString {
+            append(args.mangledBaseName)
+            if (isNotEmpty()) append(SEPARATOR)
+            append(returns.mangled)
+        }
+}
+
+data class TypeName(val pretype: PretypeEmbedding, val nullable: Boolean) : NameOfType {
```
jesyspa: Why have both a NameOfType and a TypeName?  IMHO it's sort of confusing

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SymbolicName.kt
```diff
@@ -5,14 +5,21 @@
 
 package org.jetbrains.kotlin.formver.viper
 
+/**
+ * Interface to unify all structures that are names to some entity.
+ */
+interface NamedEntity
```
jesyspa: I don't quite agree with the naming here: `NamedEntity` sounds like an entity with a name.  Maybe `AnyName`?

jesyspa (changes requested):
Generally looks good

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/KotlinName.kt
The-Ray-Man: The idea was to have `NameOfType` as a common supertype of all names that can refer to a type. Such that, for example, the `FunctionTypeName` could collect the arguments in a list, where the elements have type `NameOfType`. 
But yes, you are right, this is not necessary, because all the types that can occur as arguments etc. will be ´TypeName´. So we can just use this for such situations.

jesyspa (approved):

PR #108, opened 2026-04-10
Updated verification backend to new SiliconFrontendAPI

pawinkler:
## Summary

This PR updates the internally used verifier class from the Silicon-internal `DefaultMainVerifier` to the official `SiliconFrontendAPI`. With the new API, Silicon attempts verification using a two-phase approach consisting of consistency checks and verification. Consequently, we cannot check consistency in advance anymore and instead differentiate between different error-types. The overall behavior remains unaffected.

> The new `SiliconFrontendAPI` is only available in newer Silicon versions. It can only be merged after #95.

## Changes 

- Removed old `Verifier.kt`.
- Added new `SiliconFrontend.kt` as a wrapper around the `SiliconFrontendAPI`.
- Updated usages of the verifier to the new `SiliconFrontend`.


jesyspa (changes requested):
Please rebase onto latest main, many of these changes are already merged.

pawinkler:
The direct rebase on main was not possible, I had to cherry pick the actual changes and force-push them. I don't really understand why I couldn't rebase, but it introduced many more changes which shouldn't exist.

pawinkler:
One test case was not yet updated to the new name handling and resulted in failing checks.

jesyspa:
> One test case was not yet updated to the new name handling and resulted in failing checks.

Open a separate PR for this please since it's blocking other people's work.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SiliconFrontend.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /** Consistency-checks and verifies [viperProgram], calling [onFailure] for each error found. */
+    fun verify(viperProgram: viper.silver.ast.Program, onFailure: (VerifierError) -> Unit) {
+        val result = siliconApi.verify(viperProgram)
+        if (result is viper.silver.verifier.Failure) {
+            for (error in result.errors()) {
+                when (error) {
+                    is viper.silver.verifier.VerificationError ->
+                        onFailure(VerificationError.fromSilver(error))
```
jesyspa: Why use `fromSilver` here rather than just invoke the constructor directly?

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SiliconFrontend.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.kotlin.formver.viper.errors.GenericConsistencyError
+import org.jetbrains.kotlin.formver.viper.errors.VerificationError
+import org.jetbrains.kotlin.formver.viper.errors.VerifierError
+import viper.silver.reporter.StdIOReporter
+
+/**
+ * Passes Viper programs for verification to the Silicon backend via [viper.silicon.SiliconFrontendAPI].
+ * Use [SiliconFrontend.verify] to consistency-check and verify a given program.
+ */
+class SiliconFrontend(commandLineArgs: List<String>) {
```
jesyspa: I think it would make sense for this class to implement `Closeable`

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SiliconFrontend.kt
```diff
@@ -0,0 +1,56 @@
...
+            System.getenv("SILICON_PARALLEL_VERIFIERS")?.let {
+                add("--numberOfParallelVerifiers")
+                add(it)
+            }
+        }
+        siliconApi = viper.silicon.SiliconFrontendAPI(StdIOReporter("stdout_reporter", true))
+        siliconApi.initialize(args.toScalaSeq())
+    }
+
+    companion object {
```
jesyspa, line 28: Is this a reasonable solution?  Will the logging itself be thread-safe, or should we perhaps create a logger per thread?

I think it makes sense to check this with Malte during our meeting tomorrow.

jesyspa (changes requested):

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SiliconFrontend.kt
pawinkler: I've changed it, it was a relict from changing between the two approaches.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SiliconFrontend.kt
pawinkler: I agree. This seems to be a somewhat known problem:
- https://www.slf4j.org/codes.html#substituteLogger
- https://stackoverflow.com/questions/77924065/java-unit-test-error-class-org-slf4j-helpers-substitutelogger-cannot-be-cast-to
All fixes I've seen so far delay the usage of the logger, either by looping or blocking the resource. But it seems to be a setup problem, which doesn't affect general thread safety.

jesyspa (approved):

PR #109, opened 2026-04-11
Refactor linearization: extract visitor pattern, clean up type casts and position handling

jesyspa:
## Summary
- Extract all linearization logic (`toViper`, `toViperStoringIn`, etc.) from `ExpEmbedding` subclasses into a `LinearizationVisitor`, using the existing `ExpVisitor` infrastructure. Concrete `ExpEmbedding` types are now pure data holders.
- Simplify the `Linearizable` hierarchy: remove `toViperSideEffects`, merge `NoResultLinearizable` into `UnitResultLinearizable`.
- Move `withType()` casts from linearization time to embedding construction sites (`Assign`, `Declare`, `Return`, `If` branches, `FieldModification`), so type correctness is established earlier in the pipeline.
- Change `LinearizationContext` methods to accept `Linearizable` instead of `ExpEmbedding`, eliminating redundant `toLinearizable()` re-entry that lost visitor state.
- Decompose `addFieldAccessStoringIn` to take receiver and field separately, removing the need to re-linearize `FieldAccess` embeddings.
- Add `source` parameter to `toLinearizable()` for position propagation; capture the `LinearizationVisitor` in base `Linearizable` classes for child iteration.
- Extract `DebugTreeViewVisitor` using the same visitor pattern; add `children()` to all `ExpEmbedding` types, remove `subexpressions`.
- Convert `AccEmbedding`, `ForAllEmbedding`, `SequentialAnd`, `SequentialOr` to data classes.

## Test plan
- [x] All existing tests pass
- [ ] Review that linearization behavior is preserved for edge cases (unit-returning functions, InhaleInvariants, InvokeFunctionObject)

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizable.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizable.kt
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2010-2025 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.formver.core.linearization
+
+import org.jetbrains.kotlin.KtSourceElement
+import org.jetbrains.kotlin.formver.core.asPosition
+import org.jetbrains.kotlin.formver.core.domains.RuntimeTypeDomain
+import org.jetbrains.kotlin.formver.core.embeddings.SourceRole
+import org.jetbrains.kotlin.formver.core.embeddings.asInfo
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ToViperBuiltinOnlyError
+import org.jetbrains.kotlin.formver.core.embeddings.expression.VariableEmbedding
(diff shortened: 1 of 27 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizable.kt
```diff
@@ -0,0 +1,163 @@
...
+
+interface Linearizable {
+    fun toViper(ctx: LinearizationContext): Exp
+    fun toViperStoringIn(result: VariableEmbedding, ctx: LinearizationContext)
+    fun toViperMaybeStoringIn(result: VariableEmbedding?, ctx: LinearizationContext)
+    fun toViperBuiltinType(ctx: LinearizationContext): Exp
+    fun toViperUnusedResult(ctx: LinearizationContext)
+}
+
+fun ExpEmbedding.toLinearizable(source: KtSourceElement? = null): Linearizable = accept(LinearizationVisitor(source))
```
The-Ray-Man, line 28: This might be personal preference, but I think we should move this function to the ``ExpEmbedding``. This is a rather fundamental function for the `ExpEmbedding` and I think it is better to not have them as extension functions.

jesyspa: Generally, the rule is:
- If implementations should be able to override it, put it in the interface.
- If it should be the same for all implementations, make it an extension.

The-Ray-Man (approved):

PR #114, opened 2026-04-13
Name Resolution Moved Into Resolver

The-Ray-Man:
In this PR we remove all the logic from the name classes and move them into the resolver. This also includes predefined strings for names e.g. "public" for the public scope.

Since there is no data anymore in the `NameType` they could be changed to enums.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/FreshName.kt
```diff
@@ -6,89 +6,85 @@
...
+
+sealed interface NumberedName : FreshName {
+    val n: Int
+}
 
-abstract class TypedFreshName(override val mangledType: NameType, val baseName: String) : FreshName {
-    context(nameResolver: NameResolver)
-    override val mangledBaseName: String
-        get() = baseName
+sealed interface NameTypeVariable : FreshName {
```
jesyspa: I'm not a fan of this name, because it's unclear whether it's (NameType)Variable or Name(TypeVariable) (or something else).

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
```diff
@@ -18,10 +22,156 @@ import org.jetbrains.kotlin.formver.viper.SymbolicName
...
+            is SymbolicName -> resolveSymbolicName(name, skipType, skipScope)
+            is NameType -> if (skipType) emptyList() else resolveNameType(name)
+            else -> throw SnaktInternalException(null, "Unexpected name type: ${name::class.simpleName}")
+        }
+
+        private fun resolveNameScope(
+            name: NameScope, skipScope: Boolean
+        ): List<String> = when (name) {
+            is BadScope -> listOf("<BAD>")
+            is ClassScope -> (if (skipScope) emptyList() else resolveParts(name.parent)) + resolveParts(name.className)
```
jesyspa: There's some inconsistency in these whens (no parens, vs parens, vs braces)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
```diff
@@ -18,10 +22,156 @@ import org.jetbrains.kotlin.formver.viper.SymbolicName
...
-    ).joinToString(SEPARATOR)
-    override fun register(name: SymbolicName) {}
-}
\ No newline at end of file
+
+    val cache = mutableMapOf<AnyName, String>()
+
+    companion object {
+        private fun resolveParts(
+            name: AnyName, skipType: Boolean = false, skipScope: Boolean = false
```
jesyspa: You pass skipScope around a lot.  Maybe having an object that stores that would make sense, rather than just a companion object?  Or pass an ENUM entry as a context parameter?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
```diff
@@ -18,10 +22,156 @@ import org.jetbrains.kotlin.formver.viper.SymbolicName
...
+                name.nameType, skipType
+            ) + listOf(name.name.asStringStripSpecialMarkers()) + resolveParts(name.type.name)
+        }
+
+        private fun resolveOptionalNameType(name: NameType?, skipType: Boolean): List<String> =
+            if (name == null || skipType) emptyList() else {
+                resolveNameType(name)
+            }
+
+        private fun resolveNameType(name: NameType): List<String> = when (name) {
```
jesyspa: Why a list?

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SymbolicName.kt
```diff
@@ -20,49 +24,34 @@ interface AnyName
...
-        val debugResolver = DebugNameResolver()
-        return debugResolver.resolve(this)
-    }
 
 
 /**
  * Collects all types of names we can have.
  */
-sealed class NameType(val name: String) : AnyName {
+sealed interface NameType : AnyName {
```
jesyspa: I'm not a fan of this being in the viper/ section, since it's really a core thing.  Is it necessary?

jesyspa (changes requested):

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SymbolicName.kt
The-Ray-Man: We need to define ``SymbolicName`` in the viper package. Since the interface has the `nameType` field we must have access to `NameType`. 
The core package is dependent on the viper package. So we cannot move the ``NameType`` to the core package because that would introduce circular dependencies. 

We could just define the interface `NameType` in the viper package and the types in the core package, but then we cannot have a sealed interface anymore. I would really prefer to whenever possible to have a sealed interface.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
The-Ray-Man: Basically just to be consistent with the rest of the resolve functions. But yes, this is kind of unnecessary.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SymbolicName.kt
jesyspa: Perhaps it can be two-level?  That is, SymbolicName has as a field of a non-sealed interface, and NameType is in core/ and is sealed?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
The-Ray-Man: It is like this because I did not want to change the behaviour of the name resolver. Because I only wanted to move the logic out of the name data classes. Should I make it consistent?

jesyspa: I think it would be better, yeah

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
The-Ray-Man: I used an ENUM and context parameter. Both flags are now removed. This resulted in some changes in the golden files: 
- In the function name we do no longer have a `t` in front of the function type
- If the class comes from another package, the full package name is shown

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SymbolicName.kt
The-Ray-Man: the two-level thing is a nice idea. I had to move some stuff for the `Domain` from the viper to the core plugin.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
```diff
@@ -18,10 +18,204 @@ import org.jetbrains.kotlin.formver.viper.SymbolicName
...
+            // parent scope (which is the class, making it non-unique)
+            val parentScopes = if (name !is PublicScope) {
+                resolveOptionalNameScope(name.parent)
+            } else {
+                emptyList()
+            }
+            val scope = when (name) {
+                // BadScope should never actually be resolved. We add "<>" to the name such that the viper consistency
+                // checker fails.
+                is BadScope -> listOf("<BAD>")
```
jesyspa: Maybe worth throwing?

jesyspa (approved):
One idea: you could give Core a `CoreSymbolicName` interface that makes `nameType` have type `NameType`.  You could even parametrize `NameResolver` by the type of name.  I think this would give somewhat more compile-time guarantees, but stuff like this can get messy, so see what you prefer.  Either way, LGTM :)

PR #115, opened 2026-04-13
Short Name Resolver

The-Ray-Man:
# Short Name Resolver

Every name has a ordered list of name candidates. To build this candidates a custom DSL exists. The earlier candidates are usually shorter and the last candidate should be the most specific name.

During the registration the `shortNameResolver` collects all names and builds a dependency graph with it. It also infers if a name could appear in the Viper Program. This graph can be viewed form the debugger by calling the `render()` method. 

Once the `resolve` function is called the resolver will use the graph to find unique and short names for all the names that can appear in viper. 
First it will calculated the represented by relation on the names. This is used for things that have different names but we want to merge them. This can happen for public fields with the same name. In Viper we can allow them to be shared. It can happen that the public fields have structurally a different name, therefore they are not deduplicated when registering the names. This usually happens for getters and setters of public fields with the same name, when they are defined in nested classes.

This works in a iterative progress:
1. Identify the conflicting names and group them
2. Choose one group to resolve
3. Select one name and update it's name to the next candidate. To be more precise: If we want to move the name of `nameA`, we look at the current candidate `canA` of `nameA`. If `canA` consists of other names, we will check if we can move one of those, if there are no "parentNames" that can be moved, or it just consists of strings we will select the next candidate of `nameA`. We do this that way, because many "leaf names" are `ScopedNames` and quite early they have the name of the scope as part of their candidate. Since scopes make the code bloaty very fast, we first try to not use them.
5. Restart at 1

Note: In each iteration, we only move one single name. We do it like this, because the moving of the first name might also solve other conflicts. So by this approach we will not perform unnecessary moves.

Afterwards the resolver generated a mapping from the names to their string representation, which is then used to resolve the names during the translation to silver.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ViperKeyword.kt
```diff
@@ -0,0 +1,52 @@
+package org.jetbrains.kotlin.formver.core.names.shortNameResolver
+
+import org.jetbrains.kotlin.formver.viper.AnyName
+
+
+class ViperKeyword(val keyword: String) : AnyName
+
+object ViperKeywords {
+    val keywords = mutableSetOf(
```
jesyspa: Why mutable?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameCandidate.kt
jesyspa, line 1: Not saying one is right or wrong, but why not make `candidates` a virtual function?  It seems to behave more like one

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * A separator between name parts.
+     */
+    object Separator : NamePart
+
+    /**
+     * Indicates that no separator should be added before this part.
+     */
+    object NoSeparator : NamePart
```
jesyspa: This is a bit strange - what happens if you have neither?  Or if you have multiple?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
```diff
@@ -0,0 +1,153 @@
+package org.jetbrains.kotlin.formver.names
+
+import org.jetbrains.kotlin.formver.common.SnaktInternalException
+import org.jetbrains.kotlin.formver.viper.AnyName
+
+/**
+ * Represents a part of a candidate name.
+ */
+sealed interface NamePart {
```
jesyspa: Should the subtypes be data classes?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,610 @@
...
+                throw SnaktInternalException(null, "Unable to make names unique")
+            }
+            currentCollisions = collisions()
+        }
+
+        // Fix the names
+        mangledNames = viperElements().associateWith { current(it) }
+    }
+
+    private fun chooseSeparator() {
```
jesyspa: While this is neat, I think using a fixed separator makes sense for readability.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,610 @@
...
+    /**
+     * For some names (e.g., public fields with the same name), we want to merge them in the viper.
+     * If there are multiple such names, and we do nothing, then there will be unnecessary collisions.
+     * Hence, this function groups the names into groups that should be merged and choose a representative.
+     */
+    private fun createRepresentatives() {
+        // Field Names which are public
+        val publicFields = elements.filter {
+            it is ScopedName && it.scope is PublicScope
+        }
```
jesyspa: I'm a bit concerned about how much Kotlin-specific logic is in this resolver.  Could this be generalized somehow?

jesyspa (changes requested):
I think this generally looks nice, but I'm concerned about how intertwined the logic is.  Would it be possible to make a sharper split between what the programmer should know about names and what the name resolver knows?  That is, instead of hard-coding the logic, give the programmer tools to express it?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameCandidate.kt
The-Ray-Man: I would argue that the ``candidates`` function is logic specifically to the short name resolver. 
I think the names should not contain any resolver-specific logic. Because of this we also moved the simple name resolver logic out of the names and into the resolver. 
I think here we have a similar situation. The candidates belong more to the short name resolver as they do to the name data classes.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
The-Ray-Man: Yes, this is not that clear. So my idea was the following:
- the user should never use ``Separator``. When they use the `CandidateNameBuilder` then they cannot add the `Separator`
- When the candidate name is built, the necessary separators are inserted. (the created candidate name, should never have a `NoSeparator` in between. )
- Having multiple ``NoSeparator``s between two normal names have no special effect. 

But yes, I see how this system could be misunderstood. I try to come up with a better solution.

The-Ray-Man: I tried to make sure that a future user would always use the builder to create the names. Where the issues you mention can not appear. But not sure, if this just makes it overcomplicated.
changes: https://github.com/JetBrains/SnaKt/pull/115/changes/31517047811078d3bc9961d14d562ef3ceda456f

The-Ray-Man:
> Would it be possible to make a sharper split between what the programmer should know about names and what the name resolver knows? That is, instead of hard-coding the logic, give the programmer tools to express it?

I think there are two things that are rather specific logic: 
1. the represented by system
2. allowed collisions (e.g. between "end-up-in-viper-name" and "not-in-viper-name". 

I think one option to not hard code this logic in the name resolver is to specify some relations that are used to solve this problem. I think this could look like this:

We would have some relations, that specify which collisions could be ignored.
```kotlin
interface IgnoreCollisions {
    context(nameRelation: NameRelations)
    fun ignoreCollision(first: AnyName, second: AnyName): Boolean
}
```

As well as some relations that group names into equivalence classes:

```kotlin
interface RepresentativeClasses<C> {
    /**
     * Returns null, if the name is not in the representative class (or in their singleton class)
     * Otherwise returns a representation of the equivalence class. Every object with the same representation will be _merged_
     */
    context(nameRelation: NameRelations)
    fun equivClass(name: AnyName): C?
}
```

These relations might need some additional information to decide if something should be part of it or not, so we also have this object:

```kotlin
interface NameRelations {
    fun isScoped(name: AnyName): Boolean
}
```
Which gives the relations access to some kind of predicates/lookup functions.

Does this go into the direction you imagined? I am also a bit afraid, that this will make things more complicated then it needs to be.

jesyspa:
There are two parts here, I think:
- Some things don't need to be unique because they are never used in Viper.  That's a boolean flag on the name.
- Some things are used like names, but actually map to the same name in Viper.  Maybe we should be using a wrapper object for them, not a name, then.

I think together this gives a simple model where you don't need to explicitly mark equivalence classes, you just need to resolve all collisions between things that may appear in the Viper

The-Ray-Man:
In #120 we could eliminate the problem with structurally different names that should resolve to the same thing. Therefore we could remove the complete representedBy system.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
jesyspa: Could you elaborate on why we sometimes do vs don't want a separator?  Would always using a separator make things worse?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameCandidate.kt
jesyspa: But is it specific to the short name resolver?  Or is it general information about the name, that the resolver happens to use to pick the shortest one?

One good way to check this is to ask whether you could imagine a different resolver for the same information.  I'm on the fence here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ViperKeyword.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+object ViperKeywords {
+    val keywords = setOf(
+        "import",
+        "define",
+        "field",
+        "method",
+        "function",
+        "function",
```
jesyspa: duplicate

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ViperKeyword.kt
```diff
@@ -0,0 +1,54 @@
+package org.jetbrains.kotlin.formver.core.names.shortNameResolver
+
+import org.jetbrains.kotlin.formver.viper.AnyName
+
+
+class ViperKeyword(val keyword: String) : AnyName {
```
jesyspa: Should probably be a dataclass

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
The-Ray-Man: It is not a big deal, but when I looked at the resulting names, I thought that sometimes the number of separators were a bit much. For example, when we have local variables, they are getting prefixed with their level. For a level one variable we would have the name: `l$1$x`. And I think that `l1$x` just looks better (even though not much).

But I almost never use it. So maybe it is also not really worth keeping it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameCandidateBuilder.kt
```diff
@@ -0,0 +1,153 @@
...
+interface CandidateName {
+    fun moveableParts(): List<NamePart.Dependent> = parts.filterIsInstance<NamePart.Dependent>()
+    val parts: List<NamePart>
+}
+
+
+/**
+ * A builder for a list of [CandidateName]s.
+ */
+class CandidatesBuilder {
```
jesyspa, line 12: Option to consider: by default, if you run out of candidates, add a counter to the name.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidate.kt
```diff
@@ -0,0 +1,293 @@
...
+            +baseName
+            +n.toString()
+        }
+    }
+    is PredicateName -> nameOnlyCandidates(name)
+    DispatchReceiverName -> nameWithPrefixCandidates("this", nameType)
+    is DomainAssociatedFuncName -> nameWithPrefixCandidates(name, nameType)
+    is DomainFuncParameterName -> nameWithPrefixCandidates(name, nameType)
+    ExtensionReceiverName -> nameWithPrefixCandidates("this", nameType)
+    FunctionResultVariableName -> nameWithPrefixCandidates("result", nameType)
```
jesyspa: Actually, I'm not sure this is correct - I think maybe this should resolve to the keyword.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    /**
+     * Returns the name collisions of the current candidates.
+     * 
+     * Which names must be considered?
+     * - We do need to make all the names unique, but only once that actually end up in viper. E.g., a variable and
+     *   a scope can have the same name.
+     * - Sometimes we want to have name collisions, e.g., public fields with the same name can be merged. This behavior
+     *   is captured with the representation system. Hence, names that are represented by someone else must not be considered.
```
jesyspa: I think the representation system is outdated now?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,495 @@
...
+     */
+    override fun resolve() {
+        var currentCollisions = collisions()
+        while (currentCollisions.isNotEmpty()) {
+            val toResolve = currentCollisions.entries.first().value
+
+            val candidateToMove = toResolve.filter { canMove(it) }.ifNotEmpty { maxBy { priorityOrder(it) } }
+
+            if (candidateToMove != null) {
+                move(candidateToMove)
```
jesyspa: Worth asserting?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,495 @@
...
+    /** name SCOPED_BY nameScoped **/
+    SCOPED_BY,
+
+    /** Int TYPE_OF Function**/
+    TYPE_OF,
+
+    /** a IS_PART_OF (a,b) -> c **/
+    IS_PART_OF,
+
+    /** name NAME_TYPE nameType **/
```
jesyspa: Doc is outdated

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
```diff
@@ -0,0 +1,153 @@
+package org.jetbrains.kotlin.formver.names
```
jesyspa: Package doesn't match path

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,495 @@
...
+        addElement(a)
+        addElement(b)
+    }
+
+
+    /**
+     * Returns true iff the ``entity`` is scoped. Meaning that there is a name, which wraps around `entity`
+     */
+    private fun isScoped(entity: AnyName): Boolean =
+        tripleStore.any { (a, rel, _) -> a == entity && rel == Relation.SCOPED_BY }
```
jesyspa: Is this still needed?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/ShortNameResolver.kt
```diff
@@ -0,0 +1,495 @@
...
+            is SymbolicName -> {
+                entity.nameType?.let {
+                    link(it, Relation.KIND_OF, entity)
+                    registerEntity(it)
+                }
+                when (entity) {
+                    is FreshName -> registerFreshName(entity)
+                    is ScopedName -> {
+                        link(entity.name, Relation.SCOPED_BY, entity)
+                        registerEntity(entity.name)
```
jesyspa: Many lines like this.  Maybe link + register should have a helper?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidateBuilder.kt
jesyspa: Alternatively: two types of Seq?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -0,0 +1,495 @@
...
+            else -> -1
+        }
+        else -> -1
+    }
+
+
+    /**
+     * Generates short human-readable names. Must be called before using [lookup].
+     */
+    override fun resolve() {
```
jesyspa, line 144: This recomputes collisions each run, which is probably okay, but please time the before/after time of running all our tests (preferably with goldens deleted in each case, so that doesn't impact).  I think it'll be fine, but don't want an accidental regression.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidate.kt
The-Ray-Man: yes I think you are right. Good catch. But I think the solution should be to remove the FunctionResultVariableName and instead using the Viper Keyword name, right? Because otherwise we have two structurally different names for the same thing

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
The-Ray-Man: | Resolver        	        | Total Time 	| Avg per Test 	|
|-----------------	|------------	|--------------	|
| Simple Resolver 	| 7m03s      	| 2.7s         	|
| Short Resolver  	| 7m39s      	| 2.9s         	|
| Difference      	| 36s          	| 0.2s         	|

To be honest, not amazed by the numbers. But I am not sure if we can get away with not recalculating them every time. Because the move operation of collision A, could influence the collision B. Because we might move a parent name, which has something to do with the collision B.

jesyspa: I think it's okay.  The deduplication of conversion/validation tests will make up for tihs

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/shortNameResolver/NameCandidate.kt
jesyspa: It's possible, I guess. The point is that the name is a semantic thing on our side.  But I'm okay with your proposal.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameCandidateBuilder.kt
The-Ray-Man: I am not sure if we should do this. Generally the idea is that the last candidate Name is the most specific one. We would use the counter when the last candidate of two structurally different names are resolved to the same name. But that would mean, that the last candidate name is not as specific as it could be. With the counter, we would then hide this bug. 

But I think what would be better is to add as last name something like a "structural name", which would basically be a serialization of the data class. So basically that it holds `nameA  == nameB <=> nameA.serialized() == nameB.serialized()`. This would then ensure, that we never run out of names.

jesyspa: That's a fair point, though it is honestly quite painful to create (because of all the characters you can have in a name).

I think this is a case where for the end user, adding indices would be better (since they probably don't care about names anyway), while for us as developers we'd rather be sure stuff doesn't collide.

I'm okay with leaving it as you implemented for now.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameCandidateBuilder.kt
```diff
@@ -0,0 +1,115 @@
...
+     * Adds multiple candidates from a [List] of [CandidateName]s.
+     */
+    operator fun List<CandidateName>.unaryPlus() {
+        candidates.addAll(this)
+    }
+
+    /**
+     * Builds the list of [CandidateName]s.
+     */
+    fun build(): List<CandidateName> = candidates.toList()
```
jesyspa, line 57: Ah, one extra difficulty with the idea of generating more is that it stops being a list...
Not sure what I think is best; on the one hand, lazy building will probably mean far fewer are built, but on the other hand, I doubt that's the expensive part here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -0,0 +1,469 @@
...
+
+        /** Int TYPE_OF Function**/
+        TYPE_OF,
+
+        /** a IS_PART_OF (a,b) -> c **/
+        IS_PART_OF,
+
+        /** name NAME_TYPE nameType **/
+        KIND_OF,
+    }
```
jesyspa: I still find the documentation here pretty unclear.  Are these examples or categories?  Maybe you can elaborate?

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/SymbolicName.kt
```diff
@@ -12,9 +12,50 @@ interface AnyName {
...
+    data class Basic(val name: String) : NamePart
+
+    /**
+     * A name part that depends on another [AnyName].
+     */
+    data class Dependent(val name: AnyName) : NamePart
+
+    /**
+     * A separator between name parts.
+     * Do not use this. Use [CandidateNameBuilder] instead.
```
jesyspa: I think this comment is misplaced: it's not "don't use separator", it's "don't build nameparts directly"

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -0,0 +1,469 @@
...
+        if (currentIndex + 1 < entity.candidates().size) return true
+        return currentCandidate(entity).moveableParts().any {
+            canMove(it.name)
+        }
+    }
+
+    // END MANGLE
+
+
+    // START RENDER
```
jesyspa: Would it make sense to make a `debug` directory here and move this code into it?

jesyspa (approved):
Some comments but nothing blocking, looks good :)

PR #116, opened 2026-04-14
Introduce declaration-aware locality checker

### LINKED ISSUE #113: Implement standalone declaration-aware locality checker
Currently locality is checked alongside uniqueness by the uniqueness checker. While the uniqueness checker does employ locality information to assert the uniqueness of a path after it has been passed to a function, locality checking can still be performed independently as it doesn't require dataflow analysis in order to be performed. 

As such, we should create an additional locality checker that is better integrated within Kotlin's plugin infrastructure and that operates independently from the current uniqueness checker. Additionally, the locality checker should handle nested declarations: if a local variable or parameter is declared as local, it may only be assigned within that declaration.

paganma:
This PR implements a declaration-aware locality checker.

It introduces locality diagnostics for assignments, calls, returns, throws, properties, and default arguments. 

The locality of variables is tracked relative to the declaration that owns it. For example, a `@Borrowed` parameter is local to its enclosing function and can be passed to another borrowed position in that same declaration, or passed to a function as a borrowed parameter. Conversely, it cannot be returned or passed to a shared parameter. Borrowed values from an outer declaration are not treated as local inside a nested lambda or local function, because they belong to a different declaration. For example:

```kotlin
fun f(@Borrowed x: A) {
    borrow(x)   // ok
    share(x)    // locality violation
}

fun f(@Borrowed x: A) {
    { @Borrowed val y: A = x } // locality violation: x belongs to the outer declaration
}
```

In addition to this, this PR adds support for resolving the locality of complex expressions such as those containing conditionals. For example:
```kotlin
fun f(@Borrowed x: A, z: A) {
    val y = if (*) { x } else { z } // locality violation: expression type is local but global is expected.
}
```

This PR maintains the original experimental locality checker in the `uniqueness` subproject. The new locality checker is a standalone component residing in `plugin.compiler.locality`. It also provides separate tests in `testData/locality_checker`.


formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/locality/LocalityFunctionCallChecker.kt
```diff
@@ -0,0 +1,79 @@
...
+    context(context: CheckerContext, reporter: DiagnosticReporter)
+    private fun checkArgument(argument: FirExpression, requiredArgumentLocality: Locality) {
+        val actualArgumentLocality = argument.resolvedLocality
+
+        if (requiredArgumentLocality.accepts(actualArgumentLocality)) return
+
+        reporter.reportOn(
+            argument.source,
+            LOCALITY_VIOLATION,
+            "Argument locality mismatch: expected '${requiredArgumentLocality.render()}', " +
```
ligee: This is an unisual use of a diagnostic. In general it is better to place message in the diagnostic and pass parameters to it. It means it will be better to have a separate locality violation diagnostic for each case, so you don't need to construct the message her.
Also rendering could be moved to the diagnostics infrastructure (but it is minor).
You don't necessarily need to fix this right now, this is not critical enough to block the merge. It would be nice to do at some point.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/locality/LocalityFunctionCallChecker.kt
```diff
@@ -0,0 +1,79 @@
...
+                            "actual '${actualReceiverLocality.render()}'."
+                )
+            }
+        }
+
+        val resolvedArgumentMapping = expression.resolvedArgumentMapping
+        val argumentMappings = if (resolvedArgumentMapping != null) {
+            resolvedArgumentMapping.toList()
+        } else {
+            val argumentSymbols = callableSymbol.valueParameterSymbols
```
ligee: Have you seen the situation then the `resolvedArgumentMapping` is unavailable at this stage? I would be surprising for me, since all FIR should be resolved by the time checkers are called.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/analysis/TailValueExtractor.kt
```diff
@@ -0,0 +1,155 @@
...
+import org.jetbrains.kotlin.fir.expressions.FirThrowExpression
+import org.jetbrains.kotlin.fir.expressions.FirTypeOperatorCall
+import org.jetbrains.kotlin.fir.expressions.FirTryExpression
+import org.jetbrains.kotlin.fir.expressions.FirWhenExpression
+import org.jetbrains.kotlin.fir.expressions.FirWrappedExpression
+import org.jetbrains.kotlin.fir.lastExpression
+import org.jetbrains.kotlin.fir.visitors.FirVisitor
+
+/**
+ * Computes symbolic values for an expression at each of its tail position.
```
ligee: I'm worrying about this approach, because you're defining the semantics of the "tail position" here, and it may not completely agree with the Kotlin one, now or in the future. I would very much prefer if we'd reuse some existing semantic structures or visitors from kotlin for such things.
Maybe it is not a blocker. But it definitely a place to put an explanation/TODO, promting us to reimplement it later.

org/jetbrains/kotlin/fir/analysis/cfa/util/PathAwareControlFlowGraphVisitor.kt
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2010-2022 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.fir.analysis.cfa.util
```
ligee: This file and others around it are probably committed by mistake. They are not part of any source roots, and some of them are autogenerated

ligee (changes requested):

org/jetbrains/kotlin/fir/analysis/cfa/util/PathAwareControlFlowGraphVisitor.kt
paganma: Yup, these files were definitely committed by mistake when I tried to copy the code for DFA from the compiler infrastructure. In the end it won't be necessary as the cfa.util.CfgTraverser provided by the compiler is already enough.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/locality/LocalityFunctionCallChecker.kt
paganma: My understanding was that the argumentMapping would be resolved only if keyword arguments are used. Reading your comment, I suspect that this can only happen if there is an error in the resolution of the arguments (e.g. if there is a wrong keyword argument). If this is the case then we can just return early much like how we do when the function itself can't be resolved.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/analysis/TailValueExtractor.kt
paganma: I would also prefer to use a visitor that integrates the semantics of "tail position" but sadly I could find none. The closest thing I have found  was `FirExpressionEvaluator` which is used for constant folding, but it cannot be extended. 

The problem is that the uniqueness and locality that we want to track don't just depend on the type (or type attribute) of an expression, but rather on the path structure of the expression. 

A tail visitor is useful for extracting tails from expressions like:
```kotlin
if (*) { ... x } else { ... y }
```

Which in turn we can extend to find paths for expressions like:
```kotlin
(if (*) { ... x } else { ... y }).z // -> x.z, y.z
```

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/locality/LocalityFunctionCallChecker.kt
paganma: I looked more into the `DiagnosticReporter` and implemented the formatting as intended by the diagnostics API.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/analysis/TailValueExtractor.kt
ligee: Ok, fair enough. Please put this explanation in the comments to this class

ligee (changes requested):
One minor commenting request left.
Also please squash the commits into a reasonable set.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/analysis/TailValueExtractor.kt
paganma: To make the scope of this visitor clearer I have separated the tail visitor logic (now `ExpressionTailFolder`) and path visitor logic (now `ExpressionPathFolder`). The first one is a visitor targeting expressions containing control-flow structures (if, elvis, try-catch, etc.). The second visitor is an extension of this that unwraps and traverses individual path components. Soon I will add documentation in the code to clarify this.

ligee (approved):
Do not forget to squash the commits before merging.

Issue #102, opened 2026-04-09
Support Kotlin object declarations as Viper Algebraic Data Types


pawinkler:
## Goal

Allow Kotlin `object` declarations annotated with `@ADT` to be encoded as Viper ADTs instead of heap-based objects, enabling them to be treated as pure value types in verification.

## Motivating example

```kotlin
@ADT object Red
@ADT object Green
@ADT object Blue

@AlwaysVerify
fun colorIsNotGreen(c: Red) {
    verify(c != Green)
}
```

The objects `Red`, `Green`, `Blue` are distinct constant values with no heap representation. Verification should be able to reason about their equality and inequality directly.

## What needs to be supported

- `@ADT` annotation on `object` and `data object` declarations
- Encoding as a Viper ADT with a single nullary constructor per object
- Equality/inequality reasoning between values of the same ADT type
- Cross-type inequality: values of different ADT types are always unequal

## Restrictions

Some restrictions must be enforced on objects declared to be ADTs to avoid undefined or unsound behavior:

- Only `object` declarations (not `class`, `open class`, `interface`) are accepted
- ADTs must not have fields
- ADTs must not have member functions
- ADTs must be final and must not extend any class or interface.

Some of these restrictions will be reduced in the future.

## Known challenges

- **Upstream Silicon bug**: `DomainInstances.down1Types` in Viper Silver lacks a case for `AdtType`, causing a `MatchError` whenever a method has an ADT-typed formal argument; a fix and rebuild of the local Silicon jar avoids crashing, but verification still does not succeed; the reason is unclear.

pawinkler:
The upstream bug with silicon could be resolved by upgrading our Viper backend to the newest version.

PR #117, opened 2026-04-14
Added support for ADTs based on Kotlin objects

pawinkler:
## Summary

This PR is based upon the issue #102 and implements the requested features. Please refer to it for a proper summary. As this PR is considerably large and introduces new elements into SnaKt, the rest of this description will introduces ideas behind the changes and give an overview of the implementation.

## Architecture

The implementation of ADTs follows a similar approach to the current handling of classes, spanning four layers:

1. **Conversion**: Validation of ADT annotations and creation of embeddings for the ADT types, their constructors and references to them.
2. **Embeddings**: Holds structural information required for the creation of new types, axioms, and linking of references.
3. **Linearization**: Conversion of references to ADT objects to Viper ADT constructor applications.
4. **Viper Output**: Generation of ADT types, injection of new types into runtime with corresponding axioms.

### General idea

The entire architecture is based upon two embeddings:

1. **`AdtTypeEmbedding`**: Represents one ADT type on the Viper level and will be encoded as a new type in the output program.
2. **`AdtConstructorEmbedding`**: Represents a constructor associated with an `AdtTypeEmbedding`.

Each Kotlin object annotated as an ADT will map to exactly one constructor, but not necessarily to one type. Each `AdtTypeEmbedding` has at least one, but not necessarily only one `AdtConstructorEmbedding`. This will allow us to construct sum types in the future. For now, for each ADT object we create a new `AdtTypeEmbedding` with exactly one constructor.

### 1. Conversion

While traversing the input program, ADTs are detected in two ways:

1. **Type-level declarations** (`ProgramConverter`): A class symbol with an `@ADT` annotation is found. `embedAdtClass()` creates the `AdtTypeEmbedding` and `checkAndRegisterAdt()` validates the restrictions and initializes the constructors.
2. **Expression-level references** (`StmtConversionVisitor`): A resolved qualifier to an `@ADT` annotated object is found (e.g. `val c = Red`). The reference is encoded as an `AdtConstructorLit`, which links the mentioned qualifier with its corresponding constructor.

### 2. Embeddings

The embeddings are mostly described in the general idea section already. In addition, we have the `AdtConstructorLit`, which represents a reference to a specific ADT constructor in the expression tree. The embeddings are designed to be easily extendable for future ADT features.

### 3. Linearization

The `LinearizationVisitor` converts discovered `AdtConstructorLit`s to Viper expressions in two steps: first, it builds an `AdtConstructorApp` from the embedding's constructor, then it wraps the result in `toRefFunc(...)`. The result is always `Ref`-typed — ADT values never appear as raw ADT types in method signatures or local variables, following the same pattern as primitives (`Int`, `Bool`).

The top-level ADT type declarations are handled by the `ProgramConverter`, which takes the `AdtTypeEmbedding`s, converts them to the Viper `AdtType` and adds them to the Viper program.

### 4. Viper Output

The following elements are added to the Viper output for each ADT type:

- An ADT type declaration in the generated program
- Runtime type domain entries: a type function, type injection functions, and axioms
- Within methods, functions, and predicates: references to ADTs are encoded as constructor applications, and operations on ADTs are encoded as corresponding operations on the ADT value


jesyspa (changes requested):
Please resolve merge conflicts

pawinkler:
I've found some additional errors, please do not review yet.

pawinkler:
> I've found some additional errors, please do not review yet.

This has been resolved.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/AlgebraicDataType.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.jetbrains.kotlin.formver.viper.toSilver
+import viper.silver.plugin.standard.adt.Adt
+import viper.silver.plugin.standard.adt.AdtConstructor
+import viper.silver.plugin.standard.adt.AdtType
+
+/**
+ * Represents an ADT constructor on the Viper level.
+ *
+ * It is used during the construction of the ADT type and also called during the construction of a new instance,
+ * linking the type and its usage.
```
jesyspa: Please describe classes in terms of what they represent or what their responsibilities are, not how they are used.  If you want to emphasise that we use the constructor as a whole, rather than just its name, when we call it, please make that explicit.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/AlgebraicDataType.kt
```diff
@@ -0,0 +1,62 @@
...
+            AdtType(adtName.mangled, emptyScalaMap(), emptySeq()),
+            adtName.mangled,
+            trafos.toSilver(),
+        )
+}
+
+/** A Viper ADT declaration with a name, a list of constructors, and optional type parameters. */
+data class AlgebraicDataType(
+    val name: SymbolicName,
+    val constructors: List<org.jetbrains.kotlin.formver.viper.ast.AdtConstructor>,
```
jesyspa: I think it would make more sense to be explicit when qualifying Silver names.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Type.kt
```diff
@@ -104,4 +105,29 @@ sealed interface Type : IntoSilver<viper.silver.ast.Type> {
 
     }
 
+    data class Adt(
+        val adtName: SymbolicName,
+        val typeParams: List<TypeVar> = emptyList(),
+        val typeSubstitutions: kotlin.collections.Map<TypeVar, Type> = emptyMap(),
```
jesyspa, line 111: Do you need type parameters already?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -274,6 +292,46 @@ class ProgramConverter(
...
+        return adts.getOrPut(className) {
+            AdtTypeEmbedding(className).also { embedding ->
+                checkAndRegisterAdt(symbol, embedding, className)
+            }
+        }
+    }
+
+    /** Validate constraints on a class annotated with `@ADT` and, if valid, create its constructor embeddings. */
+    @OptIn(SymbolInternals::class)
+    private fun checkAndRegisterAdt(
```
jesyspa: This should likely be two separate functions, and should signal that conversion failed.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -81,12 +84,29 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
+                    "ADT constructors for ${adtEmbedding.name.debugMangled} are not initialized. " +
+                            "This can happen if the referenced object is not actually a valid ADT."
+                )
+            }
+            val constructorEmbedding = adtEmbedding.constructors.firstOrNull()
+                ?: throw SnaktInternalException(resolvedQualifier.source, "ADT constructors not initialized")
+            return AdtConstructorLit(type, constructorEmbedding, adtEmbedding)
         }
-        return UnitLit
+        throw SnaktInternalException(resolvedQualifier.source, "Unsupported reference used as resolved qualifier.")
```
jesyspa: Please invert the condition here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
```diff
@@ -57,3 +60,17 @@ data object NullLit : LiteralEmbedding {
     override val type = buildType { isNullable = true; nothing() }
     override val debugName = "Null"
 }
+
+/**
+ * Embeds a reference to a Kotlin object which was declared to be an ADT.
+ *
+ * Holds the [AdtConstructorEmbedding] and delegates Viper code generation to the linearization visitor.
```
jesyspa: Please go through your code and use your human judgement to evaluate which comments are worth keeping, and which were added by an LLM because of its idiosyncrasies and do not contribute anything.  Every ExpEmbedding delegates Viper code generation to the linearization visitor.  This is not useful.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.jetbrains.kotlin.formver.core.names.ScopedName
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+import org.jetbrains.kotlin.formver.viper.ast.*
+
+/**
+ * Embedding-level representation of an ADT constructor.
+ *
+ * Each Kotlin class annotated declared as an ADT maps to exactly one constructor.
+ */
+class AdtConstructorEmbedding(
```
jesyspa: I am not very happy with the naming here: an AdtConstructorEmbedding would be an embedding of an ADT constructor, but Kotlin does not have any such thing.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -185,12 +184,32 @@ data class LinearizationVisitor(
...
     // endregion
 
     // region Variables
 
     override fun visitVariableEmbedding(e: VariableEmbedding): Linearizable = object : DirectResultLinearizable(e, this@LinearizationVisitor) {
         override fun toViper(ctx: LinearizationContext): Exp = e.toViperExp(ctx)
+
+        override fun toViperBuiltinType(ctx: LinearizationContext): Exp {
+            val adtPretype = e.type.pretype as? AdtTypeEmbedding ?: return defaultToViperBuiltinType(::toViper, e.type, e.sourceRole, ctx)
+            return adtPretype.fromRefFunc(toViper(ctx), pos = ctx.source.asPosition)
```
jesyspa: This seems like a very odd special case, and I don't think it belongs here.  Could you elaborate on what you're trying to do?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -185,12 +184,32 @@ data class LinearizationVisitor(
...
     }
 
+    override fun visitAdtConstructorLit(e: AdtConstructorLit): Linearizable = object : DirectResultLinearizable(e, this@LinearizationVisitor) {
+        override fun toViperBuiltinType(ctx: LinearizationContext): Exp {
+            val adtName = e.adtTypeEmbedding.adtName
+            val viperConstructor = AdtConstructor(
+                AdtConstructorName(adtName, e.constructorEmbedding.className),
+                adtName,
+                emptyList(),
+            )
```
jesyspa: I think this should be something the embedding gives you.  In particular, constructing it like this looks fragile.

jesyspa (changes requested):
Some preliminary comments

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+@ADT
+data object Color1
+
+@ADT
+data object Color2
+
+@ADT
+data object Color3
+
+// --- ADT parameter: type information is available in Viper ---
```
jesyspa: It can be useful to add comments to tests when you want to clarify what you are testing, though ideally it should be clear from the test itself.  Here, you've just added comments to every test.  This is just noise, please think about which ones are useful and remove the rest.

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>takeAdtParam<!>(c: Color1) {}
+
+// --- Two ADT parameters of the same type ---
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>twoDifferentTypeAdtParams<!>(a: Color1, b: Color2) {}
```
jesyspa: Comment and function name and body are not aligned.

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+
+// --- Mix of ADT and regular parameters ---
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>mixedParams<!>(c: Color1, n: Int) {}
```
jesyspa: Why is this an interesting test case?

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+@AlwaysVerify
+fun <!VIPER_TEXT!>adtCrossTypeInequality<!>(c: Color1) {
+    verify(<!VIPER_VERIFICATION_ERROR!>c != Color2<!>)
+}
+
+// --- ADT same-type equality of parameters (pure context) ---
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>adtSameTypeParamEquality<!>(a: Color1, b: Color1) {
+    verify(<!VIPER_VERIFICATION_ERROR!>a == b<!>)
```
jesyspa: This looks like it should verify.

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+    val b = Color1
+    verify(a == b)
+}
+
+
+// --- ADT cross-type inequality in verify (pure context) ---
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>adtCrossTypeInequality<!>(c: Color1) {
+    verify(<!VIPER_VERIFICATION_ERROR!>c != Color2<!>)
```
jesyspa: This looks like it should verify.

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+
+@ADT
+data object Color2
+
+@ADT
+data object Color3
+
+// --- ADT parameter: type information is available in Viper ---
+
+@AlwaysVerify
```
jesyspa: What do you intend to verify here?  There are no (interesting) postconditions.

formver.compiler-plugin/testData/diagnostics/verification/classes/adt.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.jetbrains.kotlin.formver.plugin.verify
+
+@ADT
+data object Color1
+
+@ADT
+data object Color2
+
+@ADT
+data object Color3
```
jesyspa: Do you intentionally not use Color3?

formver.compiler-plugin/testData/diagnostics/conversion/classes/adt_validation.kt
```diff
@@ -0,0 +1,86 @@
...
+
+// --- Valid: bare @ADT data objects ---
+
+@ADT
+data object ValidAdt1
+
+@ADT
+data object ValidAdt2
+
+// --- Trigger class embedding for invalid ADTs to exercise the checker ---
```
jesyspa: Interesting question of whether triggering them explicitly is the right way to go, or whether it would be better to have a separate checker that flags this directly.

formver.compiler-plugin/testData/diagnostics/conversion/classes/adt_validation.kt
```diff
@@ -0,0 +1,86 @@
...
+
+// --- Functions using valid ADTs: mixed ADT + non-ADT parameter ---
+
+fun <!VIPER_TEXT!>mixedParameters<!>(a: ValidAdt1, n: Int) {}
+
+// --- Functions using valid ADTs: local variable ---
+
+fun <!VIPER_TEXT!>adtLocalVariable<!>() {
+    val x = ValidAdt1
+}
```
jesyspa: This looks like it's duplicating the logic you have in validation.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -405,6 +428,26 @@ class RuntimeTypeDomain(val classes: List<ClassTypeEmbedding>) : BuiltinDomain(D
         allInjections.forEach {
             it.apply { injectionAxioms() }
         }
+        validAdtClasses.forEach { adtEmbedding ->
+            val v = domainVar("v", adtEmbedding.viperType)
+            val typeFunc = adtClassTypes[adtEmbedding]!!
+            val toRef = adtToRefFuncs[adtEmbedding]!!
+            val fromRef = adtFromRefFuncs[adtEmbedding]!!
```
jesyspa: I think the !! here are rather suspicious: first of all, the fact that apparently this information is so spread out you need three of them, and secondly that the error if this fails will be bad.  Can't this be part of the structure you iterate over?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -414,5 +457,8 @@ class RuntimeTypeDomain(val classes: List<ClassTypeEmbedding>) : BuiltinDomain(D
                 }
             }
         }
+        adtClassTypes.forEach { (_, typeFunction) ->
+            axiom { typeFunction() subtype anyType() }
+        }
```
jesyspa: One of your goals is to support ADTs that derive from interfaces.  Is it worth decoupling ADTs from other classes to the degree that code like this has to be duplicated for them?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
```diff
@@ -0,0 +1,42 @@
...
+    val className: SymbolicName,
+)
+
+data class AdtTypeEmbedding(override val name: ScopedName) : PretypeEmbedding {
```
jesyspa: This design looks like it's trying to achieve contradictory goals:
- You only have a single Pretype, meaning that you do not support ADTs of sum types (which involve multiple Kotlin types)
- You have multiple constructors, which is only useful for ADTs of sum types.

I would recommend tackling the immediate problem first, so that you can represent individual objects.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -208,7 +210,10 @@ const val RUNTIME_TYPE_DOMAIN_NAME = "rt"
  * // same for subtraction, multiplication and so on
  * ```
  */
-class RuntimeTypeDomain(val classes: List<ClassTypeEmbedding>) : BuiltinDomain(DomainName(RUNTIME_TYPE_DOMAIN_NAME)) {
+class RuntimeTypeDomain(
+    val classes: List<ClassTypeEmbedding>,
+    val adts: List<AdtTypeEmbedding> = emptyList(),
```
jesyspa: Is this default used anywhere?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -274,6 +292,46 @@ class ProgramConverter(
...
+        symbol: FirRegularClassSymbol,
+        embedding: AdtTypeEmbedding,
```
jesyspa: What is the motivation for pre-creating the embedding here?

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Exp.kt
```diff
@@ -406,6 +406,35 @@ sealed interface Exp : IntoSilver<viper.silver.ast.Exp> {
             )
     }
 
+    /**
+     * Represents the application of an ADT constructor.
+     *
+     * Singleton ADTs do not exist as an independent entity in the Kotlin program. Instead, when we encounter a
+     * reference to an ADT type, it will be replaced with a constructor application. This will be modified for the
+     * introduction of product types.
```
jesyspa: This is not clear:
- What do you mean that singleton ADTs do not exist as an independent entity in the Kotlin program?  Isn't that what `data object`s are?
- Are constructor applications types?  If not, how do you replace a type for a constructor application?
- What will be modified?

jesyspa (changes requested):

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Type.kt
pawinkler: Not yet, but I wanted to lay the foundation for future support. Our wrappers (Adt, AlgebraicDataType, AdtConstructor, which I will rename soon) fully match the structure of Silver, with default values for constructor arguments that are not yet needed. Should I change it?

pawinkler: Now renamed to Adt (the type), AdtDecl, and AdtConstructorDecl.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
pawinkler: I chose this approach, because I considered it to be more intuitive to introduce a completely new type and implement the custom behavior for this type with possible repetitions, instead of adding an ADT flag to the ClassTypeEmbedding. The latter introduced boolean checks and meta information about ADTs at many locations in the code base, which kind of break the existing structure by introducing conditional checks withing a type.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: This was a preparation for sum types. I wanted to mimic the handling of classes as much as possible, as we will need the same lazy initialization and the existing system is well developed.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
pawinkler: This is true. It was a preparation for the sum types, but without the remaining implementation it is not helpful. I've updated it handle only ADTs with a single constructor for now.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
pawinkler: Yes, this was a lack of understanding of how Injections work. I've reused the logic of primitive types injections and this change should be obsolete.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
pawinkler: This was a relict of my attempt to recreate some of the Injection logic. This is now deprecated.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
pawinkler: No, this was a relict from an intermediate implementation. I've removed it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -81,12 +84,26 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
     override fun visitResolvedQualifier(
         resolvedQualifier: FirResolvedQualifier, data: StmtConversionContext
     ): ExpEmbedding {
-        if (!resolvedQualifier.resolvedType.isUnit) {
+        if (resolvedQualifier.resolvedType.isUnit) {
+            return UnitLit
+        }
+        val classSymbol = resolvedQualifier.symbol as? FirClassSymbol<*>
+        if (classSymbol == null || !classSymbol.classKind.isObject)
+            throw SnaktInternalException(resolvedQualifier.source, "Unsupported reference used as resolved qualifier.")
```
jesyspa: This isn't very clear.  Is "reference" your own pick here, or a standard Kotlin term? It would also be good to clarify what makes it unsupported (e.g. what's the type).

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -81,12 +84,26 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
+        val type = data.embedType(resolvedQualifier.resolvedType)
+        val adtEmbedding = type.pretype as? AdtTypeEmbedding
+            ?: throw SnaktInternalException(
+                resolvedQualifier.source,
+                "Objects can only be used as ADT constructors. Declare it as an ADT to use it."
```
jesyspa: This is assuming that the user mistakenly didn't mark it as an ADT.  The real issue is that we do not support a feature the user is legitimately using.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -272,27 +276,32 @@ class RuntimeTypeDomain(val classes: List<ClassTypeEmbedding>) : BuiltinDomain(D
         fun classTypeFunc(name: SymbolicName) = createDomainFunc(name, emptyList(), RuntimeType, true)
 
         // bijections to primitive types
-        val intInjection = Injection("int", Type.Int, intType)
-        val boolInjection = Injection("bool", Type.Bool, boolType)
-        val charInjection = Injection("char", Type.Int, charType)
-        val stringInjection = Injection("string", Type.Seq(Type.Int), stringType)
-        val allInjections = listOf(intInjection, boolInjection, charInjection, stringInjection)
+        val intInjection = Injection(UnqualifiedDomainFuncName("intToRef"), UnqualifiedDomainFuncName("intFromRef"), Type.Int, intType)
```
jesyspa: This doesn't look like a very elegant way to do this.  You can make a new name type (or two, for toRef and fromRef) and bundle things that way.

You may want to split it into a separate PR if it touches many tests.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -272,27 +276,32 @@ class RuntimeTypeDomain(val classes: List<ClassTypeEmbedding>) : BuiltinDomain(D
...
 
         // special values
         val nullValue = createDomainFunc(UnqualifiedDomainFuncName("nullValue"), emptyList(), Ref)
         val unitValue = createDomainFunc(UnqualifiedDomainFuncName("unitValue"), emptyList(), Ref)
     }
 
     val classTypes: Map<ClassTypeEmbedding, DomainFunc> = classes.associateWith { it.embedClassTypeFunc() }
+    private val validAdts: List<AdtTypeEmbedding> = adts.filter { it.isInitialized }
+    val adtClassTypes: Map<AdtTypeEmbedding, DomainFunc> = validAdts.associateWith { classTypeFunc(it.name) }
+    private val adtInjections: List<Injection> = validAdts.map { it.injection }
```
jesyspa: This is a lot of duplication: perhaps there is some abstraction for class types that would allow you to avoid writing it twice?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -272,27 +276,32 @@ class RuntimeTypeDomain(val classes: List<ClassTypeEmbedding>) : BuiltinDomain(D
...
     val builtinTypes: List<DomainFunc> =
         listOf(intType, boolType, charType, unitType, nothingType, anyType, functionType, stringType)
     val nonNullableTypes: List<DomainFunc> = buildList {
         addAll(builtinTypes)
         addAll(classTypes.values)
+        addAll(adtClassTypes.values)
     }.distinctBy { it.name }
     override val functions: List<DomainFunc> = nonNullableTypes + listOf(
         nullValue, unitValue, isSubtype, typeOf, nullable
-    ) + allInjections.flatMap { listOf(it.toRef, it.fromRef) }
```
jesyspa, line 294: Why not have `allInjections`?  You need it in multiple places.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
```diff
@@ -57,3 +59,13 @@ data object NullLit : LiteralEmbedding {
     override val type = buildType { isNullable = true; nothing() }
     override val debugName = "Null"
 }
+
+data class AdtLit(
```
jesyspa: It is not clear what this class represents.  Is it an expression that constructs an ADT?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import org.jetbrains.kotlin.formver.core.domains.Injection
+import org.jetbrains.kotlin.formver.core.domains.RuntimeTypeDomain
+import org.jetbrains.kotlin.formver.core.names.AdtConstructorName
+import org.jetbrains.kotlin.formver.core.names.AdtName
+import org.jetbrains.kotlin.formver.core.names.ScopedName
+import org.jetbrains.kotlin.formver.viper.ast.*
+
+data class AdtTypeEmbedding(override val name: ScopedName) : PretypeEmbedding {
+    private var _isInitialized: Boolean = false
```
jesyspa: Please take a look at #122, it is probably the way to go here rather than making this mutable.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/AdtName.kt
```diff
@@ -0,0 +1,28 @@
+package org.jetbrains.kotlin.formver.core.names
+
+import org.jetbrains.kotlin.formver.viper.AnyName
+import org.jetbrains.kotlin.formver.viper.CandidateName
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+
+data class AdtName(val className: SymbolicName, val suffix: String? = null) : SymbolicName {
```
jesyspa: This "suffix" construct looks rather complicated, why do you need it?

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.fir.diag.txt
```diff
@@ -0,0 +1,117 @@
+/singleton.kt:(68,90): error: Invalid ADT annotation: @ADT may only be applied to data object declarations
+
+/singleton.kt:(68,90): error: Invalid ADT annotation: @ADT may only be applied to data object declarations
+
+/singleton.kt:(68,90): error: Invalid ADT annotation: @ADT may only be applied to data object declarations
```
jesyspa: This looks like a bug.

jesyspa (changes requested):
I think the general structure looks reasonable, but there's a lot to polish.  Some suggestion; that said, please also look it over yourself again after making changes.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/AdtName.kt
pawinkler: It's used in the generation of the axioms. For example, when we are generating the {AdtName}toRef and {AdtName}fromRef axioms, the names are generated using the suffix. But I agree, I've simplified it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
pawinkler: Thanks! I've aligned it with this.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
pawinkler: Exactly. I distinguish between declarations and constructions. An `AdtLit` is a literal in the Kotlin source that will be translated to a `AdtConstructorApp` in the linearizer.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
pawinkler: I think I found a decent way on how to implement this, by introducing two convenience constructors in `Injection`. It's a small drawback, but no major changes were needed

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -236,6 +240,41 @@ class ProgramConverter(
         return embedding
     }
 
+    private fun embedAdtClass(symbol: FirRegularClassSymbol): AdtTypeEmbedding {
+        val className = symbol.classId.embedName()
+        typeResolver.lookupAdt(className)?.let { return it }
+        validateAdt(symbol)
```
jesyspa: You are not using the return value of this method.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -236,6 +240,41 @@ class ProgramConverter(
...
+    }
+
+    @OptIn(SymbolInternals::class)
+    private fun validateAdt(symbol: FirRegularClassSymbol): Boolean {
+        if (!symbol.classKind.isObject || !symbol.isData) {
+            errorCollector.addAdtError(symbol.source, "@ADT may only be applied to data object declarations")
+            return false
+        }
+        val fields = mutableListOf<FirPropertySymbol>()
+        symbol.fir.processAllDeclarations(session) { if (it is FirPropertySymbol && it.hasBackingField) fields.add(it) }
```
jesyspa: This looks like an inefficient way to track a boolean - why not just make a `hasFields`?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -236,6 +240,41 @@ class ProgramConverter(
...
+        if (fields.isNotEmpty()) {
+            errorCollector.addAdtError(symbol.source, "An @ADT data object must not have fields")
+            return false
+        }
+        val memberFunctions = mutableListOf<FirNamedFunctionSymbol>()
```
jesyspa: idem

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -236,6 +240,41 @@ class ProgramConverter(
...
+        symbol.fir.processAllDeclarations(session) { if (it is FirNamedFunctionSymbol) memberFunctions.add(it) }
+        if (memberFunctions.isNotEmpty()) {
+            errorCollector.addAdtError(symbol.source, "An @ADT data object must not have member functions")
+            return false
+        }
+        val hasSuperTypes = symbol.resolvedSuperTypes.any { !it.isAny }
```
jesyspa: idem

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -81,12 +83,15 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
     ): ExpEmbedding {
-        if (!resolvedQualifier.resolvedType.isUnit) {
-            throw SnaktInternalException(
-                resolvedQualifier.source, "Only `Unit` is supported among resolved qualifiers currently."
-            )
-        }
-        return UnitLit
+        if (resolvedQualifier.resolvedType.isUnit) return UnitLit
+        val classSymbol = resolvedQualifier.symbol as? FirClassSymbol<*>
+            ?: throw SnaktInternalException(resolvedQualifier.source, "Unsupported symbol given, expected ClassSymbol.")
```
jesyspa: Please make the error messages clear and actionable.  You're not telling the user what they did wrong, and the user should not be expected to know what a ClassSymbol even is (though exposing some details is often unavoidable).

For example, this could be: `Unsupported qualifier type ${resolvedQualifier.symbol.javaClass.simpleName}`

Similar for the errors below.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/Injection.kt
```diff
@@ -40,22 +41,33 @@ import org.jetbrains.kotlin.formver.viper.ast.Function
...
-        listOf(v.decl()),
-        Type.Ref
+    val toRef: DomainFunc = RuntimeTypeDomain.createDomainFunc(toRefName, listOf(v.decl()), Type.Ref)
+    val fromRef: DomainFunc = RuntimeTypeDomain.createDomainFunc(fromRefName, listOf(r.decl()), viperType)
+
+    /**
+     * Convenience constructors that derive `toRefName` and `fromRefName` from a single base.
+     * Two variants exist to support primitives with strings and ADTs with symbolic names.
+     */
+    constructor(prefix: String, viperType: Type, typeFunction: DomainFunc) : this(
```
jesyspa: I don't follow why this needs to be so complicated.  Why not create a `ToRefName` and `FromRefName` that just wrap any name, then have them take a SymbolicName?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -401,7 +405,7 @@ class RuntimeTypeDomain(typeResolver: TypeResolver) : BuiltinDomain(DomainName(R
                 r eq unitValue()
             }
         }
-        allInjections.forEach {
+        (allInjections).forEach {
```
jesyspa: This cange looks spurious.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
jesyspa: Please leave a comment to this effect.

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.kt
```diff
@@ -0,0 +1,40 @@
+// NEVER_VALIDATE
+
+import org.jetbrains.kotlin.formver.plugin.ADT
+
+
+<!ADT_INVALID_TARGET!>@ADT
+class NotAnObject<!>
```
jesyspa, line 6: Why does the error include the `@ADT` annotation here, but not in the next case?

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.kt
```diff
@@ -0,0 +1,40 @@
...
+@ADT
+<!ADT_INVALID_TARGET!>object NotDataObject<!>
+
+@ADT
+data <!ADT_INVALID_TARGET!>object WithField<!> {
+    val x: Int = 42
+}
+
+@ADT
+data <!ADT_INVALID_TARGET, ADT_INVALID_TARGET!>object WithFunction<!> {
```
jesyspa: Why does this one get two errors?

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun <!VIPER_TEXT!>doSomething<!>(): Int = 1
```
jesyspa: Should likely be marked `NeverConvert`

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.fir.diag.txt
```diff
@@ -0,0 +1,117 @@
...
+method doSomething(this: Ref) returns (ret: Ref)
+  requires isSubtype(typeOf(this), rt_c_pkg__c_WithFunction())
+  ensures isSubtype(typeOf(ret), intType())
+{
+  ret := intToRef(1)
+  goto ret_0
+  label ret_0
+}
+
+adt adt_c_pkg__c_WithFunction {
```
jesyspa: This is a fairly nasty name - why do we not generate a nicer one?

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.kt
```diff
@@ -0,0 +1,40 @@
...
+
+interface SomeInterface
+
+@ADT
+data <!ADT_INVALID_TARGET!>object WithSupertype<!> : SomeInterface
+
+@ADT
+data object Valid
+
+fun <!VIPER_TEXT!>triggerNotAnObject<!>(x: NotAnObject) {}
```
jesyspa: It looks like we are still emitting VIPER_TEXT even if the conversion failed, which would be a bug.

formver.compiler-plugin/testData/diagnostics/verification/adts/singleton.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.jetbrains.kotlin.formver.plugin.postconditions
+import org.jetbrains.kotlin.formver.plugin.preconditions
+import org.jetbrains.kotlin.formver.plugin.verify
+
+@ADT
+data object Red
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>adtIsAny<!>(r: Red): Red {
+    postconditions<Red> { res -> <!USELESS_IS_CHECK!>res is Red<!> }
```
jesyspa: Please suppress the warning when it isn't the one we're testing for.

formver.compiler-plugin/testData/diagnostics/verification/adts/singleton.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.jetbrains.kotlin.formver.plugin.AlwaysVerify
+import org.jetbrains.kotlin.formver.plugin.postconditions
+import org.jetbrains.kotlin.formver.plugin.preconditions
+import org.jetbrains.kotlin.formver.plugin.verify
+
+@ADT
+data object Red
+
+@AlwaysVerify
+fun <!VIPER_TEXT!>adtIsAny<!>(r: Red): Red {
```
jesyspa: The name doesn't look like it aligns with the actual implementation.

I am also not sure what this is supposed to test.  The postcondition you add is already present from the return type.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Type.kt
jesyspa: I think it's fine.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
jesyspa: Additionally: why call `processAllDeclarations` three times when you just want to check three properties of each declaration?

jesyspa: Additionally additionally: is there no helper to avoid going into `SymbolInternals` here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
jesyspa: I think generating the common axioms uniformly would be worth it; for example, you forgot to add the axioms for nullability.  (Please also add tests for that, by the way.)

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/AdtDecl.kt
```diff
@@ -0,0 +1,60 @@
+package org.jetbrains.kotlin.formver.viper.ast
```
jesyspa, line 6: Add license header (same as on all other files) please.  IntelliJ should do this for you on file creation.

formver.compiler-plugin/testData/diagnostics/verification/adts/singleton.kt
```diff
@@ -0,0 +1,14 @@
+import org.jetbrains.kotlin.formver.plugin.ADT
+import org.jetbrains.kotlin.formver.plugin.AlwaysVerify
+import org.jetbrains.kotlin.formver.plugin.postconditions
+import org.jetbrains.kotlin.formver.plugin.preconditions
+import org.jetbrains.kotlin.formver.plugin.verify
+
+@ADT
+data object Red
```
jesyspa, line 10: A test that exercises the construction and usage of an ADT data object would be useful.

jesyspa (changes requested):
I think the general structure is okay, but the code still has a lot of minor issues and things that look like oversights.  After fixing, it would be good if you could go through another time before re-requesting review.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/Injection.kt
pawinkler: They can't be handled by a direct wrapper right now, because they produce structurally different names. ADTs construct a candidate according to their definition (e.g. `Green_ToRef`) and the current construction of primitive types simply combines both (e.g. `IntToRef`). 
I think it makes sense to implement the wrapper and use `SymbolicName` in general, but this will effect most test cases. Should do this in a separate PR?

jesyspa: Splitting things up would be good, yes.  Though here it looks to me like you could use the no-separator option to avoid needing to.

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.kt
pawinkler: This is a result from the modular verification. The ADT is referenced by two independent functions and in both cases the checker complains. 
I've removed one of them here, as it doesn't contribute to the test case, but this would still happen.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/Injection.kt
pawinkler: Thanks! This worked.

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.kt
pawinkler: I don't really understand what's happening here. I'm always using the `symbol.source` for the position and, as far as I can tell, FIR source has different text ranges for objects and classes.

formver.compiler-plugin/testData/diagnostics/conversion/adts/singleton.fir.diag.txt
pawinkler: I don't fully understand why we always end up with such a complex construction. Easier candidates should exist, but aren't used. I will look into it more soon and otherwise ask @The-Ray-Man on Monday.

formver.compiler-plugin/testData/diagnostics/verification/adts/singleton.kt
pawinkler: I have two smaller commits ready which contain examples with equality and type checks. However, they also modify the code a bit more. I didn't want to inflate this PR even more. Would it be acceptable if I make two PRs as soon as this one is merged?

pawinkler:
I've tried to support the review process a bit by splitting all the changes into smaller, self-contained commits. All the commits should address one category of feedback and are hopefully independently reviewable.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -236,6 +252,35 @@ class ProgramConverter(
         return embedding
     }
 
+    private fun embedAdtClass(symbol: FirRegularClassSymbol): AdtTypeEmbedding {
+        val className = symbol.classId.embedName()
+        typeResolver.lookupAdt(className)?.let { return it }
+        val embedding = AdtTypeEmbedding(className, isValid = validateAdt(symbol))
+        typeResolver.registerAdt(embedding)
```
jesyspa: You do not use the return value of this function, and as far as I can see it is not used anywhere else, either.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -236,6 +252,35 @@ class ProgramConverter(
...
+        return embedding
+    }
+
+    private fun validateAdt(symbol: FirRegularClassSymbol): Boolean {
+        if (!symbol.classKind.isObject || !symbol.isData) {
+            errorCollector.addAdtError(AdtErrorKind.INVALID_TARGET, symbol.source, "@ADT may only be applied to data object declarations")
+            return false
+        }
+        val declarations = symbol.declarations
+        if (declarations.filterIsInstance<FirPropertySymbol>().isNotEmpty()) {
```
jesyspa: Again: you are creating several lists here, just for the sake of checking whether any of them is non-empty.  Are you sure this is the best way to do this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -108,6 +114,16 @@ class ProgramConverter(
                 body = stmtCtx.convertMethodWithBody(declaration, signature, returnTarget)
             }
         }
+        // Any invalid-ADT touch (signature, body, transitive callee, transitive field) registers an entry in
+        // `errorCollector.adtErrors` via `validateAdt`
```
jesyspa: I don't follow what the intent of this code is.  Instead of saying what other code is supposed to do, please explain what this code ensures.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -108,6 +114,16 @@ class ProgramConverter(
...
+        // TODO: Optimize this, e.g. with a body traversal or a check at expression level
```
jesyspa: Why would a body traversal be more efficient?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
```diff
@@ -57,3 +59,23 @@ data object NullLit : LiteralEmbedding {
     override val type = buildType { isNullable = true; nothing() }
     override val debugName = "Null"
 }
+
+/**
+ * Reference to a literal/constructor of an object/class which was declared to be an ADT in the Kotlin source.
+ * As ADTs are value-based, this is treated as an ADT instantiation and triggers an ADT constructor associated
```
jesyspa, line 65: I'm a bit confused, what does them being value-based have to do with it?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/DomainName.kt
```diff
@@ -48,6 +53,14 @@ data class QualifiedDomainFuncName(val domainName: SymbolicName, val funcName: S
     override val children: List<AnyName> = listOfNotNull(domainName, funcName, nameType)
 }
 
+data class InjectionFuncName(val baseName: SymbolicName, val suffix: String) : SymbolicName {
```
jesyspa: I think it would be better to use explicit toRef and fromRef classes rather than just having a string here; strings tend to be quite brittle in ways that types are not.

jesyspa (approved):
I left some comments, please resolve them and then feel free to merge.  Please make sure to squash-merge, as opposed to rebase.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: Sorry, efficient is misleading. The current location is out-of-place, so I wanted to move it, and we are only reporting on the declaration, not the actual interaction with the ADT. This is what should be updated.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: I've discovered a new property of `FirRegularClassSymbol`. Now I am iterating only once over the list, which should be more efficient

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
pawinkler: I wanted to say that we are not referencing an object here, but instead creating a new ADT value, which differs from the default behavior because ADTs are value-based.

PR #120, opened 2026-04-23
Property Name Update

The-Ray-Man:
This PR contains multiple changes on different things, but they are all related.

## Unique Program Parts
Before creating the viper programs, some structure of the "ExpEmbedding-Program" were made unique. For example this was done to only have one public field per name. This was performed in the following way:

```kotlin
fields.distinctBy { it.name.debugMangled }
```
The name of the fields get transformed to a string using the `DebugNameResolver`. But since our names are structurally unique, we should also use structural comparison instead of reliying on the NameResolver to translate them to strings.

Therefore we only use `it.name` to make the fields unique.

## Structurally Wrong Names
This change uncovered the problem on how we create the names for public fields. Since we want to merge all public fields  with the same name, the public name should have the form: `ScopedName(public, fieldName)`. But actually, the field name contained the name of the corresponding class. The class Name was included as a parent scope of the public scope. 
Before this was not an issue, because the `DebugNameResolver` ignored the parents scopes of the public scope. (So does the simple name resolver).
Therefore we changed the public scope to no longer hold a parent scope. (private field names, still have their corresponding class in their name)

## Embedding Problem
The previous change resulted in a follow up problem. In the situation where there are two public fields with the same name, but with different getters, we need to be able to decide which getter to use. The `ProgramConverter` used to look this up using the property name. This used because the property name contained the corresponding class. But now this is no longer possible.
We came up with the following solution: When embedding the name of a property we return a pair: The name of the corresponding class as well as the name of the property. This pair is then used to lookup the property.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
@@ -89,12 +89,12 @@ class ProgramConverter(
             // We need to deduplicate fields since public fields with the same name are represented differently
             // at `FieldEmbedding` level but map to the same Viper.
             fields = SpecialFields.all.map { it.toViper() } +
-                    fields.distinctBy { it.name.debugMangled }.map { it.toViper() },
+                    fields.distinctBy { it.name }.map { it.toViper() },
             functions = SpecialFunctions.all +
-                    functions.values.mapNotNull { it.viperFunction }.distinctBy { it.name.debugMangled },
+                    functions.values.mapNotNull { it.viperFunction }.distinctBy { it.name },
             methods = SpecialMethods.all +
                     methods.values.mapNotNull { it.viperMethod }
-                        .distinctBy { it.name.debugMangled } + havocMethods.values,
+                        .distinctBy { it.name } + havocMethods.values,
             predicates = classes.values.flatMap {
                 listOf(
                     it.details.sharedPredicate,
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
@@ -40,7 +40,7 @@ import org.jetbrains.kotlin.utils.addToStdlib.ifFalse
 import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 
 
-typealias PropertyName = Pair<ScopedName, ScopedName>
+typealias ClassPropertyPair = Pair<ScopedName, ScopedName>
 
 /**
  * Tracks the top-level information about the program.
@@ -66,7 +66,7 @@ class ProgramConverter(
      * [properties] contains all the properties. This includes properties defined on interfaces etc.
      * They are stored in a map. The key is the pair of the name of the class and the field.
      **/
-    private val properties: MutableMap<PropertyName, PropertyEmbedding> = mutableMapOf()
+    private val properties: MutableMap<ClassPropertyPair, PropertyEmbedding> = mutableMapOf()
 
     /** [fields] contain the properties that actually have backing fields. **/
     private val fields: MutableSet<FieldEmbedding> = mutableSetOf()
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
@@ -39,6 +39,9 @@ import org.jetbrains.kotlin.utils.addIfNotNull
 import org.jetbrains.kotlin.utils.addToStdlib.ifFalse
 import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 
+
+typealias PropertyName = Pair<ScopedName, ScopedName>
+
 /**
  * Tracks the top-level information about the program.
  * Conversions for global entities like types should generally be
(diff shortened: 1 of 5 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -59,7 +58,14 @@ class ProgramConverter(
         }.toMutableMap()
     private val functions: MutableMap<SymbolicName, PureFunctionEmbedding> = mutableMapOf()
     private val classes: MutableMap<SymbolicName, ClassTypeEmbedding> = mutableMapOf()
-    private val properties: MutableMap<SymbolicName, PropertyEmbedding> = mutableMapOf()
+
+    /**
+     * [properties] contains all the properties. This includes properties defined on interfaces etc.
+     * They are stored in a map. The key is the pair of the name of the class and the field.
+     **/
+    private val properties: MutableMap<Pair<SymbolicName, SymbolicName>, PropertyEmbedding> = mutableMapOf()
```
jesyspa: Maybe worth making a type for this and making the types more specific?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/MemberEmbeddingPolicy.kt
```diff
@@ -18,4 +18,4 @@ fun onlyPrivateScopedPolicy(isPrivate: Boolean): MemberEmbeddingPolicy =
     if (isPrivate) MemberEmbeddingPolicy.PRIVATE else MemberEmbeddingPolicy.UNSCOPED
 
 val MemberEmbeddingPolicy.isScoped: Boolean
```
jesyspa: Is this still needed?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameEmbeddings.kt
```diff
@@ -122,9 +124,17 @@ fun FirPropertySymbol.embedSetterName(ctx: ProgramConversionContext): ScopedName
...
-fun FirPropertySymbol.embedMemberPropertyName() = callableId!!.embedMemberPropertyName(
-    Visibilities.isPrivate(this.visibility)
-)
+/**
+ * Returns a pair that uniquely identifies the property.
+ * The first element is the name of the class that contains the property, and the second is the name of the property itself.
+ */
+fun FirPropertySymbol.embedMemberPropertyName(): Pair<SymbolicName, SymbolicName> {
+    val className =
+        callableId!!.classId?.embedName() ?: throw SnaktInternalException(null, "Property is not part of a class")
```
jesyspa: Why not use the symbol source for the position of this?
And why not use ?. for the first access, too?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameScope.kt
```diff
@@ -50,7 +48,8 @@ data class ClassScope(override val parent: NameScope, val className: ClassKotlin
  * We do not want to mangle field names with class and package, hence introducing
  * this special `NameScope`. Note that it still needs package and class for other purposes.
  */
-data class PublicScope(override val parent: NameScope) : NameScope {
+data object PublicScope : NameScope {
+    override val parent: NameScope? = null
     override val parentAccessible: Boolean
```
jesyspa: Is stuff like this still relevant?  IIRC it was added to handle the special public treatment

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameScope.kt
The-Ray-Man: The parent? I think so. Because the private fields are storing the class scope as a parent.

jesyspa: I mean, parentAccessible

The-Ray-Man: Ah sorry. No we do not need it anymore. I removed it

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -32,14 +32,16 @@ import org.jetbrains.kotlin.formver.core.embeddings.properties.*
...
-import org.jetbrains.kotlin.formver.names.debugMangled
 import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Method
 import org.jetbrains.kotlin.formver.viper.ast.Program
 import org.jetbrains.kotlin.utils.addIfNotNull
 import org.jetbrains.kotlin.utils.addToStdlib.ifFalse
 import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 
+
+typealias PropertyName = Pair<ScopedName, ScopedName>
```
jesyspa: nit: this isn't a name.  ClassPropertyPair maybe?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameEmbeddings.kt
```diff
@@ -122,9 +125,17 @@ fun FirPropertySymbol.embedSetterName(ctx: ProgramConversionContext): ScopedName
...
-    Visibilities.isPrivate(this.visibility)
-)
+/**
+ * Returns a pair that uniquely identifies the property.
+ * The first element is the name of the class that contains the property, and the second is the name of the property itself.
+ */
+fun FirPropertySymbol.embedMemberPropertyName(): PropertyName {
+    val className =
+        callableId?.classId?.embedName() ?: throw SnaktInternalException(source, "Property is not part of a class")
+    val propertyName = callableId!!.embedMemberPropertyName(Visibilities.isPrivate(this.visibility))
```
jesyspa: Is the !! needed here?  Since you access callableId.  You might need to put it in a local.

jesyspa (approved):

PR #122, opened 2026-04-24
Immutable class type embedding

The-Ray-Man:
## Changes

- Created `TypeResolver`, which holds information about subtype relations as well as fields. Basically it contains most of the information that was previously stored in `ClassEmbeddingDetails`.
- Since field and subtype information are needed in many parts of the conversion, we need to pass the `TypeResolver` around a lot. The general spirit is, that the `TypeResolver` should be a "utility object". It should just give information about the type hierarchy and allow for basic operations on the type hierarchy. But it should not implement "viper specific" logic, like predicate building etc.
- Completely removed `ClassEmbeddingDetails`. The functionality was moved to `ClassTypeEmbedding` or `TypeResolver`.
- `PropertyKotlinName` was moved out of our `SymbolicName` hierarchy. Because this is not a name that should be used in the end, but just the name used to lookup the properties (which can be represented as a backing field or method call and soon function call as well)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -223,31 +216,23 @@ class ProgramConverter(
...
     /**
      * Returns an embedding of the class type, with details set.
      */
     private fun embedClass(symbol: FirRegularClassSymbol): ClassTypeEmbedding {
         val className = symbol.classId.embedName()
-        val embedding = classes.getOrPut(className) {
-            buildClassPretype {
-                withName(className)
-            }
+        if (typeResolver.isRegistered(className)) return typeResolver.lookupEmbedding(className)!!
```
jesyspa: Would a lookup + conditional return work here?  It seems a bit strange to check just to look it up and then unsafe dereference

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -92,22 +90,21 @@ class ProgramConverter(
...
+                .map { it.toViper() },
+            functions = SpecialFunctions.all + functions.values.mapNotNull { it.viperFunction(typeResolver) }
+                .distinctBy { it.name },
+            methods = SpecialMethods.all + methods.values.mapNotNull { it.viperMethod(typeResolver) }
+                .distinctBy { it.name } + havocMethods.values,
+            predicates = typeResolver.embeddings().flatMap {
                 listOf(
-                    it.details.sharedPredicate,
-                    it.details.uniquePredicate
+                    it.sharedPredicate(typeResolver), it.uniquePredicate(typeResolver)
```
jesyspa: One thing to consider: would these be clearer if `typeResolver` was a context parameter?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -0,0 +1,143 @@
...
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+
+class TypeResolver {
+    /**
+     * Collection of all ClassTypeEmbeddings and InterfaceTypeEmbeddings
+     */
+    private val classEmbedding = mutableMapOf<SymbolicName, ClassTypeEmbedding>()
+    private val interfaceEmbedding = mutableMapOf<SymbolicName, ClassTypeEmbedding>()
+    private val embedding
+        get() = classEmbedding + interfaceEmbedding
```
jesyspa: This performs a pretty costly map concatenation on every access.  Is it needed?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -223,31 +216,23 @@ class ProgramConverter(
...
 
-        val newDetails =
-            ClassEmbeddingDetails(
-                embedding,
-                symbol.classKind.isInterface,
-            )
-        embedding.initDetails(newDetails)
+        typeResolver.register(embedding, symbol.classKind.isInterface)
 
         // The full class embedding is necessary to process the signatures of the properties of the class, since
```
jesyspa: I feel like with this change, the comment (and initialization order) can be relaxed?  In particular, I would expect it to be fine to create embeddings lazily, when they're first requested.

jesyspa (changes requested):
Some thoughts.  In general, this feels like a PR that touches a lot of things, with most being incidental.  Maybe there's a smaller PR looking to get out?  But in principle I'm not completely against merging it roughly as is.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -33,14 +33,16 @@ import org.jetbrains.kotlin.formver.core.embeddings.types.*
...
-import org.jetbrains.kotlin.utils.addToStdlib.ifFalse
 import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 
-
-typealias ClassPropertyPair = Pair<ScopedName, ScopedName>
+/**
+ * Name used to look up properties.
+ */
+data class PropertyKotlinName(val scope: NameScope, val name: Name)
+typealias ClassPropertyPair = Pair<ScopedName, PropertyKotlinName>
```
jesyspa: Is this only used internally?  It feels a bit random to have this here - perhaps it can be a private typealias (if such a thing exists) or a private dataclass?  (To be honest, I would generally use dataclasses over pairs)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -128,18 +118,15 @@ class ProgramConverter(
 
     @OptIn(SymbolInternals::class)
     private fun embedPureUserFunction(
-        symbol: FirFunctionSymbol<*>,
-        signature: FullNamedFunctionSignature
+        symbol: FirFunctionSymbol<*>, signature: FullNamedFunctionSignature
     ): PureUserFunctionEmbedding {
         (functions[signature.name] as? PureUserFunctionEmbedding)?.also { return it }
-        val new = PureUserFunctionEmbedding(processCallable(symbol, signature))
+        val new = PureUserFunctionEmbedding(embedCallable(symbol, signature))
```
jesyspa, line 118: Just for my understanding: why is this part of the change?  Was it a bug previously?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -223,63 +209,36 @@ class ProgramConverter(
...
      * Returns an embedding of the class type, with details set.
      */
     private fun embedClass(symbol: FirRegularClassSymbol): ClassTypeEmbedding {
         val className = symbol.classId.embedName()
-        val embedding = classes.getOrPut(className) {
-            buildClassPretype {
+        typeResolver.lookupEmbedding(className)?.let { return it }
+
+        val embedding = typeResolver.getOrPutEmbedding(className) {
+            val embedding = buildClassPretype {
```
jesyspa: The duplicate `val embedding` looks strange, perhaps there's a clearer way to express this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
jesyspa: Looking at TypeResolver: I think a dataclass in that file would make a lot of sense

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -0,0 +1,143 @@
...
+    fun register(typeEmbedding: ClassTypeEmbedding, isInterface: Boolean) = when (isInterface) {
+        true -> interfaceEmbedding.putIfAbsent(typeEmbedding.name, typeEmbedding)
+        false -> classEmbedding.putIfAbsent(typeEmbedding.name, typeEmbedding)
+    }
+
+    private fun toBackingField(property: PropertyEmbedding): FieldEmbedding? =
+        (property.getter as? BackingFieldGetter)?.field
+
+
+    fun embeddings() = (classEmbedding.values.toList() + interfaceEmbedding.values.toList()).distinctBy { it.name }
```
jesyspa: We call a lot of things embeddings, so this isn't very clear.

jesyspa (approved):
Generally this looks good, a few comments but I think nothing blocking.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
The-Ray-Man: I just aligned the function names. Every other action was called ``embed`` while this one was called `process`

jesyspa: Ahh, makes sense.  I think there was a reason, but I can imagine it may not be relevant anymore.

PR #123, opened 2026-04-24
Add dump-test-diff.sh for viewing golden file assertion diffs

komiputer:
## Summary
- Adds `scripts/dump-test-diff.sh` — a helper that runs a single test and dumps the expected vs actual content when a golden file assertion fails.
- Gradle's cross-JVM serialization strips `AssertionFailedError` expected/actual values, so you never see the diff in normal test output. This script works around that by registering a JUnit 5 `TestWatcher` extension that catches failures inside the test JVM and writes both sides to `$SNAKT_TEST_DUMP_DIR/test-assertion-dump-*.txt` (default `/tmp`; override the env var to redirect).
- Handles the Kotlin test framework's `FileInfo` wrapper by extracting the file path and reading the golden file content directly.

The extension itself (`DumpAssertionDiffExtension`) lives in `test-fixtures` as a normal `.kt` file. It is inert unless the `META-INF/services` registration and `junit-platform.properties` autodetection flag are present, so leaving it on disk has no effect on normal test runs — the script only creates and cleans up those two registration files.

This supersedes #110, which emitted the extension source via a heredoc in the shell script; pulling it into a real file makes it reviewable/editable with IDE support and shrinks the script considerably.

Usage:
```
./scripts/dump-test-diff.sh "testArithmetic"
SNAKT_TEST_DUMP_DIR=/var/tmp/snakt ./scripts/dump-test-diff.sh "testArithmetic"
```

## Test plan
- [x] Intentionally broke a golden file, ran the script, confirmed both expected and actual content appear in the dump with the correct diff
- [x] Verified `./gradlew :formver.compiler-plugin:compileTestFixturesKotlin` succeeds with the new file
- [x] Verified `SNAKT_TEST_DUMP_DIR` override lands the files in the requested directory

### DIFF
--- a/formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/DumpAssertionDiffExtension.kt
+++ b/formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/DumpAssertionDiffExtension.kt
@@ -1,0 +1,74 @@
+package org.jetbrains.kotlin.formver.plugin
+
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.junit.jupiter.api.extension.TestWatcher
+import org.opentest4j.AssertionFailedError
+import java.io.File
+
+// Dumps expected vs actual content from golden-file assertion failures to
+// /tmp/test-assertion-dump-*.txt. Inert unless registered via
+// META-INF/services and enabled via junit-platform.properties autodetection;
+// scripts/dump-test-diff.sh manages that registration transiently.
+class DumpAssertionDiffExtension : TestWatcher {
+    override fun testFailed(context: ExtensionContext, cause: Throwable) {
+        val baseName = context.displayName.replace(Regex("[^a-zA-Z0-9_]"), "_")
+        val assertions = collectAssertionErrors(cause)
+        for ((index, error) in assertions.withIndex()) {
+            val suffix = if (assertions.size > 1) "_$index" else ""
+            val name = "$baseName$suffix"
+            val expected = resolveValue(error.expected)
+            val actual = resolveValue(error.actual)
+            File("/tmp/test-assertion-dump-$name.txt").writeText(buildString {
+                appendLine("=== EXPECTED ===")
+                appendLine(expected)
+                appendLine()
+                appendLine("=== ACTUAL ===")
+                appendLine(actual)
+            })
+            System.err.println(">>> Assertion diff dumped to /tmp/test-assertion-dump-$name.txt")
+        }
+    }
+
+    private fun collectAssertionErrors(throwable: Throwable): List<AssertionFailedError> {
+        if (throwable is AssertionFailedError && throwable.isExpectedDefined && throwable.isActualDefined) {
+            return listOf(throwable)
+        }
+        // MultipleFailuresError (opentest4j, Assertions.assertAll) exposes getFailures();
+        // Gradle's DefaultMultiCauseException exposes getCauses(). Both give a List<Throwable>.
+        for (methodName in listOf("getFailures", "getCauses")) {
+            try {
+                val method = throwable.javaClass.getMethod(methodName)
+                @Suppress("UNCHECKED_CAST")
+                val children = method.invoke(throwable) as? List<Throwable>
+                if (children != null && children.isNotEmpty()) {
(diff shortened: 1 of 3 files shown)


formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/DumpAssertionDiffExtension.kt
```diff
@@ -0,0 +1,74 @@
...
+class DumpAssertionDiffExtension : TestWatcher {
+    override fun testFailed(context: ExtensionContext, cause: Throwable) {
+        val baseName = context.displayName.replace(Regex("[^a-zA-Z0-9_]"), "_")
+        val assertions = collectAssertionErrors(cause)
+        for ((index, error) in assertions.withIndex()) {
+            val suffix = if (assertions.size > 1) "_$index" else ""
+            val name = "$baseName$suffix"
+            val expected = resolveValue(error.expected)
+            val actual = resolveValue(error.actual)
+            File("/tmp/test-assertion-dump-$name.txt").writeText(buildString {
```
The-Ray-Man: We might want to supply the used directory over an environment variable.

The-Ray-Man (changes requested):
I think it could be optimized to some extend. In our golden files we do have the line numbers written. 
So if there are multiple methods, and the first methods changed, it has an output on the rest of the methods. Meaning that with this we have unnecessary output. I tried it by changing the first method of the arithmetic test and got this as output:

```
=== EXPECTED ===
/arithmetic.kt:(23,31): info: Generated Viper text for addition:
method f$addition$TF$T$Int$T$Unit(p$x: Ref) returns (ret$0: Ref)
  requires df$rt$isSubtype(df$rt$typeOf(p$x), df$rt$intType())
  ensures df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
{
  var l0$y: Ref
  l0$y := sp$plusInts(p$x, p$x)
  label lbl$ret$0
  inhale df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
}

/arithmetic.kt:(66,77): info: Generated Viper text for subtraction:
method f$subtraction$TF$T$Int$T$Unit(p$x: Ref) returns (ret$0: Ref)
  requires df$rt$isSubtype(df$rt$typeOf(p$x), df$rt$intType())
  ensures df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
{
  var l0$y: Ref
  l0$y := sp$minusInts(p$x, p$x)
  label lbl$ret$0
  inhale df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
}

[a lot more]

=== ACTUAL ===
/arithmetic.kt:(23,31): info: Generated Viper text for addition:
method f$addition$TF$T$Int$T$Unit(p$x: Ref) returns (ret$0: Ref)
  requires df$rt$isSubtype(df$rt$typeOf(p$x), df$rt$intType())
  ensures df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
{
  var l0$y: Ref
  l0$y := sp$plusInts(sp$plusInts(p$x, p$x), p$x)
  label lbl$ret$0
  inhale df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
}

/arithmetic.kt:(70,81): info: Generated Viper text for subtraction:
method f$subtraction$TF$T$Int$T$Unit(p$x: Ref) returns (ret$0: Ref)
  requires df$rt$isSubtype(df$rt$typeOf(p$x), df$rt$intType())
  ensures df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
{
  var l0$y: Ref
  l0$y := sp$minusInts(p$x, p$x)
  label lbl$ret$0
  inhale df$rt$isSubtype(df$rt$typeOf(ret$0), df$rt$unitType())
}

[a lot more]
```

The first one is actually different, but in the second one just the number changed.

I think we could do some regex magic or just remove the numbers completely. Removing them completely is not that simple because we are using the test infrastructure. However, with the new test driver #96 I needed to implement some part myself and there it would be easy to remove this numbers.

The-Ray-Man (approved):
Looks good.

PR #124, opened 2026-04-28
Use ShortNameResolver

The-Ray-Man:


The-Ray-Man:
@copilot
This PR changes the way we resolve names. Therefore the resulting programs are very different now. However only the syntax should be different. The resulting programs should semantically be equivalent.
Look at the changes in the files ending with `diag.txt`. Are there any changes that result in a semantically different program?

jesyspa (approved):
Awesome!  Congrats on landing this :D

One thought: maybe _ would be a better default choice for separator, it tends to be easier to read.

The-Ray-Man:
>One thought: maybe _ would be a better default choice for separator, it tends to be easier to read.

I agree that `_` is more readable. Since `_` is quite usual for people to use in programs this might introduce collisions we did not had before. But I think the probability is very small (not sure if it is possible at all).
I would suggest the following: We use `_` as the default. If the `resolve` function is not able to make the names unique, then we rerun the algorithm with `$` as separator. I think the rerun will almost never happen, but just to make it more secure.
What do you think?

jesyspa:
I think _ is actully used less than you may think - it isn't idiomatic Kotlin.  We use it in tests, but honestly we should migrate to using backticks instead.

I don't think rerunning the algorithm makes sense - I would rather add more candidates with numbers.  But I would only worry about it when we have (a) users (b) that complain :')

PR #125, opened 2026-04-28
Unify ListSizeFieldEmbedding with SpecialField

komiputer:
## Summary

`ListSizeFieldEmbedding` and `SpecialField` were two near-duplicate `FieldEmbedding` shapes — both used `SpecialFieldName`, both had `ALWAYS_WRITEABLE` access policy, and the only structural differences were `includeInShortDump` and the optional non-negative-size invariant.

This PR extends `SpecialField` with two optional constructor parameters and re-expresses `ListSizeFieldEmbedding` as a `SpecialField` instance:

- `includeInShortDump: Boolean = false`
- `extraInvariantsBuilder: (FieldEmbedding) -> List<TypeInvariantEmbedding> = { listOf() }` — receives `this` so an invariant can refer to the field it lives on without a singleton self-reference.

The non-negative-size invariant is now an inline anonymous `TypeInvariantEmbedding` that closes over the `field` argument rather than referencing `ListSizeFieldEmbedding` by name.

## Test plan

- [x] `:formver.compiler-plugin:core:compileKotlin` succeeds.
- [x] `./gradlew :formver.compiler-plugin:test` passes (with `Z3_EXE` set).
- [x] `./gradlew :formver.compiler-plugin:testNoVerification` passes.


The-Ray-Man:
handled by #133

PR #127, opened 2026-04-30
Move some Conversion Tests to Verification

The-Ray-Man:
With this PR we want to update conversion tests such that they are also verified. This should help to detect for example issues with type information.

For this the following changes were made:
- the `conversion` directory is always excluded from verification
- Before per default only functions with a contract were sent to viper. Now every function is sent to viper. One can opt out with `@NeverVerify`

The number of test files could be reduced from 133 to 115


jesyspa (approved):
Awesome, thanks!

For my understanding: when do we still want to put tests in conversion?  Should there be some guidelines written down somewhere?

The-Ray-Man:
>For my understanding: when do we still want to put tests in conversion? Should there be some guidelines written down somewhere?

I think it makes sense for things were the viper run does not bring any benefits. Like for checking that some errors are detected correctly (e.g. invalid ADT)

jesyspa:
I think when it comes to error detection there's actually quite little value in putting it in conversion: generally we don't want to convert that code anyway.  So I suspect maybe just unifying everything and dropping the split would be easiest

PR #129, opened 2026-04-30
Extract locality checking into a dedicated compiler-plugin module

### LINKED ISSUE #119: Decouple locality checker implementation from the Viper and experimental Uniqueness checkers.

paganma:
This PR extracts the locality checking into a standalone formver.compiler-plugin:locality module and wire it into project/build settings. In addition it includes:

- A dedicated locality extension registrar with its own diagnostics and checker set, separated from the main formal-verification registrar. The registrar can take the class-id of the locality annotation as a configuration parameter.
- A new locality analysis/resolution components (GraphLocalityAnalyzer, GraphLocalityInfoResolver, declaration/property/expression locality resolvers) to compute locality through control-flow analysis as opposed to structurally.
- The ability to annotate locality through type attributes. An additional checker TypeLocalityAttributeChecker limits the targets of the locality attribute to only the types of function parameters, explicit receivers, and local variables.
- A few improvements including support for context parameters and correct handling of varargs.


formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
```diff
@@ -30,6 +27,11 @@ object VerificationErrors : KtDiagnosticsContainer() {
     val POSSIBLE_INDEX_OUT_OF_BOUND by warning2<PsiElement, String, String>()
     val UNEXPECTED_RETURNED_VALUE by warning1<PsiElement, String>()
     val INVALID_SUBLIST_RANGE by warning2<PsiElement, String, String>()
+    val PURITY_VIOLATION by error1<PsiElement, String>()
```
ligee, line 30: Should this and below be a part of this PR? It looks like it belongs to the uniquiness checker.

ligee (approved):
It might be a good idea to have tests, at least partially, in the locality plugin, so it could be moved/copied to the compiler branch, with some tests. But this should not block the PR. Please create an issue about it, so we do not forget to do it in the future.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
paganma: That looks like a rebase error. I will attempt to remove it.

PR #132, opened 2026-04-30
Optimized Short Name Resolver

### LINKED ISSUE #130: Short Name Resolver "nondeterministic"
With the new short name resolver, there are sometimes annoying changes. The short name resolver performs an iterative process to make names unique. Basically one collision is resolved after the other.  The order of resolving the collision can have an effect on the final result. This leads to the following problems:

- If a developer changes e.g. the order in which classes are embedded. Even though the list of embedded classes contain the same elements, it can still change the expected result.

- This was not tested, but probably also the order of method definitions in kotlin files can lead to the same effect.

The-Ray-Man:
With the new short name resolver, there are sometimes annoying changes. The short name resolver performs an iterative process to make names unique. Basically one collision is resolved after the other. The order of resolving the collision can have an effect on the final result. This leads to the following problems:

If a developer changes e.g. the order in which classes are embedded. Even though the list of embedded classes contain the same elements, it can still change the expected result.

This was not tested, but probably also the order of method definitions in kotlin files can lead to the same effect.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
@@ -79,7 +79,9 @@ class ShortNameResolver : NameResolver {
     /**
      * Stores all the names that exist in the system.
      */
-    private val elements: MutableSet<AnyName> = ViperKeywords.keywords.toMutableSet()
+    private val elements: MutableSet<AnyName> = sortedSetOf<AnyName>(compareBy { it.fullName() }).also {
+        it.addAll(ViperKeywords.keywords)
+    }
 
     fun elements() = elements.toList()
 
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
@@ -79,9 +79,7 @@ class ShortNameResolver : NameResolver {
     /**
      * Stores all the names that exist in the system.
      */
-    private val elements: MutableSet<AnyName> = sortedSetOf<AnyName>(compareBy { it.fullName() }).also {
-        it.addAll(ViperKeywords.keywords)
-    }
+    private val elements: MutableSet<AnyName> = ViperKeywords.keywords.toMutableSet()
 
     fun elements() = elements.toList()
 
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
@@ -146,9 +146,9 @@ class ShortNameResolver : NameResolver {
     override fun resolve() {
         var currentCollisions = collisions()
         while (currentCollisions.isNotEmpty()) {
-            val toResolve = currentCollisions.entries.first().value
+            val toResolve = currentCollisions.entries.minByOrNull { it.key }!!.value
 
-            val candidateToMove = toResolve.filter { canMove(it) }.ifNotEmpty { maxBy { priorityOrder(it) } }
+            val candidateToMove = toResolve.filter { canMove(it) }.sortedBy { it.hashCode() }.ifNotEmpty { maxBy { priorityOrder(it) } }
 
             assert(candidateToMove != null) { "Unable to make names unique" }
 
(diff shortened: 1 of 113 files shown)


jesyspa:
Meta-comment: PR titles, like commit messages, should be verb phrases: what does this do?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -144,9 +144,9 @@ class ShortNameResolver : NameResolver {
     override fun resolve() {
         var currentCollisions = collisions()
         while (currentCollisions.isNotEmpty()) {
-            val toResolve = currentCollisions.entries.first().value
+            val toResolve = currentCollisions.entries.minByOrNull { it.key }!!.value
 
-            val candidateToMove = toResolve.filter { canMove(it) }.ifNotEmpty { maxBy { priorityOrder(it) } }
+            val candidateToMove = toResolve.filter { canMove(it) }.sortedBy { it.hashCode() }.ifNotEmpty { maxBy { priorityOrder(it) } }
```
jesyspa: I actually had something a bit different in mind when designing the algorithm: I was thinking we have a graph of name dependencies, and we can consistently shift all minimal elements in the graph that are involved in conflicts.  That would give consistent names, and never give overly general names that can be mistaken: if x collides, it'll always be updated to a_x and b_x, rather than only one of the two getting updated.  WDYT?

The-Ray-Man:
# Overview of Changes

## Algorithm Update
In each iteration we do move (almost) all candidates. Such that names that collide are moved the same amount. For example, before if we had two anonymous names like `anon(1)` and `anon(2)`, if they collided only one was moved resulting in `anon_1` and `anon` as names. But what we actually want is to have: `anon_1` and `anon_2`. 

A name `A` is not moved, if `A` is part of another name `B` that should also be moved. So, if `B` is a name of the form: `B = SomeName(A, C)` and both `B` and `A` are part of collisions, only `B` will be moved. We do this, because otherwise the change of `B` is the result of two moves, which might result in non optimal names.

The heuristics for deciding which part should be moved was improved. For every name that is part of a collision, we check all the possible moves (moving the name itself, or moving a part of the name) and select the one that is associated with the lowest cost. The cost is a combination of the amount of names added as well as the kind of names added. For example adding the `PackageScope` is more expensive then adding a `NameType` to a name.  
 

## Candidate Name Changes
There were some names that had bad candidate names:
- `ExtensionReceiverName `: The name was just "this", but this collids with the `DispatchReceiverName`
- `PackageScope`, If the name was just the global package, the name returned is just a empty string. This resulted in a empty candidate name, which resulted in having `__` (two underscored) in the final viper name.
- `ScopedName`: The `FakeScope` and `BadScope` are now excluded from becoming part of a candidate.
- `TypeName`: The nullability information will always be part of the name.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -178,32 +229,23 @@ class ShortNameResolver : NameResolver {
...
-    private fun canMove(entity: AnyName): Boolean {
+    private fun findMoveableName(entity: AnyName): Pair<Int, AnyName>? {
         val currentIndex = currentCandidate[entity] ?: 0
-        if (currentIndex + 1 < entity.candidates().size) return true
-        return currentCandidate(entity).moveableParts().any {
-            canMove(it.name)
-        }
+        val options = currentCandidate(entity).moveableParts().mapNotNull { namePart ->
+            findMoveableName(namePart.name)
+        } + (if (currentIndex + 1 < entity.candidates().size) {
```
jesyspa: Maybe a helper function would be good?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -114,44 +113,96 @@ class ShortNameResolver : NameResolver {
...
     override fun resolve() {
         var currentCollisions = collisions()
         while (currentCollisions.isNotEmpty()) {
-            val toResolve = currentCollisions.entries.first().value
+            val allToMove = currentCollisions.flatMap { it.value }.mapNotNull { findMoveableName(it)?.second }.toSet()
+
+            val toMove = removeNonLeaves(allToMove)
+            assert(toMove.isNotEmpty()) { "No moveable names found, unable to resolve collisions" }
+
+            toMove.forEach { name ->
```
jesyspa: `toMove.forEach(::move)`, I think?

jesyspa (approved):
Looks good, some nits

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
The-Ray-Man: beautiful!

PR #133, opened 2026-04-30
Special Properties Unification

The-Ray-Man:
## Changes

- in #122 I did not find all the unclear names of functions. Therefore `lookupEmbedding()`  was renamed to `lookupClassTypeEmbedding()`
- Before the lookup process of special properties was a bit all over the place. This PR tries to unify this. 
  - The distinction between special properties and special fields was removed. Special fields can be represented as special properties using the `BackingFieldGetter/Setter` classes.
  - Every Special Property must be defined together with a `match` function. This function given the fir property symbol decided if its property should be used or not.
- All those special properties are collected in the object `SpecialProperties`. This object has a lookup method. Which is used by the `ProgramConverter`

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
@@ -245,15 +245,18 @@ class ProgramConverter(
 
     override fun embedProperty(symbol: FirPropertySymbol): PropertyEmbedding {
 
-        SpecialProperties.byCallableId[symbol.callableId]?.let {
-            return it
-        }
-
         if (symbol.receiverParameterSymbol != null) {
             return embedCustomProperty(symbol)
         } else {
+
             val name = symbol.embedMemberPropertyName()
             return typeResolver.getOrPutProperty(name) {
+                with(typeResolver) {
+                    with(session) {
+                        SpecialProperties.lookup(symbol)?.let { return@getOrPutProperty it }
+                    }
+                }
+
                 embedBackingField(symbol)?.let {
                     return@getOrPutProperty PropertyEmbedding(
                         BackingFieldGetter(it), BackingFieldSetter(it)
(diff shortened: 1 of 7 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -245,21 +245,27 @@ class ProgramConverter(
...
     override fun embedProperty(symbol: FirPropertySymbol): PropertyEmbedding {
 
-        SpecialProperties.byCallableId[symbol.callableId]?.let {
-            return it
-        }
-
         if (symbol.receiverParameterSymbol != null) {
             return embedCustomProperty(symbol)
         } else {
+
```
jesyspa: To be honest, I find these kinds of block start newlines really hurt readability :/ I haven't been indicating that you should remove them from your code, but I'd definitely avoid adding them just for the sake of it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/properties/SpecialProperties.kt
```diff
@@ -5,14 +5,59 @@
...
+import org.jetbrains.kotlin.formver.core.names.embedMemberBackingFieldName
+import org.jetbrains.kotlin.formver.core.names.embedName
+
+abstract class SpecialProperty(val property: PropertyEmbedding) {
+    context(typeResolver: TypeResolver, session: FirSession)
+    abstract fun match(symbol: FirPropertySymbol) : Boolean
+}
+
+
+val StringSizeProperty = object: SpecialProperty(PropertyEmbedding(LengthFieldGetter, setter = null)) {
```
jesyspa: Why not `object StringSizeProperty : SpceialProperty(...) { ... }`?

jesyspa (approved):
Looks good!  This should also close 125, right?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
The-Ray-Man: I will keep an eye out for them, and when I see them, I will remove them.

The-Ray-Man:
>This should also close 125, right?

Yes

PR #134, opened 2026-05-01
Drop havoc on field accesses through unique receivers

komiputer:
## Summary

- Wires the uniqueness analyzer's per-CFG-node results into Viper conversion for the first time. Two reads of `x.v` separated by a local-val (and writes through a unique receiver) now agree instead of bottoming out in a fresh havoc per access.
- Builds in three commits: read-side fix + plumbing, write-side mirror, and a function-scope unfold/fold hoist that keeps the emitted Viper compact enough for Silicon's evaluator at default JVM stack size.

## Background

The CLI plugin's linearizer was emitting `havocMethod.toMethodCall(...)` for every `var`-field read through a unique receiver (`AccessPolicy.BY_RECEIVER_UNIQUENESS` branch in `Linearizer.addFieldAccessStoringIn`), with the matching write branch dropping the assignment entirely. The TODO at `Linearizer.kt:107` ("Handling a unique field on a shared receiver must be added here") was the load-bearing comment. The uniqueness checker had the data needed to refine this — a per-CFG-node `UniquenessTrie` — but its output was thrown away when `UniquenessDeclarationChecker.check` returned, and `core` had no dep on `uniqueness`.

## What changes

**Architectural plumbing** (commit 1): new `UniquenessFacts` adapter in the `uniqueness` module exposes a per-`FirPropertyAccessExpression` query (`isReceiverUniqueAt`). `core` gains a `compileOnly` dep on `uniqueness`; `ProgramConverter.createBodyConversionContext` runs `UniquenessGraphAnalyzer` once per function body when `config.checkUniqueness` is on and stashes the facts on `MethodConversionContext`. `embedPropertyAccess` queries the facts and tags `ClassPropertyAccess` (and downstream `FieldAccess` / `FieldModification`) with `receiverIsUnique`. New `Linearizer.emitUniqueReceiverFieldRead` (commit 1) and `visitFieldModification.emitUniqueReceiverFieldWrite` (commit 2) emit unfold-of-unique-predicate / direct field op / fold instead of havoc / no-op.

**Function-scope hoist** (commit 3): the per-access unfold/fold cycles compounded into a deep evaluator tree on Silicon's side — `twoWritesOneRead` overflowed a default JVM thread stack on the Gradle compile path (the test fork uses `-Xss30M` and masked it). `SharedLinearizationState` now tracks `@Unique` parameters that have been registered for function-scope unfolding; `LinearizationVisitor.visitFunctionExp` registers them at body entry but defers the actual `Stmt.Unfold` until first access via `Linearizer.tryUseFunctionScopedReceiver`. The function epilogue and `Linearizer.addReturn` emit `Stmt.Fold` only for receivers that were actually unfolded — so functions that never touch the parameter emit no unfold/fold pair, keeping existing baselines quiet. Output for the pinned test cases collapses (e.g. `writeTwiceReadOnce` from nine statements to five in the unfold/op/fold band).

## Test surface

- New `// CHECK_UNIQUENESS` test directive enables the uniqueness checker alongside verification (without forcing the `NO_TARGETS` selection that `// UNIQUE_CHECK_ONLY` does).
- New `verification/uniqueness/unique_field_read.kt` covers `readMemberUnique(@Unique x: X)` and `constructAndRead` (`@Unique val x = X(42)`).
- New `verification/uniqueness/unique_field_write.kt` covers `writeThenRead`, `writeTwiceReadOnce`, `readWriteRead`, and `writeOneReadAnother` (no cross-talk between two unique receivers).
- Verified manually in a sandbox Gradle project at *default* JVM args — the multi-write cases pass without a `-Xss` override.
- Full `:formver.compiler-plugin:test` is green.

## Out of scope (flagged in commit bodies for whoever picks up next)

- Function calls within the body that consume a unique receiver (`@Unique` argument elsewhere) — needs facts-at-call-site to fold-before-call and drop the receiver from the unfolded set.
- `@Unique val` locals (non-parameter) — still on the per-access unfold/fold path; works for ≤2 accesses.
- Multi-step hierarchy in the unique-receiver predicate path — `emitUniqueReceiverFieldRead/Write` punt to havoc when the field's containing class isn't the receiver type itself.
- Constructor postcondition that establishes `bf$v == par$v` for primary-constructor field initializers — orthogonal to read-stability.
- Throw-on-first-violation in `UniquenessGraphChecker` — orthogonal smell.
- Promoting the existing `failing-tests/{binary_tree,linked_list}.kt` to positive tests — needs the `@Unique val`-local case (gap 2 above) plus a `// CHECK_UNIQUENESS` directive.

## Test plan

- [ ] `./gradlew :formver.compiler-plugin:test` passes
- [ ] `./gradlew publishToMavenLocal` succeeds
- [ ] `./gradlew compileKotlin` from a downstream Gradle project that sets `formver { checkUniqueness(true) }` and exercises a `@Unique` parameter with a `var` field passes verification at default JVM stack size
- [ ] Functions with `@Unique` parameters that *don't* access fields produce byte-identical Viper output to the previous baseline

### DIFF
--- a/formver.compiler-plugin/core/build.gradle.kts
+++ b/formver.compiler-plugin/core/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
 dependencies {
     compileOnly(project(":formver.common"))
     compileOnly(project(":formver.compiler-plugin:viper"))
+    compileOnly(project(":formver.compiler-plugin:uniqueness"))
     compileOnly(kotlin("compiler"))
 
     // TODO: figure out how to avoid this dependency
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/MethodConversionContext.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/MethodConversionContext.kt
@@ -18,6 +18,7 @@ import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
 import org.jetbrains.kotlin.formver.core.names.FunctionResultVariableName
 import org.jetbrains.kotlin.formver.core.names.ReturnLabelName
 import org.jetbrains.kotlin.formver.core.names.ReturnVariableName
+import org.jetbrains.kotlin.formver.uniqueness.UniquenessFacts
 
 
 class ReturnTarget private constructor(val variable: VariableEmbedding, val label: LabelEmbedding?) {
@@ -44,6 +45,13 @@ interface MethodConversionContext : ProgramConversionContext {
     val signature: FunctionSignature
     val defaultResolvedReturnTarget: ReturnTarget
     val isValidForForAllBlock: Boolean
+
+    /**
+     * Per-function uniqueness analysis result. Null when uniqueness checking is disabled,
+     * the function has no CFG, or the analyzer could not complete. Consumers must treat
+     * `null` as "no uniqueness information available" and fall back accordingly.
+     */
+    val uniquenessFacts: UniquenessFacts?
 
     fun resolveParameter(symbol: FirValueParameterSymbol): ExpEmbedding
     fun resolveLocal(symbol: FirVariableSymbol<*>): VariableEmbedding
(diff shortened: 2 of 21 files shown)


jesyspa:
Closing: superseded by #168, which took over the useful parts of this approach, and by the uniqueness infrastructure merged in #208. Any revival will happen on top of the new in-compiler uniqueness system.

PR #135, opened 2026-05-01
Support reference equality (=== and !==)

komiputer:
## Summary
- Adds an `IdentityCmp` expression embedding that linearizes to Viper `Ref` equality without unwrapping operands to their builtin type.
- Wires `FirOperation.IDENTITY` and `NOT_IDENTITY` in `StmtConversionVisitor`, alongside the existing `EQ` / `NOT_EQ` cases.
- New conversion-only test `operators/identity_equality.kt` covering object/object and object/null forms.

## Why a separate embedding rather than reusing `EqCmp`
`EqCmp` linearizes to a comparison on the *builtin* type when both operands match — fine for `==`, wrong for `===`. Even where they happen to coincide today, `convertEqCmp` carries a TODO to dispatch through `.equals()`; once that lands, `===` must remain raw `Ref` equality. Keeping the two embeddings distinct preserves that.

## Test plan
- [x] `./gradlew :formver.compiler-plugin:test` — full suite green
- [x] New goldens for `identity_equality.kt` show `boolToRef(x == y)` for `===` and `notBool(...)` for `!==`, with `nullValue()` on the null side

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
@@ -43,4 +43,24 @@ data class NeCmp(
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitNeCmp(this)
 }
 
+/**
+ * Reference equality (Kotlin `===`).
+ *
+ * Unlike [EqCmp], this never unwraps to a builtin type — operands are compared
+ * as Viper `Ref`s, so two distinct boxed values are not identified even when
+ * their unwrapped representations match.
+ */
+data class IdentityCmp(
+    val left: ExpEmbedding,
+    val right: ExpEmbedding,
+    override val sourceRole: SourceRole? = null,
+) : ExpEmbedding {
+    override val type
+        get() = buildType { boolean() }
+
+    override fun children(): Sequence<ExpEmbedding> = sequenceOf(left, right)
+
+    override fun <R> accept(v: ExpVisitor<R>): R = v.visitIdentityCmp(this)
+}
+
 fun ExpEmbedding.notNullCmp(): ExpEmbedding = NeCmp(this, NullLit)
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
@@ -21,6 +21,10 @@ sealed interface AnyComparisonExpression : ExpEmbedding {
 
     val comparisonOperation: Operator
 
+    /** Whether to compare the operands at their builtin type; they are compared as `Ref`s otherwise. */
+    val comparesUnwrapped: Boolean
+        get() = left.type == right.type
+
     override fun children(): Sequence<ExpEmbedding> = sequenceOf(left, right)
 }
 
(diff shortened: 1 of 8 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
```diff
@@ -43,4 +43,24 @@ data class NeCmp(
     override fun <R> accept(v: ExpVisitor<R>): R = v.visitNeCmp(this)
 }
 
+/**
+ * Reference equality (Kotlin `===`).
+ *
+ * Unlike [EqCmp], this never unwraps to a builtin type — operands are compared
```
jesyspa: Please narrow this down to what is actually needed for understanding the class; more comments aren't always better.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
komiputer: Trimmed to a single line stating the one thing the code does not say on its own — that the operands are compared as `Ref`s rather than unwrapped.

Reading the file to match its comment density, I noticed a better answer to this than shortening prose: `visitIdentityCmp` was carrying a near-verbatim copy of `comparisonLinearizable`. `AnyComparisonExpression` turned out to have exactly one use site (the linearizer), so `IdentityCmp` now implements it with `EqAny` as its comparison operation, and `comparisonLinearizable` takes `unwrapOperands` defaulting to whether the operand types agree. Reference equality passes `false`. Net −25 lines, and the distinction is now mostly stated in code.

Emitted Viper is unchanged by construction, and the goldens confirm it — `identity_equality.fir.diag.txt` needed no regeneration.

Happy to drop that second commit if you would rather keep the PR to what was originally reviewed; it does move the `===`/`==` distinction out of one self-contained block and into a flag read in another file.

komiputer:
Rebased onto `main` (was based on `04ef5b9e`, now on `6cc07d0b`). Clean — the two intervening commits touch CLI/gradle plumbing only, so nothing near the embeddings and no golden churn.

Verified locally with the same set CI runs: `detekt`, `apiCheck`, `:formver.compiler-plugin:locality:test`, `:formver.compiler-plugin:test` — all green, and `generateTests` produced no diff.

Review comment addressed in the thread above.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Comparison.kt
komiputer: Correcting my reply above — the version it describes is superseded.

I had threaded the unwrap/compare-as-`Ref` choice through `comparisonLinearizable` as a parameter defaulting to `left.type == right.type`. That was wrong on three counts: the default was one of the two policies rather than a neutral fallback, so `visitEqCmp` read as silent while `visitIdentityCmp` spelled its policy out; it moved evaluation of the type check from lowering time into the argument list for no gain; and `comparisonOperation` already establishes that this family carries its per-comparison lowering choice on the embedding, so that is where the flag belongs.

It is now a property of `AnyComparisonExpression`:

```kotlin
/** Whether to compare the operands at their builtin type; they are compared as `Ref`s otherwise. */
val comparesUnwrapped: Boolean
    get() = left.type == right.type
```

`IdentityCmp` overrides it to `false` beside `comparisonOperation = EqAny`. `comparisonLinearizable` takes a single argument again, all three `visit*` methods are the same one-liner, and the check is read inside `toViperBuiltinType` as it originally was.

That also settles your original point more cleanly: the KDoc is now just `/** Kotlin reference equality (`===`). */`, because the clause about not unwrapping is the line `override val comparesUnwrapped = false` two lines below it.

komiputer:
Rebased again — `main` had moved to `270c45c5` and the PR had gone to a conflicting state, which is why no CI ran on the previous head.

The interesting collision is with `270c45c5` (Enforce ExpVisitor coverage at compile time). This PR had added `visitIdentityCmp` to `ExpVisitor` with a `= visitDefault(e)` body, which is exactly the construct that commit removes. Adapted as follows:

- `ExpVisitor` gets an abstract `fun visitIdentityCmp(e: IdentityCmp): R`
- `DefaultingExpVisitor` gets `override fun visitIdentityCmp(e: IdentityCmp): R = visitDefault(e)`
- `LinearizationVisitor` and `ExprPurityVisitor` already override it explicitly, so they satisfy the full-coverage interface; `DebugTreeViewVisitor` picks up the fallback via `DefaultingExpVisitor`

The PR is better for it — reference equality is now a case a future visitor cannot silently skip.

Also resolved two import conflicts in `LinearizationVisitor`, where neither side survives: `main` dropped `predicateAccess` (moved into `FieldAccessLowering` in `3ad85cca`) and this branch no longer needs `EqAny` now that the shared helper uses `comparisonOperation`.

`identity_equality.fir.diag.txt` still needed no regeneration across both rebases, which is the check that the emitted Viper is unaffected by any of this.

Green locally on `detekt`, `apiCheck`, `:formver.compiler-plugin:locality:test`, `:formver.compiler-plugin:test`, and CI is now green on `fbee8e6a`.

jesyspa (approved):

PR #139, opened 2026-05-06
Support product ADTs based on data classes

pawinkler:
**Summary**:

- Extends ADT support from singleton to product types by allowing data classes with immutable constructor fields
- Adds ADT field naming infrastructure (`AdtFieldName`, `AdtFieldEmbedding`) and constructor/destructor embeddings for fields and property accesses
- Refactors ADT validation into `validateAdtSignature` and `validateAdtProperties` with non-short-circuiting error collection and simplifies ADT diagnostic
- Adds Viper `AdtDestructorApp` AST node and runtime type axioms for destructor fields

**Scope limitations**:

- Does not concern equality
- Does not concern casts
- Does not address re-evaluation of ADTs resulting in multiple error messages

**PR Structure**:

The commits are small, self-contained chunks which hopefully are individually reviewable. They concern:

1. **Added ADT field naming infrastructure** — `AdtFieldName`, `NameType.AdtField`, resolver/renderer support
2. **Refactored and extended ADT validation** — validation split, error consolidation, new checks (`var`, body fields, generics, supertypes, secondary constructors)
3. **ADT Declaration supports fields** — field registration in `TypeResolver`, `AdtFieldEmbedding`, constructor decl now carries field parameters
4. **Mapped constructor calls** — `AdtConstructorRef` now carries args, `StmtConversionVisitor` intercepts primary constructor calls on ADT types
5. **Added ADT deconstructors** — `AdtFieldAccess` expression node, `AdtDestructorApp` Viper AST, linearization of field access via `fromRef` unwrap + destructor app
6. **Axiomatization** — runtime type domain axiom for destructor field types


pawinkler:
A few things I realized while reviewing myself:

- There was a naming issue in the earlier commits, which resulted in complicated ADT names. It's later resolved.
- The comment regarding the automatically generated functions is not correct, but the underlying reasoning holds. I will fix it with the next commit.
- The line count is very high. This mostly due to a full embedding I included to assess the ADT encoding. Should I remove this?

jesyspa:
> There was a naming issue in the earlier commits, which resulted in complicated ADT names. It's later resolved.

I generally do not think reviewing commit-by-commit is very helpful - if there is a change that can be extracted out, it's better to move it into its own PR - but if I did, this kind of issue would defeat the point of the split.

> The comment regarding the automatically generated functions is not correct, but the underlying reasoning holds. I will fix it with the next commit.

If you are planning to still make changes, then the PR is not ready for review.  Please first make your changes and only then request a review.

> The line count is very high. This mostly due to a full embedding I included to assess the ADT encoding. Should I remove this?

In principle this is not an issue, but there is a whole bunch of duplicate code there.  Please fix that.

pawinkler:
> I generally do not think reviewing commit-by-commit is very helpful - if there is a change that can be extracted out, it's better to move it into its own PR - but if I did, this kind of issue would defeat the point of the split.

Okay thanks, makes sense. To get a better understanding, would you recommend splitting these commits into different PRs? I considered them to be logically connected, so a split wouldn't help understanding it, but this might be wrong.

> If you are planning to still make changes, then the PR is not ready for review. Please first make your changes and only then request a review.

I think this was a misunderstanding. The PR is done, but I considered this to be a minor mistake which I would have fixed while addressing your feedback or before the final merge.

> [...] whole bunch of duplicate code there. Please fix that

I'm sorry, I don't understand what exactly is meant here. Is the implementation inefficient and uses a lot of duplicate patterns, or is it regarding the general architecture? I tried to make it as efficient as possible and to include it nicely into the existing architecture, so hopefully there shouldn't be a lot of duplication.

formver.common/src/org/jetbrains/kotlin/formver/common/ErrorCollector.kt
jesyspa, line 1: The changes to error-handling look independent from the rest and could probably be sent as a separate PR.  It would be a good opportunity to explain why the switch is a good idea, given you just setup error in the previous PR.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -98,7 +98,9 @@ class ProgramConverter(
                     )
                 }
             },
-            adts = typeResolver.adtTypeEmbeddings().filter { it.isValid }.map { it.toViper() },
+            adts = typeResolver.adtTypeEmbeddings().filter { it.isValid }.map { embedding ->
+                embedding.toViper(typeResolver.lookupAdtFields(embedding.name))
+            },
```
jesyspa: I don't think it's a great idea to glue things together at this stage.  Is there a reason this information isn't already linked e.g. at the TypeResolver level?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -116,11 +118,10 @@ class ProgramConverter(
         // Ensures every function that touches an invalid ADT (signature, body, transitive callee, transitive field)
         // receives an ADT usage error to communicate the misuse to the user.
         // TODO: Move this to the linearizer level and annotate the specific touch instead
-        if (errorCollector.collectedAdtErrorOfKind(AdtErrorKind.INVALID_TARGET)) {
+        if (errorCollector.collectedAdtError() && declaration.origin == FirDeclarationOrigin.Source) {
```
jesyspa: Two things:
- If you want to do this, leave a comment about the motivation for suppressing errors for other classes of things.
- This looks like the wrong place to do this.  If you don't care about errors from other things, why are you verifying them?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -254,30 +255,93 @@ class ProgramConverter(
     private fun embedAdtClass(symbol: FirRegularClassSymbol): AdtTypeEmbedding {
         val className = symbol.classId.embedName()
         typeResolver.lookupAdt(className)?.let { return it }
-        return AdtTypeEmbedding(className, isValid = validateAdt(symbol)).also { typeResolver.registerAdt(it) }
+
+        val adtName = AdtName(className)
+        val isValid = validateAdtSignature(symbol) && validateAdtProperties(symbol, adtName)
+        return AdtTypeEmbedding(className, isValid = isValid).also { typeResolver.registerAdt(it) }
```
jesyspa: I don't think this approach to ADT error handling is particularly flexible.  You are still creating ADTs, but now you need to remember to call `isValid` everywhere.  I would strongly encourage a refactor where you use a representation that doesn't have this shortcoming (e.g. how we did with pure functions), but it's not blocking here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -254,30 +255,93 @@ class ProgramConverter(
...
+            isValid = false
+        }
+        return isValid
+    }
+
+    @OptIn(DirectDeclarationsAccess::class)
+    private fun validateAdtProperties(symbol: FirRegularClassSymbol, adtName: AdtName): Boolean {
+        var isValid = true
+        val checkedPropertyNames = mutableSetOf<String>()
         for (declaration in symbol.declarationSymbols) when (declaration) {
```
jesyspa: I missed this on previous reading, but putting a for and a when on one line like this is not good style.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -254,30 +255,93 @@ class ProgramConverter(
...
+                    isPropertyValid = false
+                }
+                if (isPropertyValid) {
+                    typeResolver.getOrPutAdtField(declaration.embedMemberPropertyName()) {
+                        AdtFieldEmbedding(
+                            name = AdtFieldName(adtName, SimpleKotlinName(declaration.name)),
+                            type = embedType(declaration.resolvedReturnType),
+                        )
+                    }
+                }
```
jesyspa: If your function is called `validate`, it should not be assembling some extra state on the side.  You also have an asymmetry here: for ADTs you do validation then add them separately, while for ADT properties you've merged it.  Please refactor this to be clear and consistent.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -254,30 +255,93 @@ class ProgramConverter(
...
+            )
+            isValid = false
+        }
+        return isValid
+    }
+
+    @OptIn(DirectDeclarationsAccess::class)
+    private fun validateAdtProperties(symbol: FirRegularClassSymbol, adtName: AdtName): Boolean {
+        var isValid = true
+        val checkedPropertyNames = mutableSetOf<String>()
```
jesyspa: Under what conditions might you see a property twice?  This looks quite strange and probably deserves a comment.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -254,30 +255,93 @@ class ProgramConverter(
     private fun embedAdtClass(symbol: FirRegularClassSymbol): AdtTypeEmbedding {
         val className = symbol.classId.embedName()
         typeResolver.lookupAdt(className)?.let { return it }
-        return AdtTypeEmbedding(className, isValid = validateAdt(symbol)).also { typeResolver.registerAdt(it) }
+
+        val adtName = AdtName(className)
+        val isValid = validateAdtSignature(symbol) && validateAdtProperties(symbol, adtName)
```
jesyspa: Is the short-circuiting here intentional?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -131,10 +132,22 @@ fun StmtConversionContext.embedPropertyAccess(accessExpression: FirPropertyAcces
...
             when {
                 accessExpression.dispatchReceiver != null -> {
-                    val property = calleeSymbol.findFinalParentProperty()?.let {
-                        embedProperty(it)
-                    } ?: embedProperty(calleeSymbol)
-                    ClassPropertyAccess(convert(accessExpression.dispatchReceiver!!), property, type)
+                    val adtPretype = embedType(accessExpression.dispatchReceiver!!.resolvedType).pretype as? AdtTypeEmbedding
+                    if (adtPretype != null) {
+                        val field = typeResolver.lookupAdtFields(adtPretype.name)
+                            .firstOrNull { it.name.propertyName.name == calleeSymbol.name }
```
jesyspa: If you would like them to be looked up by names, why store them as a list?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -131,10 +132,22 @@ fun StmtConversionContext.embedPropertyAccess(accessExpression: FirPropertyAcces
             val type = embedType(calleeSymbol.resolvedReturnType)
             when {
                 accessExpression.dispatchReceiver != null -> {
-                    val property = calleeSymbol.findFinalParentProperty()?.let {
-                        embedProperty(it)
-                    } ?: embedProperty(calleeSymbol)
-                    ClassPropertyAccess(convert(accessExpression.dispatchReceiver!!), property, type)
+                    val adtPretype = embedType(accessExpression.dispatchReceiver!!.resolvedType).pretype as? AdtTypeEmbedding
```
jesyspa: This looks backwards: instead of looking up the property directly and deciding based on its type, you look up the type and then start digging through its properties.  It would be better to do it the same way we do it for classes (ideally, without even having to touch this code).

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -300,6 +300,14 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
         val symbol = functionCall.toResolvedCallableSymbol() as? FirFunctionSymbol<*>
             ?: throw NotImplementedError("Only functions are expected as callables of function calls, got ${functionCall.toResolvedCallableSymbol()}")
 
+        if (symbol is FirConstructorSymbol && symbol.isPrimary) {
```
jesyspa: Similarly here: why should this be handled here, rather than in `embedAnyFunction`?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
```diff
@@ -413,5 +417,17 @@ class RuntimeTypeDomain(typeResolver: TypeResolver) : BuiltinDomain(DomainName(R
                 }
             }
         }
+        // Encodes type information of ADT deconstructors
+        adtFields.forEach { (embedding, fields) ->
```
jesyspa, line 420: This looks a bit strange.  Why have one axiom for all fields, rather than an axiom per field?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
```diff
@@ -67,6 +67,7 @@ data object NullLit : LiteralEmbedding {
  */
 data class AdtConstructorRef(
     override val type: TypeEmbedding,
+    val args: List<ExpEmbedding> = emptyList(),
```
jesyspa: Is there value to making this optional?  It looks like something that's fine to keep explicit.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
```diff
@@ -8,19 +8,32 @@ package org.jetbrains.kotlin.formver.core.embeddings.types
...
-        AdtConstructorDecl(AdtConstructorName(adtName, name), adtName, emptyList())
+    fun viperConstructorDecl(fields: List<AdtFieldEmbedding>): AdtConstructorDecl =
+        AdtConstructorDecl(
+            AdtConstructorName(adtName, name),
+            adtName,
+            fields.map { Declaration.LocalVarDecl(it.name, Type.Ref) },
+        )
 
-    fun toViper() = AdtDecl(name = adtName, constructors = listOf(viperConstructorDecl))
+    fun toViper(fields: List<AdtFieldEmbedding>) = AdtDecl(
```
jesyspa: Our `toViper` functions generally do not take parameters.  I don't think it makes sense here, either.  (I also think that naming what you're mapping to would be good; in this case, a declaration.)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/AdtTypeEmbedding.kt
```diff
@@ -8,19 +8,32 @@ package org.jetbrains.kotlin.formver.core.embeddings.types
...
 data class AdtTypeEmbedding(override val name: ScopedName, val isValid: Boolean) : PretypeEmbedding {
     val adtName: AdtName = AdtName(name)
     val viperType: Type.Adt = Type.Adt(adtName)
     val injection: Injection = Injection(name, viperType, RuntimeTypeDomain.classTypeFunc(name))
 
     override val runtimeType = RuntimeTypeDomain.classTypeFunc(name)()
 
-    val viperConstructorDecl: AdtConstructorDecl =
-        AdtConstructorDecl(AdtConstructorName(adtName, name), adtName, emptyList())
+    fun viperConstructorDecl(fields: List<AdtFieldEmbedding>): AdtConstructorDecl =
```
jesyspa: Function names should generally contain a verb.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/NameType.kt
```diff
@@ -32,6 +32,7 @@ sealed interface NameType : NameTypeBase {
                 Base.DomainFunction -> "df"
                 Base.Adt -> "adt"
                 Base.AdtCons -> "constr"
+                Base.AdtField -> "af"
```
jesyspa: It would be good to be consistent, e.g. `ac` and `af` (or even `adtc` and `adtf`)

formver.compiler-plugin/testData/diagnostics/conversion/adts/full_embedding.fir.diag.txt
jesyspa, line 1: This file contains very long and repetitive `VIPER_TEXT` for four functions, while one seems enough.

formver.compiler-plugin/testData/diagnostics/verification/adts/product_errors.kt
```diff
@@ -0,0 +1,50 @@
+// NEVER_VALIDATE
+// FULL_JDK
```
jesyspa, line 2: Is `FULL_JDK` needed for data classes?

formver.compiler-plugin/testData/diagnostics/conversion/adts/data_classes.kt
```diff
@@ -0,0 +1,50 @@
+// NEVER_VALIDATE
+// FULL_JDK
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+@ADT
+data object Tag
+@ADT
+data class TwoFields<!VIPER_TEXT!>(val <!VIPER_TEXT!>a<!>: Int, val <!VIPER_TEXT!>b<!>: Boolean)<!>
```
jesyspa: I don't see the point in generating Viper for `a` and `b`, or for the class.  What are you verifying?

formver.compiler-plugin/testData/diagnostics/conversion/adts/data_classes.kt
```diff
@@ -0,0 +1,50 @@
+// NEVER_VALIDATE
+// FULL_JDK
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+@ADT
+data object Tag
```
jesyspa: What is this needed for?

formver.compiler-plugin/testData/diagnostics/verification/adts/product.kt
jesyspa, line 1: Do you already support recursive types?  If so, it would be good to test them.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
jesyspa: This function looks like it should be relevant for you, but you don't seem to use it anywhere. Is there logic you're duplicating because of that?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -131,10 +132,22 @@ fun StmtConversionContext.embedPropertyAccess(accessExpression: FirPropertyAcces
...
+                    if (adtPretype != null) {
+                        val field = typeResolver.lookupAdtFields(adtPretype.name)
+                            .firstOrNull { it.name.propertyName.name == calleeSymbol.name }
+                            ?: throw SnaktInternalException(
+                                accessExpression.source,
+                                "Property ${calleeSymbol.name} does not exist on ADT ${adtPretype.name}",
```
jesyspa: This can happen if your ADT is invalid. Please take this into account when redesigning error handling for ADTs.

jesyspa (changes requested):
Many small-ish issues, I think the main problems are things being done in the wrong place or being done through a shortcut instead of a proper solution.

> To get a better understanding, would you recommend splitting these commits into different PRs?

Typically, you want to identify individual contributions and send PRs for them separately.  Here, I think the two main ones are changes to the error-handling (which I think can be redesigned further) and the actual feature contribution.

> I'm sorry, I don't understand what exactly is meant here. Is the implementation inefficient and uses a lot of duplicate patterns, or is it regarding the general architecture? 

Of the +2k lines added, +1.2 is duplication because of unnecessary VIPER_TEXTs.  Once that's accounted for, the PR isn't that big (though I would still prefer to include ADTs into the existing FULL_VIPER_DUMP rather than generating a new one).  The actual effect on the code is something like 200-300 lines touched, which is not much at all.

formver.common/src/org/jetbrains/kotlin/formver/common/ErrorCollector.kt
pawinkler: Alright, I will move it to a new PR with the reasoning.

formver.compiler-plugin/testData/diagnostics/verification/adts/product_errors.kt
pawinkler: Yes, without it we get the following error: `Cannot access 'Serializable' which is a supertype of 'Int'. Check your module classpath for missing or conflicting dependencies.`

formver.compiler-plugin/testData/diagnostics/conversion/adts/data_classes.kt
pawinkler: This is a result from the automatically generated `component1` and `component2` member functions in the data class. The source is automatically set to the field they are returning and it's not related to the ADTs. Is this acceptable or should I implement some special treatment here?

formver.compiler-plugin/testData/diagnostics/verification/adts/product.kt
pawinkler: We do, but the main proof with the linked list requires equality, which is still missing. I could create an additional, simpler proof with a recursive ADT. However, I'm not sure if it's actually needed, everything should be covered by the linked list example.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: Thanks, this should be helpful.

formver.compiler-plugin/testData/diagnostics/conversion/adts/data_classes.kt
pawinkler: Sorry, this was a relict from a test case I removed earlier.

pawinkler:
> +1.2 is duplication because of unnecessary VIPER_TEXTs. Once that's accounted for, the PR isn't that big (though I would still prefer to include ADTs into the existing FULL_VIPER_DUMP rather than generating a new one). The actual effect on the code is something like 200-300 lines touched, which is not much at all.

An underlying problem here are automatically generated functions (like `copy()` and `componentN()`). They are responsible for the additional Viper texts and also the duplicated error messages, as each function is converted in a modular way. I could suppress the conversion and validation of these, but this would probably be a different PR, as it requires some new kind of annotation to indicate it.

formver.compiler-plugin/testData/diagnostics/conversion/adts/full_embedding.fir.diag.txt
pawinkler: This is related to this [comment](https://github.com/JetBrains/SnaKt/pull/139#issuecomment-4421410273).

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/RuntimeTypeDomain.kt
pawinkler: I considered it to be less cluttering in the domain, especially for ADTs with many fields.

jesyspa: I don't think this is a good idea, since it means you put something much larger in the SMT context whenever one of the triggers fires.  I don't see any downside to having multiple axioms.

formver.compiler-plugin/testData/diagnostics/verification/adts/product_errors.kt
jesyspa: Please add a comment to our documentation that describes how to decide whether to include `FULL_JDK` in a test.

formver.compiler-plugin/testData/diagnostics/conversion/adts/data_classes.kt
jesyspa: Is there a reason to verify these functions?

formver.compiler-plugin/testData/diagnostics/verification/adts/product.kt
jesyspa: I don't think a proof is required, but you do have to test that conversion works.

jesyspa (changes requested):
I don't think there's any new annotation needed for component functions; either there is something we want to verify about them (in which case, okay, but we should be clear about what) or we should skip verification.  I suspect the latter is the case, but let me know if there's some reason to keep it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: I agree, this was a prevention of surfacing errors from compiler-inferred functions. I've now blocked the verification of those and removed this change

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: Regarding your question, I don't think so. We are iterating over all properties anyways to check the ADT validity and register them on the fly. This method cannot be used to replicate the behavior and I don't think a strict separation between both is necessary.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
pawinkler: It seemed more aligned with the previous handling of the ADT qualifiers, but I agree, this is not the best way to implement this. I've updated it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Literal.kt
pawinkler: We didn't need it for singleton ADTs, so I thought I make it optional here instead of passing an empty list there.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -279,43 +279,108 @@ class ProgramConverter(
...
+            typeResolver.registerAdt(name, InvalidAdtTypeEmbedding)
+            return InvalidAdtTypeEmbedding
+        }
+        val allMembersValid = symbol.declarationSymbols.fold(true) { acc, sym ->
+            acc and validateAdtMember(sym)
+        }
+        if (!allMembersValid) {
+            typeResolver.registerAdt(name, InvalidAdtTypeEmbedding)
+            return InvalidAdtTypeEmbedding
+        }
```
jesyspa, line 293: This looks like it would be good to split into its own function

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -279,43 +279,108 @@ class ProgramConverter(
...
+            .filterIsInstance<FirConstructorSymbol>()
+            .firstOrNull { it.isPrimary }
+            ?.let { cs ->
+                methods[cs.embedName(this)] =
+                    AdtConstructorEmbedding(adtEmbedding, typeResolver.lookupAdtFields(adtEmbedding.name))
+            }
+        return adtEmbedding
+    }
+
     private fun validateAdtHeader(symbol: FirRegularClassSymbol): Boolean {
```
jesyspa, line 321: Kotlin doesn't use "header" as a standard term.  What do you mean by this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -44,6 +46,16 @@ class TypeResolver {
      */
     private val properties = mutableMapOf<ClassPropertyPair, PropertyEmbedding>()
 
+    /**
+     * All ADT fields, keyed by (class, property) pair for deduplication.
+     */
+    private val adtFields = mutableMapOf<ClassPropertyPair, AdtFieldEmbedding>()
```
jesyspa: Similarly to what we discussed previously: why is this special case special enough to distinguish it from other property embeddings?

formver.compiler-plugin/testData/diagnostics/verification/full_viper_dump.kt
```diff
@@ -1,11 +1,17 @@
 // FULL_VIPER_DUMP
+// FULL_JDK
```
jesyspa, line 2: Your comment in the README says FULL_JDK should be used if the test uses Java standard library types.  Which standard library types are you using here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -279,43 +279,108 @@ class ProgramConverter(
...
+        }
+
+        // 2. Embed ADT and members
+        val adtEmbedding = AdtTypeEmbeddingImpl(name)
+        typeResolver.registerAdt(name, adtEmbedding)
+        symbol.declarationSymbols.filterIsInstance<FirPropertySymbol>().forEach {
+            embedAdtField(it, adtEmbedding)
+        }
+        // Proactively embed the constructor to support ADT-specific generation
+        symbol.declarationSymbols
```
jesyspa, line 302: Why is this necessary for ADTs and not for other classes?

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: For example, for `@ADT data class Point(val a: Int, val b: Int): Geometry`, my "header" would correspond to `@ADT data class Point: Geometry`. I didn't really find a name for it, it's a mix of different Kotlin concepts.

formver.compiler-plugin/testData/diagnostics/verification/full_viper_dump.kt
pawinkler: I'm implicitly using `Serializable` for `Int`s. It's a result of the combination of the ADT verification and the `data` modifier. By iterating over the properties in the check, we force the evaluation of synthetic members by the Kotlin compiler, which in turn accesses the field types and triggers the evaluation of their supertypes. For example, `Pair.hashCode()` relies on `Int.hashCode()`, which relies on `Serializable`.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: I thought it was a good idea, because it doesn't really correspond to a function. Unlike with normal constructors, we do not actually call the corresponding function, but instead insert the `AdtConstructorEmbedding` which is translated fundamentally different. 

If I use the normal `embedAnyFunction` path, I either have to diverge directly and return the `AdtConstructorEmbedding`, which requires a case distinction on types on a level where we usually avoid these, or make significant changes to how we currently handle pure functions for one special case which shouldn't occur otherwise. I thought this would be the better approach here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
pawinkler: I agree, this is not necessary. I've removed it. Just to be sure, is the `adtFieldsByAdt` map fine? I didn't really see a way to get rid of it, as we need it for the construction of the ADT declaration.

PR #143, opened 2026-05-07
Cleanup Method Embeddings

The-Ray-Man:
In this PR we added the return target to the function signature. This allowed us to simplify some things:
- the `ContractDescriptionConversionVisitor` does not need a context anymore, because the necessary information is included in the signature.
- Since we do know the return variable, while we convert the header of a function, it is no longer necessary to delay the pre+postcondition extraction in `FullNamedFunctionSignature `. 

Additionally a `isPure` flag was added to the function signature. This also simplifies logic, because once we have the signature we do not need the `firSession` anymore to know if it is pure or not. 

We created classes for the Constructor method as well as normal user method. (Instead of creating objects on the fly). This introduced some duplicated code for creating the pre+post conditions. This will be removed in a future PR.

Note: The distinction between `ReturnVariableName(override val n: Int)` and `PlaceholderReturnVariableName` is quite annoying. Because for methods without a body we are suppost to use `PlaceholderReturnVariableName`, but during the creation of a the signature it is not always clear if this will have a body or not. Maybe just always use the numbered version with incremental numbers would be easier.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionSignature.kt
```diff
@@ -19,6 +20,10 @@ interface FunctionSignature {
 
     val params: List<VariableEmbedding>
 
+    val returns : ReturnTarget
```
jesyspa: I'm not a fan of putting the label here, since it's not something anything outside the function needs access to.  Maybe it's best to only put the variable here, and then forward it to returnTarget.variable in cases when that's possible?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
```diff
@@ -10,33 +10,119 @@ import org.jetbrains.kotlin.fir.symbols.impl.FirFunctionSymbol
...
 
     val declarationSource: KtSourceElement?
 }
 
+
+class ConstructorSignature(
+    val signature : NamedFunctionSignature,
+    val propertiesPostconditions : List<ExpEmbedding>,
+    override val symbol: FirFunctionSymbol<*>,
+    typeResolver: TypeResolver
```
jesyspa: I have to say, I am not super happy with this, but I'm not sure how to do it better.  I guess I just don't like constructors doing actual work.  Maybe it's fine, though.   That said, splitting out some of the logic would make it easier to share with constructors.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -521,7 +521,8 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
         data: StmtConversionContext,
     ): ExpEmbedding {
         val function = anonymousFunctionExpression.anonymousFunction
-        return LambdaExp(data.embedFunctionSignature(function.symbol), function, data, function.symbol.label!!.name)
+        val signature = data.embedFunctionSignature(function.symbol)
+        return LambdaExp(signature, function, data, function.symbol.label!!.name)
```
jesyspa, line 525: Any way we could avoid the `label!!`, since you're doing a cleanup anyway?  (if it's out of scope, also fine)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/MethodConversionContext.kt
```diff
@@ -9,20 +9,43 @@ import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
...
+class ReturnTargetImpl(depth: Int, type: TypeEmbedding) : ReturnTarget {
+    override val variable = PlaceholderVariableEmbedding(ReturnVariableName(depth), type)
+    override val label = LabelEmbedding(ReturnLabelName(depth))
 }
 
+class ReturnTargetNoLabel(override val variable: VariableEmbedding) : ReturnTarget {
+    companion object {
+        fun forPureFunction(type: TypeEmbedding) = ReturnTargetNoLabel(PlaceholderVariableEmbedding(
+            FunctionResultVariableName, type))
+        fun forNoBodyFunction(type: TypeEmbedding) = ReturnTargetNoLabel(PlaceholderVariableEmbedding(
```
jesyspa: I get what you mean, but it doesn't seem too bad...

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/MethodConversionContext.kt
```diff
@@ -9,20 +9,43 @@ import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
...
 import org.jetbrains.kotlin.formver.core.names.ReturnLabelName
 import org.jetbrains.kotlin.formver.core.names.ReturnVariableName
 
-class ReturnTarget(depth: Int, type: TypeEmbedding) {
-    val variable = PlaceholderVariableEmbedding(ReturnVariableName(depth), type)
-    val label = LabelEmbedding(ReturnLabelName(depth))
+
+interface ReturnTarget {
+    val variable: VariableEmbedding
+    val label: LabelEmbedding
```
jesyspa: If you do want to keep it, maybe make it nullable?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -5,51 +5,42 @@
...
  * Conversions for global entities like types should generally be
  * performed via this context to ensure they can be deduplicated.
  * We need the FirSession to get access to the TypeContext.
  */
 class ProgramConverter(
-    val session: FirSession, override val config: PluginConfiguration, override val errorCollector: ErrorCollector
+    val session: FirSession,
+    override val config: PluginConfiguration,
+    override val errorCollector: ErrorCollector,
+    val declaration: FirSimpleFunction
```
jesyspa: I don't think this is a good idea.  The whole idea of `registerForVerification` is that we can verify multiple things.  This basically makes that impossible.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -317,24 +355,22 @@ class ProgramConverter(
...
         val extensionReceiverType = symbol.extensionReceiverType
         val isExtensionReceiverUnique = symbol.receiverParameterSymbol?.isUnique(session) ?: false
         val isExtensionReceiverBorrowed = symbol.receiverParameterSymbol?.isBorrowed(session) ?: false
-        return object : FunctionSignature {
+
+        val returnType = embedType(symbol.resolvedReturnType)
+
+        val returnTarget = when {
+            symbol.isPure(session) -> ReturnTargetNoLabel.forPureFunction(returnType)
+            isVerifiedMethod(symbol) -> returnTargetProducer.getFresh(returnType)
```
jesyspa: I'm fine with using numbered return variables, yeah.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -387,90 +420,80 @@ class ProgramConverter(
...
-                firSpec?.precond?.let {
-                    addAll(createCtx().collectInvariants(it))
-                }
-            }
+    private fun embedKotlinContract(
+        signature: NamedFunctionSignature
+    ): Pair<List<ExpEmbedding>, List<ExpEmbedding>> {
+        val contractVisitor = ContractDescriptionConversionVisitor(this@ProgramConverter, signature)
+        val postConditions = contractVisitor.getPostconditions()
+        return Pair(emptyList(), postConditions)
```
jesyspa: If it can never return preconditions, maybe just postconditions would be cleaner?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -387,90 +420,80 @@ class ProgramConverter(
...
-                // TODO: this process should be a separate step in the conversion.
-                firSpec?.precond?.let {
-                    addAll(createCtx().collectInvariants(it))
-                }
-            }
+    private fun embedKotlinContract(
+        signature: NamedFunctionSignature
+    ): Pair<List<ExpEmbedding>, List<ExpEmbedding>> {
+        val contractVisitor = ContractDescriptionConversionVisitor(this@ProgramConverter, signature)
+        val postConditions = contractVisitor.getPostconditions()
```
jesyspa: nit: postconditions is one word, no need for camelCase

jesyspa (changes requested):
Thanks, this looks much better!  Still some stuff to discuss I think, so flagging this for another round of review.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
The-Ray-Man: Some design that I experimented with was to have "ContractClasses", so that we split up the pre+post conditions into smaller contracts where each collects some type of invariants. For example there could be a "pureContract", "accessContract", "constructorFieldContract" etc. 
And when embedding a method, we just give a list of those objects, which are used to generate the pre+post conditions. With this I don't have that much logic in the constructor and can also share logic between different methods. 
Basically in a similar way on how I did it with the pre-postconditions provided by the user.

jesyspa: I think a straightforward way would be to just make pre/postconditions constructor parameters and make a function (free or in a companion object) that constructs this.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
The-Ray-Man: At the moment I don't see an easy way to do this. (this label is from fir and not from us)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionSignature.kt
The-Ray-Man: I did something similar: When embedding a signature, we create the returnTarget (with the fresh entity producer). And then the variable is added to the signature and the returnTarget together with the signature is returned from the embedding function. If we also need to embed the body of the method, then we can use the return target to create the corresponding conversion context.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -137,34 +121,81 @@ class ProgramConverter(
...
     }
 
-    fun embedUserFunction(symbol: FirFunctionSymbol<*>, signature: FullNamedFunctionSignature): UserFunctionEmbedding {
-        (methods[signature.name] as? UserFunctionEmbedding)?.also { return it }
-        val new = UserFunctionEmbedding(embedCallable(symbol, signature))
-        methods[signature.name] = new
+    fun embedUserFunction(declaration: FirSimpleFunction): UserFunctionEmbedding {
+        val symbol = declaration.symbol
+        val name = symbol.embedName(this)
+        (methods[name] as? UserFunctionEmbedding)?.also { return it }
```
jesyspa, line 192: I want to remark that I'm not super happy with this, and I hope the stuff we discussed with regards to a typeresolver-like construct for the body would make a lot of sense for me.  I think then this can become a `getOrPut` call.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -192,7 +223,7 @@ class ProgramConverter(
         val lookupName = symbol.embedName(this)
         return when (val existing = methods[lookupName]) {
             null -> {
-                val signature = embedFullSignature(symbol)
+                val (_, signature) = embedFullSignature(symbol)
```
jesyspa: Given there's an identical bit below: maybe there should be an `embedCallable(symbol)` overload?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
jesyspa: Ahh, I see, makes sense

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/MethodBuilder.kt
```diff
@@ -80,7 +80,7 @@ class MethodBuilder private constructor() {
 
     fun returns(action: () -> Type): Exp.Result {
         val retType = action()
-        val variable = Var(PlaceholderReturnVariableName, retType)
+        val variable = Var(ReturnVariableName(0), retType)
```
jesyspa, line 83: Hmm, why will this always be the correct one?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
```diff
@@ -11,32 +11,113 @@ import org.jetbrains.kotlin.fir.symbols.impl.FirPropertySymbol
...
+}
+
+
+class ConstructorSignature(
+    val signature : NamedFunctionSignature,
+    override val symbol: FirFunctionSymbol<*>,
+    propertiesPostconditions : List<ExpEmbedding>,
+    typeResolver: TypeResolver
+) : FullNamedFunctionSignature, NamedFunctionSignature by signature {
+   override val preconditions = signature.basicPreconditions(typeResolver)
```
jesyspa, line 98: Please run a formatter over the code

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
The-Ray-Man: I don't quite follow. I don't see how the overload would help us here. Because the symbol is the same for both cases. It depends on what the type of the already (maybe partially) embedded type is.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/MethodBuilder.kt
The-Ray-Man: I think so yes. Because at the moment we do not allow to build methods with bodies (which I think is not something that we want to support since viper ignores the method body anyways).  If a user wants to refer to the return variable they can use `result` on the builder.

I think there is a problem here: The method builder operates on the viper level. And if we want to ensure for example a unique predicate, we need a ExpEmbedding return variable. So here the user needs to manually construct the correct return variable.

Generally I think it would make sense to move the method builder to the ExpEmbedding stage. Even though the function builder is on the viper level, which seems to be fine

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
jesyspa: Just to avoid the `val (_, sig) = ...; val callable = ...` duplication; not using a return value is a bit of a code smell, so by making it clearly the intended behaviour in a particular case the code becomes easier to read.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/domains/MethodBuilder.kt
jesyspa: I'm not sure.  I think the whole point of the MethodBuilder is to build precisely a Viper method; it is strange to build it at a higher level and then lower it.  Rather, I would add these kinds of things to the builder as functions and then systematically use the builder to build *all* methods.  But that may be optimistic.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
The-Ray-Man: Addressed in #149

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -387,90 +421,78 @@ class ProgramConverter(
          * from other modules.
          */
         val firSpec = when {
-            declaration !is FirSimpleFunction -> null
-            body == null -> null
+            declaration !is FirSimpleFunction -> return Pair(emptyList(), emptyList())
+            body == null -> return Pair(emptyList(), emptyList())
             else -> extractFirSpecification(body, declaration.symbol.resolvedReturnType)
         }
```
jesyspa: This looks a bit strange.  Maybe just have an if-return here?

jesyspa (approved):

PR #144, opened 2026-05-09
Add detekt linter

komiputer:
## Summary

Adds [detekt](https://detekt.dev/) for static analysis of the Kotlin sources, wires it into the Gradle build and CI, and clears the findings that came up.

**Build wiring** — applied at the root and to every subproject. Custom source layout (`src/`, `test-fixtures/`, `test-gen/`) is configured explicitly because the project doesn't use `src/main/kotlin`.

**Config** (`config/detekt/detekt.yml`) — `buildUponDefaultConfig`, with these tweaks for this codebase:
- `WildcardImport` disabled — compiler-plugin code routinely uses `org.jetbrains.kotlin.fir.*` and similar wildcards deliberately.
- `MatchingDeclarationName` disabled — Kotlin files here often group related declarations under a category name (e.g. `SpecialAccessors.kt`, `Invariant.kt`).
- `CyclomaticComplexMethod` threshold raised to 35 — AST-visiting `when` expressions are inherently wide.
- `ReturnCount.max` raised to 8 (test fixtures use guard-clause style).
- `MaxLineLength` and a few stylistic / documentation rules turned off — they have hundreds of findings each in this codebase and the value is mostly cosmetic.

**Findings fixed** rather than baselined:
- `SimpleNameResolver` and `debugMangled` moved from `org.jetbrains.kotlin.formver.names` to `.formver.core.names`. Only one file was in the dotless package — looks like a leftover from an earlier reorg, and detekt's `InvalidPackageDeclaration` flagged it.
- `throw IllegalStateException(...)` in exhaustive-when else branches replaced with `error(...)`.
- `throw IllegalArgumentException(...)` in test config replaced with `require { ... }`.
- Empty `()` constructors removed (`SsaStartNode`, `AbstractPhasedDiagnosticTest`).
- Unused private properties dropped (including a typo'd `unsued` in `VerificationFacade`).
- Explicit `it` lambda parameter dropped in `TagCollector`.
- One outlier `NameGraphRenderer.render` (complexity 54) suppressed locally — refactoring it is out of scope for this PR.
- Dead `platform` parameter removed from `PluginAnnotationsProvider.findLib`/`findLibFromProperty` (only "jvm" was ever passed).
- Nine pure-data classes converted to `data class`: `CatchBlockData`, `CatchBlockListData`, `PropertyEmbedding`, `UserFieldEmbedding`, `PlaceholderVariableEmbedding`, `FirVariableEmbedding`, `LinearizationVariableEmbedding`, `DiagnosticData`, `UserFunction`.
- `MagicNumber` suppressed locally on `ShortNameResolver.costOfName` (the literals are arbitrary cost weights — only relative ordering matters; clarified the doc comment).

## Test plan

- [x] `./gradlew detekt` passes locally
- [x] `./gradlew :formver.compiler-plugin:test` passes locally (with Z3)
- [ ] CI confirms

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
@@ -143,7 +143,9 @@ class ShortNameResolver : NameResolver {
     /**
      * The cost of having [name] in a viper name.
      * Higher cost means that [name] is more likely to not be used in the final viper name.
+     * Only the relative ordering of these values matters; the absolute numbers are arbitrary.
      */
+    @Suppress("detekt:MagicNumber")
     private fun costOfName(name: AnyName): Int = when (name) {
         is NameScope -> when (name) {
             BadScope -> 100
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
@@ -143,7 +143,11 @@ class ShortNameResolver : NameResolver {
     /**
      * The cost of having [name] in a viper name.
      * Higher cost means that [name] is more likely to not be used in the final viper name.
-     * Only the relative ordering of these values matters; the absolute numbers are arbitrary.
+     *
+     * A candidate name's total cost is the sum of its parts' costs, so the absolute
+     * gaps between the per-scope values shape which combinations are preferred.
+     * PackageScope is intentionally far above the rest so a single package part
+     * outweighs several cheaper ones.
      */
     @Suppress("detekt:MagicNumber")
     private fun costOfName(name: AnyName): Int = when (name) {
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -43,6 +43,8 @@ jobs:
 
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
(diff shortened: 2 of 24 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -143,7 +143,9 @@ class ShortNameResolver : NameResolver {
     /**
      * The cost of having [name] in a viper name.
      * Higher cost means that [name] is more likely to not be used in the final viper name.
+     * Only the relative ordering of these values matters; the absolute numbers are arbitrary.
```
The-Ray-Man: The added comment is technically not true since the total cost of a name is a sum of multiple parts. So the absolute difference can matter.

The-Ray-Man (commented):
I don't have a lot of experience with linters but it seems reasonable to me.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
jesyspa: Ah, that's good to know.  Would a broader range of numbers make the results more predictable, then?

The-Ray-Man: Probably yes. But tbh, I did not try out many different combinations. I think the most important thing for the numbers is that the packageScoep is much higher than the others (apart from the scopes that we never want to see)

The-Ray-Man (approved):

PR #145, opened 2026-05-11
Reduce SimpleNameResolver to rendering the last candidate

komiputer:
## Summary
- `SimpleNameResolver` reimplemented all of name decomposition (scope walking, `FreshName` cases, `KotlinName` cases, `NameTypeAction` context, etc.) just to produce a single string. Since every `AnyName` exposes a `candidates` list ordered shortest-to-fullest, rendering the last one already gives a fully-disambiguated name. That is all `SimpleNameResolver` was reconstructing by hand.
- Split the existing helper in `ShortNameResolver.kt`:
  - `fullName()` is renamed to `debugId()`. Behaviour unchanged: wraps multi-part dependent candidates in `[…]`, which is not a valid Viper identifier. Used by `NameGraphRenderer` and by `ShortNameResolver`'s unresolved-name fallback.
  - `longName()` is new. Same recursive descent without the bracketing, so the result is a valid Viper identifier.
- `SimpleNameResolver.lookup` and `SymbolicName.debugMangled` now call `longName()` directly. The 200-line companion object goes away.

## Test plan
- [x] `./gradlew :formver.compiler-plugin:test` passes
- [x] `./gradlew :formver.compiler-plugin:update` produces no fixture deltas, which confirms that `SimpleNameResolver` only feeds debug paths (`ToViperBuiltinOnlyError`, `PureExpLinearizer.catchNameResolver`, `debugMangled`) and that `ShortNameResolver` continues to own the real output.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
@@ -26,32 +26,44 @@ fun CandidateName.name(): String = parts.joinToString("") { it.name() }
 
 
 /**
- * Gives the most specific name.
+ * The fully-disambiguated form of a name, rendered as a valid Viper
+ * identifier. Distinct from [debugId], which adds readability decorations
+ * that aren't valid in Viper identifiers.
+ *
+ * Use this when a resolver needs a guaranteed-unique name without doing
+ * collision resolution (see [SimpleNameResolver]).
  */
-fun NamePart.fullName(): String = when (this) {
-    is NamePart.Dependent -> fullName()
-    is NamePart.Basic -> this.name
+private fun NamePart.longName(): String = when (this) {
+    is NamePart.Dependent -> name.longName()
+    is NamePart.Basic -> name
     NamePart.Separator -> SEPARATOR
 }
 
-/**
- * Gives the most specific name.
- */
-fun NamePart.Dependent.fullName(): String = name.candidates().last().fullName()
+private fun CandidateName.longName(): String = parts.joinToString("") { it.longName() }
+
+fun AnyName.longName(): String = candidates().last().longName()
+
 
 /**
- * Gives the most specific name.
+ * Debug rendering of a name. Same structure as [longName] but wraps
+ * multi-part dependent candidates in `[…]` for readability.
+ *
+ * The result is not a valid Viper identifier and is intended for
+ * debug/dump output only.
  */
-fun CandidateName.fullName(): String = if (parts.size > 1) {
-    "[" + parts.joinToString("") { it.fullName() } + "]"
-} else {
-    parts.joinToString("") { it.fullName() }
+private fun NamePart.debugId(): String = when (this) {
+    is NamePart.Dependent -> name.debugId()
+    is NamePart.Basic -> name
+    NamePart.Separator -> SEPARATOR
 }
 
-/**
- * Gives the most specific name.
- */
-fun AnyName.fullName(): String = candidates().last().fullName()
+private fun CandidateName.debugId(): String = if (parts.size > 1) {
+    "[" + parts.joinToString("") { it.debugId() } + "]"
+} else {
+    parts.joinToString("") { it.debugId() }
+}
+
+fun AnyName.debugId(): String = candidates().last().debugId()
 
 
 // END UTILITY SECTION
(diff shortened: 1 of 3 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -100,9 +108,11 @@ class ShortNameResolver : NameResolver {
 
     // START RESOLVE NAMES
     override fun lookup(name: SymbolicName): String = mangledNames.getOrElse(name) {
-        // The name.fullName() contains characters which are not allowed in viper. This should only happen when we dump the ExpEmbedding, which can contain
```
jesyspa, line 103: You misunderstood the comment; please keep the old one.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
```diff
@@ -26,32 +26,40 @@ fun CandidateName.name(): String = parts.joinToString("") { it.name() }
...
+    is NamePart.Dependent -> name.longName()
+    is NamePart.Basic -> name
     NamePart.Separator -> SEPARATOR
 }
 
-/**
- * Gives the most specific name.
- */
-fun NamePart.Dependent.fullName(): String = name.candidates().last().fullName()
+fun CandidateName.longName(): String = parts.joinToString("") { it.longName() }
```
jesyspa: Why is this needed?  Please explain.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/debug/NameGraphRenderer.kt
```diff
@@ -138,7 +138,7 @@ class NameGraphRenderer(val shortNameResolver: ShortNameResolver) {
             is NameType -> "Type: "
             is SymbolicName -> {
                 if (entity.nameType != null) {
-                    entity.nameType!!.fullName() + ": "
+                    entity.nameType!!.debugId() + ": "
```
jesyspa: The !! is a bit ugly here, could you use a local?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
```diff
@@ -1,226 +1,26 @@
...
-import org.jetbrains.kotlin.formver.core.names.*
-import org.jetbrains.kotlin.formver.viper.*
+import org.jetbrains.kotlin.formver.core.names.longName
+import org.jetbrains.kotlin.formver.viper.AnyName
+import org.jetbrains.kotlin.formver.viper.NameResolver
+import org.jetbrains.kotlin.formver.viper.SymbolicName
 
 /**
- * Resolves mangled names into Viper identifiers while maintaining uniqueness.
+ * Resolves a name by rendering its longest (last) candidate via [longName].
```
jesyspa: Don't mention implementation details in comments.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
komiputer: Restored the original wording, just with `fullName()` → `debugId()`.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/ShortNameResolver.kt
komiputer: Expanded the KDoc to explain why it's distinct from `debugId`: `debugId` adds readability decorations (the `[…]` wrapping) that aren't valid in Viper identifiers, so resolvers that want a guaranteed-unique-but-valid name need a separate entry point. `SimpleNameResolver` is the immediate user.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/debug/NameGraphRenderer.kt
komiputer: Done — switched to a local.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/SimpleNameResolver.kt
komiputer: Reworded to describe what the resolver does, without mentioning `longName`.

The-Ray-Man (approved):
Looks good. 
I checked some generated names and they are as intendent.

jesyspa (approved):

PR #147, opened 2026-05-11
Clean up the names package

komiputer:
## Summary
A handful of small dedupe/dead-code passes in the names package. None of them
change Viper output (no test fixtures regenerate):

- Delete the `AnyName.candidates()` wrapper, which only returned the
  `candidates` property. Switch the five call sites in `ShortNameResolver`
  and `NameGraphRenderer` to the property directly.
- Drop `nameWithDependentPrefixCandidates(String, AnyName)` — its body
  was byte-identical to `nameWithPrefixCandidates(String, AnyName)`. Move
  its two call sites (`PrivateScope`, `NamedDomainAxiomLabel`) over and
  rename the surviving helper's second parameter from `nameType` to
  `prefix`, since it now spans both `NameType` and `NameScope` prefixes.
- Make `ScopedNameBuilder` extend `ScopeBuilder` instead of wrapping it.
  Removes 8 forwarder methods.
- Replace the `when (this)` block in `LabelName` with a `labelString`
  constructor parameter. The five subclasses stay so that
  `SimpleFreshEntityProducer<CatchLabelName>` and the `::CatchLabelName`
  constructor references continue to typecheck.

Net: −44 lines.

## Test plan
- [x] `./gradlew :formver.compiler-plugin:core:compileKotlin` — green
- [x] `./gradlew :formver.compiler-plugin:test` — green (no fixture diffs)

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/DomainName.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/names/DomainName.kt
@@ -72,7 +72,7 @@ data class FromRefFuncName(val baseName: SymbolicName) : SymbolicName {
 data class NamedDomainAxiomLabel(val domainName: SymbolicName, val baseName: String) : SymbolicName {
     override val inViper: Boolean = true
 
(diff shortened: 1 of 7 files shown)


The-Ray-Man (approved):
looks good!

Issue #148, opened 2026-05-11
Refactor conversion process into a pipeline


komiputer:
## Motivation

The current conversion process braids four logical stages together inside `ProgramConverter.registerForVerification` and `StmtConversionContext.convertMethodWithBody` / `convertFunctionWithBody`. In particular, `convertMethodWithBody` linearizes the body **before** running `checkValidity`, then has to return an `InvalidFunctionBodyEmbedding` to model "we already linearized, but pretend we didn't." Mutable `body` fields on `UserFunctionEmbedding` / `PureUserFunctionEmbedding` carry state across stages, and bail-out on errors is scattered across `ViperPoweredDeclarationChecker.check` (purity bail, ADT bail) and the body-conversion functions (validity bail).

The four stages are:

1. **Register** functions for verification (embed signatures).
2. **Convert** function bodies to `ExpEmbedding`.
3. **Check** errors on the `ExpEmbedding`s.
4. **Linearize** valid `ExpEmbedding`s into Viper.

This issue proposes untangling them.

## Target shape

The checker drives four phases sequentially with a single bail point:

```kotlin
pc.register(declaration)                       // signatures into pc.methods / pc.functions
val converted = pc.convertBody(declaration)    // returns ConvertedBodyResolver
pc.check(converted)                            // walks ExpEmbeddings, writes errorCollector
reportErrors(); if (errorCollector.anyErrors()) return
val linearized = pc.linearize(converted)       // returns LinearizedBodyResolver
val program = pc.program(linearized)           // reads linearized for body-vs-header decisions
```

Key principles:

- Errors are only inspected once, after `check`. Conversion is allowed to produce invalid `ExpEmbedding`s — they get caught in `check`.
- Bodies live in dedicated resolver structures, not on the embedding classes. `UserFunctionEmbedding` and `PureUserFunctionEmbedding` become immutable.
- `InvalidFunctionBodyEmbedding` disappears: invalid bodies simply have no entry in `LinearizedBodyResolver`.

## Resolvers

```kotlin
class ConvertedBodyResolver {
    private val impure: MutableMap<SymbolicName, ImpureConvertedBody> = mutableMapOf()
    private val pure:   MutableMap<SymbolicName, ExpEmbedding> = mutableMapOf()
    // store / lookup / iterate-all
}

class LinearizedBodyResolver {
    private val impure: MutableMap<SymbolicName, FunctionBodyEmbedding> = mutableMapOf()
[...]

PR #149, opened 2026-05-11
Split function-body conversion into convert and linearize phases

komiputer:
## Summary

Refactor steps 1-3 from #148: pull function-body state off the embedding classes and split the conversion path into discrete convert/linearize halves.

- New `ConvertedBodyResolver` and `LinearizedBodyResolver` hold per-function `ExpEmbedding` and Viper outputs, keyed by `SymbolicName`. Each resolver exposes lookup, store, and (for the convert resolver) a `forEach` traversal — no live `Map` is handed out.
- `UserFunctionEmbedding.body` and `PureUserFunctionEmbedding.body` are gone; both classes are now immutable. `viperMethod` / `viperFunction` take the linearized body as a parameter and `ProgramConverter.program` looks it up.
- `convertMethodWithBody` is split into `convertImpureBody` (writes the `FunctionExp` wrap to `ConvertedBodyResolver`) and `linearizeImpureBody` (reads it back, linearizes, runs `checkValidity`, and writes `FunctionBodyEmbedding` to `LinearizedBodyResolver` if valid). The unit-invariant tail still lives at linearize time per planning. Pure counterpart split the same way.
- `debugExpEmbeddings` now sources from the convert resolver. `FunctionBodyEmbedding.debugExpEmbedding` (and the corresponding interface method on `FunctionBodyConversionResult`) was dead-weighted by the move, so dropped.
- A defensive `check(...)` in `ProgramConverter.program` asserts the invariant "if a body was converted, it was also linearized" so a future bail-condition reshuffle can't silently emit header-only Viper methods.

Out of scope (slated for the remaining steps in #148): a dedicated `check` phase between convert and linearize, and the removal of `InvalidFunctionBodyEmbedding` / `FunctionBodyConversionResult` / tightening of the now-unused return types on the convert functions.

## Test plan

- [x] `./gradlew :formver.compiler-plugin:assemble`
- [x] `./gradlew :formver.compiler-plugin:test`

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
@@ -12,8 +12,12 @@ import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Exp
 
 /**
- * The output of converting an impure (method-style) function body: the `FunctionExp` wrap built
- * during conversion, together with the `ReturnTarget` it was converted against.
+ * The output of converting an impure (method-style) function body.
+ *
+ * The [returnTarget] is preserved alongside [bodyExp] because linearization needs to attach a
+ * Unit invariant to the return variable when the function returns Unit and the body has no
+ * explicit `return`. The body alone does not expose that variable, so conversion has to hand
+ * it off explicitly.
  */
 data class ConvertedMethodBody(
     val bodyExp: FunctionExp,
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
@@ -5,11 +5,11 @@
 
 package org.jetbrains.kotlin.formver.core.conversion
 
-import org.jetbrains.kotlin.formver.core.embeddings.FunctionBodyEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.expression.FunctionExp
 import org.jetbrains.kotlin.formver.viper.SymbolicName
-import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.Function
+import org.jetbrains.kotlin.formver.viper.ast.Method
 
 /**
  * The output of converting an impure (method-style) function body.
@@ -55,22 +55,24 @@ class ConvertedBodyResolver {
 }
 
 /**
- * Stores per-function Viper outputs from the linearization phase, keyed by symbolic name.
+ * Stores the finalized Viper [Method] / [Function] for each user-defined or partially-special callable,
+ * keyed by symbolic name. Populated by `linearizeAll` so that `buildProgram` reduces to concatenation.
  *
- * Each entry is written at most once; storing under a name that already has an entry is an error.
+ * Iteration order over [methods] / [functions] reflects insertion order, which is the order the
+ * `ProgramConverter` emits.
  */
 class LinearizedBodyResolver {
-    private val impure: MutableMap<SymbolicName, FunctionBodyEmbedding> = mutableMapOf()
-    private val pure: MutableMap<SymbolicName, Exp> = mutableMapOf()
+    private val methodsByName: MutableMap<SymbolicName, Method> = mutableMapOf()
+    private val functionsByName: MutableMap<SymbolicName, Function> = mutableMapOf()
 
-    fun storeImpure(name: SymbolicName, body: FunctionBodyEmbedding) {
-        check(impure.put(name, body) == null) { "Linearized method body for $name was already stored" }
+    fun storeMethod(name: SymbolicName, method: Method) {
+        check(methodsByName.put(name, method) == null) { "Linearized method for $name was already stored" }
     }
 
-    fun storePure(name: SymbolicName, body: Exp) {
-        check(pure.put(name, body) == null) { "Linearized function body for $name was already stored" }
+    fun storeFunction(name: SymbolicName, function: Function) {
+        check(functionsByName.put(name, function) == null) { "Linearized function for $name was already stored" }
     }
 
-    fun lookupImpure(name: SymbolicName): FunctionBodyEmbedding? = impure[name]
-    fun lookupPure(name: SymbolicName): Exp? = pure[name]
+    val methods: Collection<Method> get() = methodsByName.values
+    val functions: Collection<Function> get() = functionsByName.values
 }
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2010-2025 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.formver.core.conversion
+
+import org.jetbrains.kotlin.formver.core.embeddings.FunctionBodyEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+
+/**
+ * The output of converting an impure function body: the fully-built `ExpEmbedding` (typically a
+ * `FunctionExp` wrapping the body) together with the `ReturnTarget` it was converted against.
+ */
+data class ImpureConvertedBody(
+    val bodyExp: ExpEmbedding,
+    val returnTarget: ReturnTarget,
+)
+
+/**
+ * Stores per-function `ExpEmbedding` outputs from the conversion phase, keyed by symbolic name.
+ * Pure and impure functions live in separate maps so the value type can stay precise.
+ */
+class ConvertedBodyResolver {
+    private val impure: MutableMap<SymbolicName, ImpureConvertedBody> = mutableMapOf()
+    private val pure: MutableMap<SymbolicName, ExpEmbedding> = mutableMapOf()
+
+    fun storeImpure(name: SymbolicName, body: ImpureConvertedBody) {
+        impure[name] = body
+    }
+
+    fun storePure(name: SymbolicName, body: ExpEmbedding) {
+        pure[name] = body
+    }
+
+    fun lookupImpure(name: SymbolicName): ImpureConvertedBody? = impure[name]
+    fun lookupPure(name: SymbolicName): ExpEmbedding? = pure[name]
+
+    fun forEachImpure(action: (SymbolicName, ImpureConvertedBody) -> Unit) {
+        impure.forEach { (name, body) -> action(name, body) }
+    }
+
+    fun forEachPure(action: (SymbolicName, ExpEmbedding) -> Unit) {
+        pure.forEach { (name, body) -> action(name, body) }
+    }
+}
+
+/**
+ * Stores per-function Viper outputs from the linearization phase, keyed by symbolic name.
+ * Bodies that fail validity checks have no entry here.
+ */
+class LinearizedBodyResolver {
+    private val impure: MutableMap<SymbolicName, FunctionBodyEmbedding> = mutableMapOf()
+    private val pure: MutableMap<SymbolicName, Exp> = mutableMapOf()
+
+    fun storeImpure(name: SymbolicName, body: FunctionBodyEmbedding) {
+        impure[name] = body
+    }
+
+    fun storePure(name: SymbolicName, body: Exp) {
+        pure[name] = body
+    }
+
+    fun lookupImpure(name: SymbolicName): FunctionBodyEmbedding? = impure[name]
+    fun lookupPure(name: SymbolicName): Exp? = pure[name]
+}
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
@@ -7,38 +7,41 @@ package org.jetbrains.kotlin.formver.core.conversion
 
 import org.jetbrains.kotlin.formver.core.embeddings.FunctionBodyEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.FunctionExp
 import org.jetbrains.kotlin.formver.viper.SymbolicName
 import org.jetbrains.kotlin.formver.viper.ast.Exp
 
 /**
- * The output of converting an impure function body: the fully-built `ExpEmbedding` (typically a
- * `FunctionExp` wrapping the body) together with the `ReturnTarget` it was converted against.
+ * The output of converting an impure (method-style) function body: the `FunctionExp` wrap built
+ * during conversion, together with the `ReturnTarget` it was converted against.
  */
-data class ImpureConvertedBody(
-    val bodyExp: ExpEmbedding,
+data class ConvertedMethodBody(
+    val bodyExp: FunctionExp,
     val returnTarget: ReturnTarget,
 )
 
 /**
  * Stores per-function `ExpEmbedding` outputs from the conversion phase, keyed by symbolic name.
  * Pure and impure functions live in separate maps so the value type can stay precise.
+ *
+ * Each entry is written at most once; storing under a name that already has an entry is an error.
  */
 class ConvertedBodyResolver {
-    private val impure: MutableMap<SymbolicName, ImpureConvertedBody> = mutableMapOf()
+    private val impure: MutableMap<SymbolicName, ConvertedMethodBody> = mutableMapOf()
     private val pure: MutableMap<SymbolicName, ExpEmbedding> = mutableMapOf()
 
-    fun storeImpure(name: SymbolicName, body: ImpureConvertedBody) {
-        impure[name] = body
+    fun storeImpure(name: SymbolicName, body: ConvertedMethodBody) {
+        check(impure.put(name, body) == null) { "Converted method body for $name was already stored" }
     }
 
     fun storePure(name: SymbolicName, body: ExpEmbedding) {
-        pure[name] = body
+        check(pure.put(name, body) == null) { "Converted function body for $name was already stored" }
     }
 
-    fun lookupImpure(name: SymbolicName): ImpureConvertedBody? = impure[name]
+    fun lookupImpure(name: SymbolicName): ConvertedMethodBody? = impure[name]
     fun lookupPure(name: SymbolicName): ExpEmbedding? = pure[name]
 
-    fun forEachImpure(action: (SymbolicName, ImpureConvertedBody) -> Unit) {
+    fun forEachImpure(action: (SymbolicName, ConvertedMethodBody) -> Unit) {
         impure.forEach { (name, body) -> action(name, body) }
     }
 
@@ -49,18 +52,19 @@ class ConvertedBodyResolver {
 
 /**
  * Stores per-function Viper outputs from the linearization phase, keyed by symbolic name.
- * Bodies that fail validity checks have no entry here.
+ *
+ * Each entry is written at most once; storing under a name that already has an entry is an error.
  */
 class LinearizedBodyResolver {
     private val impure: MutableMap<SymbolicName, FunctionBodyEmbedding> = mutableMapOf()
     private val pure: MutableMap<SymbolicName, Exp> = mutableMapOf()
 
     fun storeImpure(name: SymbolicName, body: FunctionBodyEmbedding) {
-        impure[name] = body
+        check(impure.put(name, body) == null) { "Linearized method body for $name was already stored" }
     }
 
     fun storePure(name: SymbolicName, body: Exp) {
-        pure[name] = body
+        check(pure.put(name, body) == null) { "Linearized function body for $name was already stored" }
     }
 
     fun lookupImpure(name: SymbolicName): FunctionBodyEmbedding? = impure[name]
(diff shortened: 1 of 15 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.jetbrains.kotlin.formver.core.embeddings.FunctionBodyEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+
+/**
+ * The output of converting an impure function body: the fully-built `ExpEmbedding` (typically a
+ * `FunctionExp` wrapping the body) together with the `ReturnTarget` it was converted against.
+ */
+data class ImpureConvertedBody(
```
jesyspa: I think `ConvertedMethodBody` makes more sense.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
```diff
@@ -0,0 +1,68 @@
...
+package org.jetbrains.kotlin.formver.core.conversion
+
+import org.jetbrains.kotlin.formver.core.embeddings.FunctionBodyEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+
+/**
+ * The output of converting an impure function body: the fully-built `ExpEmbedding` (typically a
+ * `FunctionExp` wrapping the body) together with the `ReturnTarget` it was converted against.
```
jesyspa: Why only typically?  Can't we make this required?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Stores per-function `ExpEmbedding` outputs from the conversion phase, keyed by symbolic name.
+ * Pure and impure functions live in separate maps so the value type can stay precise.
+ */
+class ConvertedBodyResolver {
+    private val impure: MutableMap<SymbolicName, ImpureConvertedBody> = mutableMapOf()
+    private val pure: MutableMap<SymbolicName, ExpEmbedding> = mutableMapOf()
+
+    fun storeImpure(name: SymbolicName, body: ImpureConvertedBody) {
+        impure[name] = body
```
jesyspa: Here and elsewhere: may be worth throwing if it's present?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -77,6 +75,8 @@ class ProgramConverter(
     override val anonBuiltinVarProducer = FreshEntityProducer(::AnonymousBuiltinVariableEmbedding)
     override val returnTargetProducer = FreshEntityProducer(::ReturnTarget)
     override val nameResolver = ShortNameResolver()
+    override val convertedBodyResolver = ConvertedBodyResolver()
+    override val linearizedBodyResolver = LinearizedBodyResolver()
 
 
     val program: Program
```
jesyspa: Maybe this should be a function, not a property?  Given how much work it can require.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -103,15 +114,15 @@ class ProgramConverter(
...
         if (declaration.symbol.isPure(session)) {
             embedPureUserFunction(declaration.symbol, signature)
         } else {
             val (returnTarget, stmtCtx) = createBodyConversionContext(signature)
-            embedUserFunction(declaration.symbol, signature).apply {
-                body = stmtCtx.convertMethodWithBody(declaration, signature, returnTarget)
-            }
+            embedUserFunction(declaration.symbol, signature)
+            stmtCtx.convertImpureBody(declaration, signature, returnTarget)
+            stmtCtx.linearizeImpureBody(declaration, signature.name)
```
jesyspa: I think this is the wrong place to do this: to bail out on time, we need to call convert and linearize from outside.

I think a better approach is:
- register function adds to a list
- convertAll populates conversion resolver
- linearizeAll populates linearized resolver
- getProgram gets back the program.

That said: I would try to factor ProgramConverter out as much as possible.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -241,41 +239,58 @@ fun StmtConversionContext.insertForAllFunctionCall(
...
     declaration: FirSimpleFunction,
     signature: FullNamedFunctionSignature,
     returnTarget: ReturnTarget,
-): FunctionBodyConversionResult? {
-    val firBody = declaration.body ?: return null
+) {
+    val firBody = declaration.body ?: return
     val body = convert(firBody)
     val bodyExp = FunctionExp(signature, body, returnTarget.label)
+    convertedBodyResolver.storeImpure(signature.name, ImpureConvertedBody(bodyExp, returnTarget))
```
jesyspa: This feels like the wrong place to do this: rather, the caller should be responsible for storing it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -241,41 +239,58 @@ fun StmtConversionContext.insertForAllFunctionCall(
...
-    returnTarget.variable.withIsUnitInvariantIfUnit(typeResolver).toLinearizable(declaration.source)
-        .toViperUnusedResult(linearizer)
-    val isValid = body.checkValidity(declaration.source, errorCollector)
-    return if (isValid) {
-        FunctionBodyEmbedding(seqnBuilder.block, returnTarget, bodyExp)
-    } else {
-        InvalidFunctionBodyEmbedding(declaration.source, bodyExp)
+    converted.returnTarget.variable.withIsUnitInvariantIfUnit(typeResolver)
+        .toLinearizable(declaration.source).toViperUnusedResult(linearizer)
+    val isValid = converted.bodyExp.checkValidity(declaration.source, errorCollector)
```
jesyspa: Validity checking should be after conversion.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -241,41 +239,58 @@ fun StmtConversionContext.insertForAllFunctionCall(
...
+fun StmtConversionContext.convertPureBody(
+    declaration: FirSimpleFunction,
+    name: SymbolicName,
+) {
     val firBody = declaration.body ?: throw SnaktInternalException(
         declaration.source,
         "Pure functions expect a function body to exist"
     )
-    val body = convert(firBody)
+    convertedBodyResolver.storePure(name, convert(firBody))
```
jesyspa: Same as above (also in other functions here)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionEmbedding.kt
```diff
@@ -5,52 +5,38 @@
...
-}
-
-interface PureFunctionEmbedding : CallableEmbedding {
-    fun viperFunction(ctx: TypeResolver): Function?
-}
+/**
+ * Marker for embeddings that live in the `methods` bucket of a `ProgramConverter`.
+ * The hierarchy is sealed so renderers can dispatch over the leaves exhaustively.
+ */
+sealed interface FunctionEmbedding : CallableEmbedding
```
jesyspa: I'm not sure there's added value here, over just having a CallableEmbedding.

jesyspa (changes requested):

komiputer:
Addressed the review in b6643753:

- **`ConvertedMethodBody`**: renamed from `ImpureConvertedBody`, and the `bodyExp` field is now typed `FunctionExp` (was `ExpEmbedding`) so the "typically" wording in the doc is gone — it's required.
- **Throw on duplicate store**: both resolvers' `store*` methods now `check(map.put(name, body) == null)` and fail loudly if the same name is stored twice.
- **`program()` as a function**: was a `val program` getter, now `fun program(): Program`. Checker call site updated.
- **Pipeline split**: `registerForVerification` is gone, replaced with `register` / `convertAll` / `validateAll` / `linearizeAll` on `ProgramConverter`. The checker now drives them explicitly:
  ```
  register(declaration)
  convertAll()
  validateAll()        // walks converted bodies, surfaces validity + purity errors
  reportErrors(); if errors return
  linearizeAll()
  program()
  ```
  Pure callees reached transitively during `convertAll` (via `embedPureFunction`) still get their bodies converted on the fly into the same resolver; `linearizeAll` iterates the resolver, not the registered list, so callees get linearized too.
- **Caller stores, not the convert/linearize helpers**: `convertImpureBody` / `convertPureBody` now return their result, and the orchestration in `convertAll` is responsible for the `storeImpure` / `storePure`. Same for `linearize*` (which also moved off `StmtConversionContext` onto `ProgramConversionContext` since they no longer need any method-context state).
- **Validity check after conversion**: removed from `linearize*`; it lives in `validateAll`, which runs between convert and linearize.
- **No more `FunctionEmbedding` marker**: dropped the interface entirely. `methods` is typed `MutableMap<SymbolicName, CallableEmbedding>`. The rendering dispatch is an explicit `when` over `UserFunctionEmbedding | FullySpecialKotlinFunction | PartiallySpecialKotlinFunction` inside `program()`. `SpecialKotlinFunction` is sealed and extends `CallableEmbedding` directly. While there, narrowed `PartiallySpecialKotlinFunction.baseEmbedding` from `FunctionEmbedding?` to `UserFunctionEmbedding?` (which is what is ever actually stored).

Builds + full plugin test suite still green.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/BodyResolvers.kt
```diff
@@ -0,0 +1,72 @@
...
+
+import org.jetbrains.kotlin.formver.core.embeddings.FunctionBodyEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.ExpEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.expression.FunctionExp
+import org.jetbrains.kotlin.formver.viper.SymbolicName
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+
+/**
+ * The output of converting an impure (method-style) function body: the `FunctionExp` wrap built
+ * during conversion, together with the `ReturnTarget` it was converted against.
```
jesyspa: Don't comment on how a data type is built.  I think a more interesting point here is the reason that `returnTarget` needs to be there at all.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/SpecialKotlinFunction.kt
```diff
@@ -45,14 +41,12 @@ interface PartiallySpecialKotlinFunction : SpecialKotlinFunction {
...
     fun tryInsertCall(args: List<ExpEmbedding>, ctx: StmtConversionContext): ExpEmbedding?
     override fun insertCall(args: List<ExpEmbedding>, ctx: StmtConversionContext): ExpEmbedding {
         return tryInsertCall(args, ctx) ?: baseEmbedding?.insertCall(args, ctx)
         ?: error("Base embedding for partially special function $name not specified")
     }
 
-    fun initBaseEmbedding(embedding: FunctionEmbedding)
-
-    override fun viperMethod(ctx: TypeResolver): Method? = baseEmbedding?.viperMethod(ctx)
+    fun initBaseEmbedding(embedding: UserFunctionEmbedding)
```
jesyspa, line 51: Would it be possible to avoid this statefulness?  For example, having a resolver that constructs either a special Kotlin function or a normal Kotlin method or something.  If it's easy, please do so - if not, please reply to this comment with why not.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
     override val anonBuiltinVarProducer = FreshEntityProducer(::AnonymousBuiltinVariableEmbedding)
     override val returnTargetProducer = FreshEntityProducer(ReturnTarget::createForDepth)
     override val nameResolver = ShortNameResolver()
+    override val convertedBodyResolver = ConvertedBodyResolver()
+    override val linearizedBodyResolver = LinearizedBodyResolver()
+
+
+    fun program(): Program = Program(
```
jesyspa: `buildProgram`

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
...
+        domains = listOf(RuntimeTypeDomain(typeResolver)),
+        // We need to deduplicate fields since public fields with the same name are represented differently
+        // at `FieldEmbedding` level but map to the same Viper.
+        fields = typeResolver.backingFields().distinctBy { it.name }
+            .map { it.toViper() },
+        functions = SpecialFunctions.all + functions.entries.mapNotNull { (name, embedding) ->
+            val linearized = linearizedBodyResolver.lookupPure(name)
+            check(linearized != null || convertedBodyResolver.lookupPure(name) == null) {
```
jesyspa: We're a bit inconsistent with whether we throw SnaktInternalError or just use `check`, WDYT?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
...
+            val linearized = linearizedBodyResolver.lookupImpure(name)
+            check(linearized != null || convertedBodyResolver.lookupImpure(name) == null) {
+                "Internal error: impure function $name was converted but not linearized"
+            }
+            val callable = when (embedding) {
+                is UserFunctionEmbedding -> embedding.callable
+                is FullySpecialKotlinFunction -> null
+                is PartiallySpecialKotlinFunction -> embedding.baseEmbedding?.callable
+                else -> error("Unexpected embedding in methods map for $name: $embedding")
+            } ?: return@mapNotNull null
```
jesyspa: Maybe this should be in a helper?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
...
+    override val linearizedBodyResolver = LinearizedBodyResolver()
+
+
+    fun program(): Program = Program(
+        domains = listOf(RuntimeTypeDomain(typeResolver)),
+        // We need to deduplicate fields since public fields with the same name are represented differently
+        // at `FieldEmbedding` level but map to the same Viper.
+        fields = typeResolver.backingFields().distinctBy { it.name }
+            .map { it.toViper() },
+        functions = SpecialFunctions.all + functions.entries.mapNotNull { (name, embedding) ->
```
jesyspa: I feel like these long chains of + aren't great.  Maybe `buildList` would make this easier, or adding them earlier?  (it's fine if the `distinctBy` is the last step, there should be no collisions... honestly, if the `distinctBy` fires that's kinda suspicious)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
...
-                .distinctBy { it.name } + typeResolver.backingFields().map { it.type.havocMethod(typeResolver) }
-                .distinctBy { it.name },
-            predicates = typeResolver.classTypeEmbeddings().flatMap {
-                with(typeResolver) {
-                    listOf(
-                        it.sharedPredicate(), it.uniquePredicate()
-                    )
+    /**
+     * Convert each registered declaration's body to an `ExpEmbedding`, populating [convertedBodyResolver].
+     * Pure callees reached transitively are converted via [embedPureFunction]; their entries land in the
```
jesyspa: Note non-pure callees (e.g. inline ones) may also be converted this way, not sure it's worth calling this out.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
...
+     * same resolver.
+     */
+    fun convertAll() {
+        for ((declaration, signature, returnTarget) in registered) {
+            val stmtCtx = createBodyConversionContext(signature, returnTarget)
+            if (declaration.symbol.isPure(session)) {
```
jesyspa: After the latest rebase, doesn't the signature know about the purity?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -67,62 +70,142 @@ class ProgramConverter(
...
+            val source = functions[name]?.callable?.declarationSource
+            linearizedBodyResolver.storePure(name, linearizePureBody(source, body))
         }
     }
 
+    private fun ensurePureUserFunctionEmbedding(
+        symbol: FirFunctionSymbol<*>, signature: FullNamedFunctionSignature
+    ): PureUserFunctionEmbedding {
+        (functions[signature.name] as? PureUserFunctionEmbedding)?.also { return it }
+        val new = PureUserFunctionEmbedding(embedCallable(symbol, signature))
```
jesyspa: Consider using `also` or perhaps `getOrPut`

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -186,20 +269,18 @@ class ProgramConverter(
...
+        signature: FullNamedFunctionSignature,
+    ): UserFunctionEmbedding {
         val name = symbol.embedName(this)
         (methods[name] as? UserFunctionEmbedding)?.also { return it }
-        val (returnTarget, fullSignature) = embedFullSignature(symbol)
-        val new = UserFunctionEmbedding(embedCallable(symbol, fullSignature)).apply {
-            val stmtCtx = createBodyConversionContext(fullSignature, returnTarget)
-            body = stmtCtx.convertMethodWithBody(declaration, fullSignature)
-        }
+        val new = UserFunctionEmbedding(embedCallable(symbol, signature))
```
jesyspa: Consider `also` or `getOrPut`

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/SpecialKotlinFunction.kt
komiputer: Not easy with the current shape. The `methods` map is pre-populated by name with all `PartiallySpecialKotlinFunction`s before any FIR is seen, so we can short-circuit `embedFunction` for calls to (e.g.) `String.plus(Any)`. But the base `UserFunctionEmbedding` needs a `FirFunctionSymbol` we don't have at registration time, so it has to be filled in later via `initBaseEmbedding`.

Removing the statefulness would mean moving partial-special recognition into `embedFunction` itself (look up by name when the symbol arrives, then construct the partial-special with its base embedding up front). Doable but a separate refactor from this PR.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
komiputer: Looking at the rest of the file the convention seems to be: `SnaktInternalException` when there is a FIR `source`/symbol worth surfacing in diagnostics (e.g. the `as? FirSimpleFunction` failures), and plain `check`/`error`/`require` for invariant violations on internal resolver state with no relevant source to attach. The two checks here are about the converted/linearized resolver maps disagreeing, so `check` fits. Happy to switch if you'd rather standardize on one.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -68,60 +71,150 @@ class ProgramConverter(
...
+        symbol: FirFunctionSymbol<*>, signature: FullNamedFunctionSignature
+    ): PureUserFunctionEmbedding = functions.getOrPut(signature.name) {
+        PureUserFunctionEmbedding(embedCallable(symbol, signature))
+    }
+
     @OptIn(SymbolInternals::class)
     private fun embedPureUserFunction(
         symbol: FirFunctionSymbol<*>, signature: FullNamedFunctionSignature, returnTarget: ReturnTarget
     ): PureUserFunctionEmbedding {
         (functions[signature.name] as? PureUserFunctionEmbedding)?.also { return it }
```
The-Ray-Man: The type check is not necessary, since the map contains always ``PureUserFunctionEmbedding`` as values. A null check would be enough

The-Ray-Man (changes requested):
I think this generally looks good. A couple of things came to mind:

First, I noticed we treat pure and impure functions quite differently. Is there a specific reason for this? I feel like there should be a way to handle them uniformly.

Also, I think we're doing too much work for the methods in `buildProgram`. Ideally, we should have all the Viper methods and functions ready right after `linearizeAll`. The `LinearizedBodyResolver ` currently stores function bodies mapped to their symbolic names, but why don't we just store the complete method or function there directly? `linearizeAll ` could simply iterate over all methods/functions, check if they need a body, and then linearize the body and create the function/method as necessary.

jesyspa (approved):
@The-Ray-Man Could you do a final look-over please?  I think the split looks reasonable.

The-Ray-Man (approved):

PR #151, opened 2026-05-12
SSA Condition Ordering

The-Ray-Man:
This PR solves a minor issue with the order of aggregated conditions in the SSA conversion.

Example:
```kotlin
if (nextNode != null) {
      if (nextNode.value < 0) {
          false
      } else {
          nextNode.value > node.value && isStrictlyAscendingAndPositive(nextNode)
      }
}
```
This generated the viper code :

```viper
let anon_3_0 ==
        ((intFromRef(anon_2_0) < 0 && !(nextNode_0 == nullValue()) ?
          boolToRef(false) :
          null)
```
But the order of the `&&` should be the other way around. This was not really a problem yet, because we used leaky wildcard permissions. When I removed the shared predicates and used the unique instead viper complained about missing permissions. This was the case because we first read the field `anon_2_0` and only then checked if the receiver is not null.


jesyspa (approved):
Thanks for catching this!

PR #153, opened 2026-05-12
Completely move Conversion into Verification

The-Ray-Man:
Fixes #127 which was merged a bit too fast.

Also, I noticed that I missed to delete some diagnostic files in the previous PR


jesyspa (approved):
Really happy to see the unification donel

PR #155, opened 2026-05-12
Fields as Functions

The-Ray-Man:
This should be the first step for implementing fields-as-functions

The main changes contain:
- the `isDefaultBehavedProperty` function. With default behaviour we mean that we know how a property behaves. These properties are the once that can be represented as pure functions or backing fields
- The `MemberEmbeddingPolicy` was updated to reflect the new changes. For every well-behaved immutable property we need a unique function. Therefore for those properties we add the class scope. If a property is not well-behaved we represent it as a method. Since the method has no special postconditions at the moment we use one method for multiple properties, hence for this we still just use the public scope.
- All the shared predicates were removed.
- The `GetterEmbedding` got a second method to get the value. The new simple getter will return the getter without adding any type inhaling around it. It is used in the postcondition of the constructor to fix the value of well behaved properties. The properties that made problems here are the special fields, because for those we do not have type information around and therefore when accessing them the type information is inhaled which is not permitted in the pure context.

At the moment the getter method (the one we create for ill behaved properties) does not have any postconditions. In the future we might want to add a case distinction over subtypes to be able to reason about values read before and after a cast.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -49,6 +49,13 @@ interface ProgramConversionContext {
     fun embedFunctionPretype(symbol: FirFunctionSymbol<*>): FunctionTypeEmbedding
     fun embedType(exp: FirExpression): TypeEmbedding = embedType(exp.resolvedType)
     fun embedProperty(symbol: FirPropertySymbol): PropertyEmbedding
+
+    /**
+     * Returns true if the property is well-behaved. With this we mean that it is:
+     * Cannot be overwritten and does not have custom getters or setters
+     * If this is the case, then we know exactly how the property will behave.
```
jesyspa: I'm not a fan of "well behaved" as the name here, since it doesn't really capture what is going on.  Perhaps `isGuaranteedDefaultProperty` or something?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -327,27 +337,52 @@ class ProgramConverter(
...
                         SpecialProperties.lookup(symbol)?.let { return@getOrPutProperty it }
                     }
                 }
-                // Check if the symbol can be represented using a backing field
-                embedBackingField(symbol)?.let {
-                    return@getOrPutProperty PropertyEmbedding(
-                        BackingFieldGetter(it), BackingFieldSetter(it)
-                    )
+
+                val classSymbol = symbol.dispatchReceiverType?.toClassSymbol(session) as? FirRegularClassSymbol ?: throw SnaktInternalException(symbol.source, "Properties dispatch receiver is not a class")
```
jesyspa: This is rather long... maybe there's a neater way to phrase this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -327,27 +337,52 @@ class ProgramConverter(
...
+                val wellBehaved = isWellBehavedProperty(symbol)
+                val isImmutable = symbol.isVal
+                val isManual = symbol.isManual(session)
+
+                return@getOrPutProperty when {
+                    (wellBehaved && !isImmutable) || isManual -> {
```
jesyspa: Interesting question about manual, worth discussing...

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -96,6 +98,16 @@ class TypeResolver {
 
     fun lookupBackingField(name: ClassPropertyPair): FieldEmbedding? = properties[name]?.let { toBackingField(it) }
 
+    fun lookupFinalProperties(name: ClassPropertyPair) : PropertyEmbedding? {
+        val property = properties[name]
+        // TODO: Make this nicer
```
jesyspa: Agreed :)

jesyspa (changes requested):
Not completely thorough yet, but some thoughts.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
The-Ray-Man: I added a flag on the PropertyEmbedding. Not sure if that is so much better... But I am kind of out of ideas on this problem

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
The-Ray-Man: I think it makes sense to handle manual fields as viper fields. This are my reasons:
- I think it is kind of a expert feature. So I think we can trust the user to understand when it is safe to add a manual annotation on a field and when not. The user should think about inheritance etc. 
- As far as I know the idea behind the manual field is to allow the user to handle the permissions for that field themselves. If we would convert manual fields to functions or methods then there would not be any permissions to manage. 

I think we need to take the one of this two approaches:
1. We only allow `@Manual` annotations on fields that have default behaviour. Such a field is always treated as a viper field
2. We allow `@Manual` annotations on every field. We give the responsibility to handle open fields correctly to the user.

jesyspa: I agree it's an expert feature, but I don't see a case when you'd want to *not* convert things to a function if you could.  I guess the one thing I could see happening is a user marking an open val field as manual to indicate it really is a field and they plan to override it as var and handle the permissions manually.  But that's quite an edge case...

Maybe let's prohibit `@Manual` on `val` for now?

The-Ray-Man: That's a good point. So I will add the following behaviour:
- `@Manual` is not allowed on `val` fields (regardless of open vs closed)
- If a field has a `@Manual` it is always treated as a viper field.

I would suggest to throw an error if the annotation is added to a `val` field. Otherwise the user may become very confused.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -50,6 +50,12 @@ interface ProgramConversionContext {
     fun embedFunctionPretype(symbol: FirFunctionSymbol<*>): FunctionTypeEmbedding
     fun embedType(exp: FirExpression): TypeEmbedding = embedType(exp.resolvedType)
     fun embedProperty(symbol: FirPropertySymbol): PropertyEmbedding
+
+    /**
+     * Returns true if the property is having default behavior. That is:
```
jesyspa: nit: "if the property *has* default behavior"

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -399,27 +400,60 @@ class ProgramConverter(
         embedFunctionPretypeWithBuilder(symbol)
     }
 
+    @OptIn(SymbolInternals::class)
```
jesyspa: This is a pretty big function to opt into this.  What is it needd for?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -399,27 +400,60 @@ class ProgramConverter(
...
+                        // we can replace it with a function call
+                        PropertyEmbedding(
+                            CustomGetter(embedPureGetterFunction(symbol)),
+                            null,
+                            hasDefaultBehaviour = true
+                        )
+                    }
+
+                    else -> {
+                        // The property could be overwritten by anything, or has a custom getter/setter.
```
jesyspa: nit: overridden?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -399,27 +400,60 @@ class ProgramConverter(
...
+                val regularClass = classSymbol as? FirRegularClassSymbol ?: throw SnaktInternalException(
+                    symbol.source,
+                    "Properties dispatch receiver is not a regular class"
+                )
+
+                val isDefaultProperty = isGuaranteedDefaultProperty(symbol)
+                val isImmutable = symbol.isVal
+                val isManual = symbol.isManual(session)
+
+                return@getOrPutProperty when {
```
jesyspa, line 431: We had some design discussion around these choices.  I think this would be a good place to state the principles we used.

jesyspa (approved):
LGTM!  Thanks for getting this done, I think it should significantly simplify some stuff.  In particular, I think a logical follow-up is identifying how the SSA converter can be simplified now that there's no permissions to worry about (though there's still nullability, so maybe this doesn't change that much)

PR #156, opened 2026-05-13
Improved ADT error handling

pawinkler:
**Changes**:

- Multiple ADT errors were unified into `ADT_VIOLATION`
- Emitted errors are more precise and point to the exact location of a violation (if possible)
- Replaces the `isValid` flag in `AdtTypeEmbedding` with nullable entries in the `TypeResolver`
- Introduces a special `InvalidPretypeEmbedding` to express invalid type references 
- Prevents the generation of function bodies referencing an invalid ADT.
- Prevents the generation of functions and methods which reference an invalid ADT in their signature
- Prevents compiler-derived function from being verified


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -280,40 +279,41 @@ class ProgramConverter(
         return embedding
     }
 
-    private fun embedAdtClass(symbol: FirRegularClassSymbol): AdtTypeEmbedding {
+    private fun embedAdtClass(symbol: FirRegularClassSymbol): AdtTypeEmbedding? {
         val className = symbol.classId.embedName()
-        typeResolver.lookupAdt(className)?.let { return it }
-        return AdtTypeEmbedding(className, isValid = validateAdt(symbol)).also { typeResolver.registerAdt(it) }
+        if (typeResolver.isAdtKnown(className)) return typeResolver.lookupAdt(className)
```
jesyspa: This seems like it should be a getOrPut

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -280,40 +279,41 @@ class ProgramConverter(
...
+        if (!validateAdtSignature(symbol)) {
```
jesyspa: Every access to an invalid ADT will re-check validity (and potentially throw on errors?).  Would it be possible to cache this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -87,24 +88,22 @@ class ProgramConverter(
...
-            adts = typeResolver.adtTypeEmbeddings().filter { it.isValid }.map { it.toViper() },
+            adts = typeResolver.adtTypeEmbeddings().map { it.toViper() },
         )
 
 
     fun registerForVerification(declaration: FirSimpleFunction) {
         // Note: it is important that `body` is only set after embedding is complete, as we need to
         // place the embedding in the map before processing the body.
-        if (declaration.symbol.isPure(session)) {
+        val embedding: CallableEmbedding = if (declaration.symbol.isPure(session)) {
```
jesyspa: Is the explicit type necessary?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -61,11 +62,13 @@ class TypeResolver {
 
     fun lookupClassTypeEmbedding(name: SymbolicName) = classEmbedding[name] ?: interfaceEmbedding[name]
 
-    fun registerAdt(typeEmbedding: AdtTypeEmbedding) = adtEmbedding.put(typeEmbedding.name, typeEmbedding)
+    fun registerAdt(typeEmbedding: AdtTypeEmbedding) = adtEmbedding.putIfAbsent(typeEmbedding.name, typeEmbedding)
```
jesyspa: This silently drops the put if the ADT is already registers, it would be better to catch this as an error.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/CallableEmbedding.kt
```diff
@@ -19,6 +19,11 @@ interface CallableEmbedding {
     fun insertCall(args: List<ExpEmbedding>, ctx: StmtConversionContext): ExpEmbedding
 }
 
+fun CallableEmbedding.signatureIsValid(): Boolean {
+    val ct = callableType
+    return (listOfNotNull(ct.dispatchReceiverType, ct.extensionReceiverType) + ct.paramTypes + ct.returnType).all { it.isValid }
```
jesyspa: This is a rather complicated expression, perhaps it can be simplified?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/CallableEmbedding.kt
```diff
@@ -19,6 +19,11 @@ interface CallableEmbedding {
     fun insertCall(args: List<ExpEmbedding>, ctx: StmtConversionContext): ExpEmbedding
 }
 
+fun CallableEmbedding.signatureIsValid(): Boolean {
```
jesyspa: I would expect this to defer to the signature.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionEmbedding.kt
```diff
@@ -52,5 +54,8 @@ class PureUserFunctionEmbedding(private val callable: RichCallableEmbedding) : P
 
     var body: Exp? = null
 
-    override fun viperFunction(ctx: TypeResolver): Function = callable.toViperFunction(ctx, body)
+    override fun viperFunction(ctx: TypeResolver): Function? {
+        if (!callable.signatureIsValid()) return null
```
jesyspa: This seems really strange to me: why would we want some functions to be missing like this?  If the signature isn't valid, verification as a whole doesn't make sense.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: It wouldn't work here. There is a difference between a key existing in the map and a value of `null`. If the key exists in the map, the ADT was already encountered. If the value is `null`, this means the ADT is invalid. With this, we can never access an invalid ADT by accident and we cache results.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: It is being cached, an invalid ADT has a key in the map, but a value of `null`. It would return at the previous statement.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/CallableEmbedding.kt
pawinkler: I could make an additional method returning all used types in a signature as a set, but I thought this would be less invasive. Should I change it?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionEmbedding.kt
pawinkler: I agree, but it would still be generated. For invalid pure functions and invalid types we stop the generation early, but not for a function with an invalid signature. I also considered this to be more elegant than the previous solution. It required us to manually check if any errors were collected using the `errorCollector`, this prevents accidental verification of invalid functions.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -280,40 +279,41 @@ class ProgramConverter(
...
+        if (!validateAdtSignature(symbol)) {
+            typeResolver.invalidateAdt(className)
+            return null
+        }
+        return AdtTypeEmbedding(className).also { typeResolver.registerAdt(it) }
     }
 
     @OptIn(DirectDeclarationsAccess::class)
-    private fun validateAdt(symbol: FirRegularClassSymbol): Boolean {
+    private fun validateAdtSignature(symbol: FirRegularClassSymbol): Boolean {
```
jesyspa: What does a signature of an ADT mean?  Typically, functions have signatures.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
jesyspa: I'm not a fan of using `null` to represent "this value has been checked and is invalid", since it means there's a special value you have to test for.

That said, if you use `null` for this, why wouldn't `getOrPut` work?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -61,11 +62,16 @@ class TypeResolver {
 
     fun lookupClassTypeEmbedding(name: SymbolicName) = classEmbedding[name] ?: interfaceEmbedding[name]
 
-    fun registerAdt(typeEmbedding: AdtTypeEmbedding) = adtEmbedding.put(typeEmbedding.name, typeEmbedding)
+    fun registerAdt(typeEmbedding: AdtTypeEmbedding) {
+        check(!isAdtKnown(typeEmbedding.name)) { "ADT '${typeEmbedding.name}' was already registered" }
+        adtEmbedding.putIfAbsent(typeEmbedding.name, typeEmbedding)
+    }
+    fun invalidateAdt(name: SymbolicName) = adtEmbedding.put(name, null)
```
jesyspa: Can this happen after the ADT has been registered?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FunctionEmbedding.kt
jesyspa: > For invalid pure functions and invalid types we stop the generation early, but not for a function with an invalid signature.

That seems like the logical way to change it.

jesyspa (changes requested):
Let's discuss this over our upcoming call.  I am not very happy that the code ends up with a lot of `isValid` checks.  Typically, such checks should happen when you can handle the violation of validity, but that doesn't seem to be the case here.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
pawinkler: Yes, the ADT is registered on discovery and initially valid. If one of the validity checks later fails, it is invalidated.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
pawinkler: Sorry, I meant the header. I've renamed it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
pawinkler: Please ignore my previous comment. I've improved this and now I'm using `getOrPut` without any deferred invalidation.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
```diff
@@ -27,8 +27,9 @@ class TypeResolver {
 
     /**
      * Collection of all AdtTypeEmbeddings.
+     * A key mapped to `InvalidTypeEmbedding` means the ADT was encountered but is invalid.
```
jesyspa: Maybe better to have a layer between these?  `AdtTypeEmbedding` be the interface, `AdtTypeEmbeddingImpl` the concrete implementation, and `ErrorAdtTypeEmbedding` be the invalid one.

jesyspa (approved):
One optional suggestion, other than that lgtm

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/TypeResolver.kt
pawinkler: I agree, thanks! I've changed it.

PR #158, opened 2026-05-13
Ignore Empty Pre/Postconditions

### LINKED ISSUE #78: Empty preconditions error
Attempting to convert programs with an empty preconditions gives the following error: Every statement in invariant block must be a pure boolean invariant.

Minimal test case for reproducing the error:

`fun f(Int a): Int{
preconditions { }
return a;
}
`

The-Ray-Man:
This PR resolves the issue that an internal error is thrown if a user has a empty lambda as pre/postcondition

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
@@ -302,9 +302,13 @@ data class InvariantsAndTriggers(
     val triggers: List<ExpEmbedding>
 )
 
+private fun FirBlock.isEmptyLambdaBody(): Boolean {
+    if (statements.isEmpty()) return false
+    return (statements.size == 1 && (statements.first() as? FirReturnExpression)?.result?.resolvedType?.isUnit ?: false)
+}
+
 fun StmtConversionContext.collectInvariants(block: FirBlock) = buildList {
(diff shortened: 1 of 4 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -302,6 +303,10 @@ data class InvariantsAndTriggers(
 )
 
 fun StmtConversionContext.collectInvariants(block: FirBlock) = buildList {
+    // if the lambda is empty the compiler inserts a return Unit
+    if (block.statements.size == 1 && (block.statements.first() as? FirReturnExpression)?.result?.resolvedType?.isUnit ?: false) {
```
jesyspa: Maybe worth defining a private helper so you can give this a name?

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/empty.kt
```diff
@@ -0,0 +1,44 @@
...
+    preconditions {
+        arg == true
+    }
+    postconditions<Boolean> {}
+    return arg
+}
+
+fun <!VIPER_TEXT!>testInsertedReturn<!>() {
+    preconditions {
+        <!RETURN_NOT_ALLOWED!>return<!> Unit
```
jesyspa: Note that this return would return from the outer function; maybe you mean `return@preconditions`?

jesyspa (approved):

PR #165, opened 2026-05-18
Add locality-checking for lambda captured locals

### LINKED ISSUE #146: Check locality of properties captured by closures

paganma:
This PR updates locality checking with respect to local variables captured by lambdas.

Instead of tracking declaration ownership inside the locality object, captured-local locality is now represented directly through the locality attribute. Locality is inferred from expressions themselves rather than through a separate data-flow analysis pass, with intermediate results cached while walking expression tails.

Note that since we do not track declaration ownership in locality, the locality type-checkers (e.g. locality assignment / call checkers) only report a mismatch if a `local` value is assigned to a `global` variable. Ownership checking is instead taken care of by the new `PropertyAccessLocalityChecker`.

The PR also improves diagnostics for context locality mismatches, making it clearer which locality was expected and which locality was inferred.


formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/Locality.kt
```diff
@@ -5,63 +5,7 @@
 
 package org.jetbrains.kotlin.formver.locality.plugin
 
-import org.jetbrains.kotlin.diagnostics.rendering.Renderer
-import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
-import org.jetbrains.kotlin.fir.symbols.impl.FirCallableSymbol
-import org.jetbrains.kotlin.formver.locality.plugin.Locality.Global
-import org.jetbrains.kotlin.formver.locality.plugin.Locality.Local
+typealias Locality = LocalityAttribute?
```
ligee: Do you need it only for transition? 
To me it looks unnecessary to keep, so unless you have a good reasun, I'd suggest to inline/drop it.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/ExpressionLocalityResolver.kt
```diff
@@ -5,33 +5,56 @@
...
+    @OptIn(SymbolInternals::class)
+    context(context: CheckerContext)
+    private fun extractLocalityOf(expression: FirExpression): Locality {
+        val expression = expression.unwrapCast()
+
+        return when (expression) {
+            is FirPropertyAccessExpression ->
+                (expression.calleeReference.symbol as? FirVariableSymbol<*>)?.fir?.resolveLocality()
+            else -> {
+                expression.resolvedType.locality ?:
```
ligee: Are type aliases covered here? Do we have tests for it?

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/ExpressionLocalityResolver.kt
```diff
@@ -5,33 +5,56 @@
...
+            return Factory { session -> ExpressionLocalityResolver(session) }
+        }
+    }
+
+    private val cacheFactory = session.firCachesFactory
+
+    @OptIn(SymbolInternals::class)
+    context(context: CheckerContext)
+    private fun extractLocalityOf(expression: FirExpression): Locality {
+        val expression = expression.unwrapCast()
```
ligee: What about smartcasts, should they be handled specifically too? 
Do we have appropriate tests?

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeLocalityAttributeChecker.kt
```diff
@@ -7,29 +7,48 @@ package org.jetbrains.kotlin.formver.locality.plugin
...
+    }
+
+    private fun CheckerContext.isValidFunctionTypeArgumentTarget(typeRef: FirResolvedTypeRef): Boolean {
+        if (containingElements.none { it.isValidLocalityTarget() }) return false
+
+        return when (val target = containingElements.dropLast(1).lastOrNull()) {
+            is FirFunctionTypeParameter ->
+                true
+            is FirFunctionTypeRef ->
+                target.receiverTypeRef === typeRef || target.contextParameterTypeRefs.any { it === typeRef }
```
ligee: Why do you think it is safe to use identity comparison here?

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/VariableLocalityResolver.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.jetbrains.kotlin.fir.analysis.checkers.context.CheckerContext
+import org.jetbrains.kotlin.fir.declarations.FirVariable
+import org.jetbrains.kotlin.fir.expressions.FirExpression
+import org.jetbrains.kotlin.fir.types.FirImplicitTypeRef
+import org.jetbrains.kotlin.fir.types.FirTypeRef
+import org.jetbrains.kotlin.fir.types.coneType
+
+private fun FirTypeRef.isImplicit(): Boolean =
+    this is FirImplicitTypeRef ||
+            source?.kind == KtFakeSourceElementKind.ImplicitTypeRef
```
ligee: The second part looks redundant to me. Have you encountered some situation when it was necessary?

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/VariableLocalityResolver.kt
```diff
@@ -0,0 +1,32 @@
...
+
+context(context: CheckerContext)
+fun FirVariable.resolveLocality(): LocalityAttribute? =
+    resolveLocality { it.resolveLocality() }
+
+internal fun FirVariable.resolveLocality(
+    resolveExpressionLocality: (FirExpression) -> LocalityAttribute?
+): LocalityAttribute? {
+    if (!returnTypeRef.isImplicit()) return returnTypeRef.coneType.locality
```
ligee: Is it supposed to be used outside of the checkers phase?
If not, I'm surprised you need to check for implicit types.

ligee (changes requested):

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/Locality.kt
paganma: Originally this was to ease the transition, but in the end I opted for going back to using a concrete ADT for Locality as it seems more explicit and in line with what I am doing with the uniqueness checker. The ADT is now just a simple Enum but I think it makes things more explicit than just using `null` as the `Global` value.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/ExpressionLocalityResolver.kt
paganma: Type aliases cannot be annotated with locality attributes, as I show in the new `invalid.kt` test.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/ExpressionLocalityResolver.kt
paganma: Smart casts are handled by `unwrapExpression` I made `extractLocalityOf` explicitly use `unwrapExpression` before traversing it and I added a few tests in `assign_local.kt` that exercise this case.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeLocalityAttributeChecker.kt
paganma: This was for a half-experimental version of locality for function types. For this PR I scrapped this function as there is an improved version in the implementation for locality contracts, and as of this PR the ability of adding locality to function types is still not available.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/VariableLocalityResolver.kt
paganma: This function is redundant, but removing the second clause makes the tests fail on cases where locals are initialized with casts. For example
```kotlin
var a = x as A
```
where `x` is local will give a locality violation, but in this case since the user hasn't specified the type of `a`, it should be immediately assumed to be of type `local`.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/VariableLocalityResolver.kt
paganma: This function is called by `ExpressionLocailtyResolver` which lazily computes the locality of an expression during the checkers phase. The reason for checking for implicit types is to provide a behavior that is consistent with Kotlin's type inference i.e. If the type of a variable is not explicitly declared to be `local` or `global` then it is inferred by the initializer expression. I added a comment to this function to clarify this.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/ExpressionLocalityResolver.kt
ligee: I actually meant a different scenario. Sorry for being vague. E.g.:
```
class A
typealias B = A
fun share(x: B) {}
fun test(x: @Borrowed A) { share(x) }
```
Will you be able to detect the locality mismatch? 
Maybe it is a wrong example, because of the way the checking will be performed, but maybe in some other example the incoming marked type will be typealiased for some reason.
Because at some places in the compiler you may get typealias instead of the underlying type, and need to unwrap it using `fullyExpandedType`, to get to the attribute holder.

ligee (approved):
I'm not insisting on doing anything about typealiases, but if I pointed out a possible real issue with them in my comment, please  at least create an YT issue to address it in the future.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/ExpressionLocalityResolver.kt
paganma: I added a new test in `call_with_typealias` with this exact example and the locality mismatch is correctly detected. In this case `fullyExpandedType` because the checked locality attribute is resolved directly on the argument declaration. Although the locality attribute is a property of the type, it must be resolvable directly from the declaration, as adding the attribute to any other declaration type will always result in an error.

PR #166, opened 2026-05-18
Eliminate ErrorCollector, report diagnostics directly

komiputer:
## Summary

Conversion now reports diagnostics through the FIR `DiagnosticReporter` instead of buffering in an `ErrorCollector` and draining at the end of the checker.

- `ProgramConverter` takes `(DiagnosticContext, DiagnosticReporter)` and exposes `reportPurityViolation` / `reportAdtViolation` / `reportMinorInternalError` helpers, plus a `hadConversionError` bit the checker reads to decide whether to skip `linearizeAll` / `buildProgram`.
- The per-function "references an invalid ADT" summary that #156 added in `convertAll`'s loop used a cumulative `collectedAdtError()` check that was only correct because the checker registers one declaration per invocation. It now uses an `adtErrorCount` snapshot captured in `register()` before signature embedding, so signature-, body- and transitive-callee touches all attribute to the right function even when multiple declarations are batched.
- `PURITY_VIOLATION`, `ADT_VIOLATION` and `MINOR_INTERNAL_ERROR` move into a new `ConversionErrors` container in `core` (with its own renderer factory), so `core` can report them without a dependency on the plugin module. The plugin's extension registrar registers the new container alongside the existing two.
- The uniqueness module never actually wrote to its `ErrorCollector` parameter (it threw `UniquenessCheckException` instead), so the dead parameter is dropped.
- `formver.common/.../ErrorCollector.kt` is deleted.

## Test plan

- [x] Full plugin test suite green (141 / 141, 0 failures)
- [x] Compiles across core / plugin / uniqueness / locality
- [ ] Reviewer check: the `context(diagnosticContext) { reporter.reportOn(...) }` pattern in the `ProgramConverter` helpers is how this PR re-introduces the context parameter from a regular value; happy to switch to threading `context(_)` through the conversion call chain instead if you'd prefer.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -42,6 +41,18 @@ interface ProgramConversionContext {
     val convertedBodyResolver: ConvertedBodyResolver
     val linearizedBodyResolver: LinearizedBodyResolver
 
+    /** Number of ADT-violation diagnostics reported so far; used to detect ADT touches per iteration. */
```
jesyspa: Maybe this would be worth putting in a different context, e.g. ErrorCollectionContext?  ProgramConverter can still implement both, but this makes it clear what a function really requires (to avoid ProgConvCtx becoming a god interface).

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -43,16 +48,60 @@ import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 class ProgramConverter(
     val session: FirSession,
     override val config: PluginConfiguration,
-    override val errorCollector: ErrorCollector
+    private val diagnosticContext: DiagnosticContext,
+    private val reporter: DiagnosticReporter,
 ) : ProgramConversionContext {
+
+    /**
+     * True once any blocking error (purity or ADT) has been reported. The checker reads this to
```
jesyspa: Please document what the property communicates, not how it is currently used.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -130,36 +182,39 @@ class ProgramConverter(
                     convertedBodyResolver.storeImpure(signature.name, it)
                 }
             }
-            if (errorCollector.collectedAdtError()) {
-                errorCollector.addAdtError(
+            // Snapshot is taken in [register] before signature embedding so that ADT errors
```
jesyspa: I realize this is done to be consistent with existing behaviour, but it is not actually desirable: it's a special case that says "we decided not to convert this function for reason X".  If we want to surface this reasoning to the user, we should do it uniformly.  Please just remove it and have a uniform "function was supposed to be verified but isn't due to user error" (phrased somewhat nicer, of course)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -43,16 +48,60 @@ import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
...
+
     private val specialFunctions: Map<SymbolicName, SpecialKotlinFunction> = buildMap {
         putAll(SpecialKotlinFunctions.byName)
         putAll(PartiallySpecialKotlinFunctions.generateAllByName())
     }
     private val impureFunctions: MutableMap<SymbolicName, UserFunctionEmbedding> = mutableMapOf()
     private val pureFunctions: MutableMap<SymbolicName, PureUserFunctionEmbedding> = mutableMapOf()
 
-    private val registered: MutableList<Triple<FirSimpleFunction, FullNamedFunctionSignature, ReturnTarget>> = mutableListOf()
+    private data class RegisteredFunction(
```
jesyspa, line 98: I like this addition.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -43,16 +48,66 @@ import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
 class ProgramConverter(
     val session: FirSession,
     override val config: PluginConfiguration,
-    override val errorCollector: ErrorCollector
+    private val diagnosticContext: DiagnosticContext,
+    private val reporter: DiagnosticReporter,
 ) : ProgramConversionContext {
+
+    /**
+     * Whether any blocking diagnostic (a purity or ADT violation) has been reported.
```
jesyspa: I think *any* diagnostic is blocking, we just weren't accounting for them previously.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/diagnostics/ErrorCollectionContext.kt
```diff
@@ -0,0 +1,25 @@
...
+ * builders, etc.) can take only what it needs.
+ */
+interface ErrorCollectionContext {
+    /** Report a purity violation at [source]. */
+    fun reportPurityViolation(source: KtSourceElement?, msg: String)
+
+    /** Report an ADT violation at [source]. */
+    fun reportAdtViolation(source: KtSourceElement?, msg: String)
+
+    /** Report a non-blocking internal-error notice; the source is supplied by the implementation. */
```
jesyspa: nit: these errors are still blocking.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/diagnostics/ErrorCollectionContext.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package org.jetbrains.kotlin.formver.core.diagnostics
+
+import org.jetbrains.kotlin.KtSourceElement
+
+/**
+ * Surface for emitting conversion diagnostics.
+ *
+ * Kept separate from `ProgramConversionContext` so that subordinate code (purity checks, embedding
```
jesyspa: nit: "kept" is talking about the history of how this code came to be.  Comments should be about the code as it is; please rephrase so that the comment is useful for the reader.  (This is general advice, too.)

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -43,16 +49,62 @@ import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
...
     private val impureFunctions: MutableMap<SymbolicName, UserFunctionEmbedding> = mutableMapOf()
     private val pureFunctions: MutableMap<SymbolicName, PureUserFunctionEmbedding> = mutableMapOf()
 
-    private val registered: MutableList<Triple<FirSimpleFunction, FullNamedFunctionSignature, ReturnTarget>> = mutableListOf()
+    private data class RegisteredFunction(
+        val declaration: FirSimpleFunction,
+        val signature: FullNamedFunctionSignature,
+        val returnTarget: ReturnTarget,
+        /** [blockingErrorCount] snapshot taken just before this declaration's signature was embedded. */
+        val blockingErrorsBeforeRegister: Int,
```
jesyspa: I don't think this makes sense.  If any function has conversion errors, don't verify any function.  Tracking the number of blocking errors shouldn't be necessary.

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/diagnostics/ErrorCollectionContext.kt
komiputer: Stale doc — the 'non-blocking' wording is from before 28c47541 promoted these to blocking. Will drop it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/diagnostics/ErrorCollectionContext.kt
komiputer: Fair, will rephrase to describe what the interface is for as it stands, without the split-out history.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
komiputer: Agreed — dropping the per-function snapshot. If `hadConversionError` is true after `validateAll`, no function is verified; I'll emit the `VERIFICATION_SKIPPED` summary for every registered declaration so the surfacing stays uniform (per your earlier comment on this file).

jesyspa (approved):

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
```diff
@@ -53,20 +53,12 @@ class ViperPoweredDeclarationChecker(private val session: FirSession, private va
...
             programConversionContext.validateAll()
-            errorCollector.forEachPurityError { source, errorMessage ->
-                reporter.reportOn(source, PluginErrors.PURITY_VIOLATION, errorMessage)
-            }
-            errorCollector.forEachAdtError { source, errorMessage ->
-                reporter.reportOn(source, PluginErrors.ADT_VIOLATION, errorMessage)
-            }
-            if (errorCollector.collectedPurityError()) return
-            if (errorCollector.collectedAdtError()) return
+            if (programConversionContext.hadConversionError) return
```
The-Ray-Man, line 76: When we do the early return here, then we can't dump a wrong ExpEmbedding anymore. This could be useful if someone wants to investigate why a function is not accepted as pure. 

Generally, I think it would make sense to dump the ExpEmbedding before the program. Since it is earlier in the pipeline.

The-Ray-Man (approved):
One minor consern but generally looks very nice.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
jesyspa: Good point, let's move it to before here.

komiputer: Done in d21ff5df — moved the EXP_EMBEDDING dump above the `hadConversionError` early return and switched it to a `SimpleNameResolver` (the `ShortNameResolver` is not resolved yet at that point). Updated the `full_viper_dump` golden accordingly.

PR #167, opened 2026-05-19
Adapted Predicates to Final Fields

The-Ray-Man:
This PR adds the necessary information to the unique predicates. This includes the following changes:

- The `ALWAYS_READABLE` AccessPolicy was removed. It is not neaded anymore, because it was used for val fields. But val fields are now replaced by function calls. 
- To still be able to distinct different types of `PropertyEmbedding`, I added some fields on it, which are very close to the once that we have on the `UserField`.  We need those, because if we perform actions on fields of a class, we can no longer work with just the `FieldEmbeddings`, because we also have the `field-functions`. So we need to move that logic to the `PropertyEmbedding` level, and need the necessary information there.
- Because of the above change we needed to updated the `ClassPredicateBuilder` to also allow to add invariants for each property.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
@@ -266,9 +266,7 @@ class ProgramConverter(
         val name = symbol.embedGetterName(this)
         val signature = ImpureGetterFunctionSignature(name, symbol)
         val callable = NonInlineNamedFunction(signature)
-        val function = UserFunctionEmbedding(
-            callable
-        )
+        val function = UserFunctionEmbedding(callable)
         impureFunctions.putIfAbsent(name, function)
         return callable
     }
@@ -279,9 +277,7 @@ class ProgramConverter(
         val returnType = embedType(symbol.resolvedReturnType)
         val signature = PureGetterFunctionSignature(name, symbol, classType, returnType)
         val callable = NonInlineNamedFunction(signature)
-        val function = PureUserFunctionEmbedding(
-            callable
-        )
+        val function = PureUserFunctionEmbedding(callable)
         pureFunctions.putIfAbsent(name, function)
         return callable
     }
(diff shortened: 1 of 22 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -262,36 +262,39 @@ class ProgramConverter(
...
         val name = symbol.embedGetterName(this)
-        return impureFunctions.getOrPut(name) {
-            val signature = ImpureGetterFunctionSignature(name, symbol)
-            UserFunctionEmbedding(
-                NonInlineNamedFunction(signature)
-            )
-        }
+        val signature = ImpureGetterFunctionSignature(name, symbol)
+        val callable = NonInlineNamedFunction(signature)
+        val function = UserFunctionEmbedding(
```
jesyspa: nit: Formatting looks a bit weird

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -262,36 +262,39 @@ class ProgramConverter(
...
-            val signature = PureGetterFunctionSignature(name, symbol, classType, returnType)
-            PureUserFunctionEmbedding(
-                NonInlineNamedFunction(signature)
-            )
-        }
+        val classType = embedType(symbol.dispatchReceiverType!!)
+        val returnType = embedType(symbol.resolvedReturnType)
+        val signature = PureGetterFunctionSignature(name, symbol, classType, returnType)
+        val callable = NonInlineNamedFunction(signature)
+        val function = PureUserFunctionEmbedding(
```
jesyspa: idem

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -636,10 +648,13 @@ class ProgramConverter(
         return backingField
     }
 
-    private fun embedCustomProperty(symbol: FirPropertySymbol) = PropertyEmbedding(
+    private fun embedExtensionProperty(symbol: FirPropertySymbol) = PropertyEmbedding(
         CustomGetter(embedImpureGetterFunction(symbol)),
         symbol.isVar.ifTrue { CustomSetter(embedSetterFunction(symbol)) },
-        hasDefaultBehaviour = false
+        hasDefaultBehaviour = false,
+        false,
```
jesyspa: nit: would be nice to specify the arg name, so people reading the code can see what it means

jesyspa (approved):
Some nits, but looks good generally :)

PR #169, opened 2026-05-20
Refactor Callable Embedding Type Hierarchy

### LINKED ISSUE #159: Having Recursion in the Contract Results in Stack Overflow
Example:

```kotlin
@Pure
fun fact(n: Int): Int {
    preconditions {
        n >= 0
    }
    postconditions<Int> { res ->
        (n == 0) implies (res == 1)
        (n > 0) implies  (res == n * fact(n - 1))
    }

    return if (n == 0) {
        1
    } else {
        n * fact(n - 1)
    }
}

```

This results in a stack overflow. The loop happens when embedding the function signature.

The-Ray-Man:
This PR resolves the issue with recursive contracts. In the process the embedding process was cleaned up.  There are also some changes in the type hierarchy of signatures and callables.

## Changes in the Type Hierarchy
Now, the signatures are callable ones they are on the `NonInlineNamedFunctionSignature`/`InlineFunctionSignature` level.
- For **Non-Inline**, the necessary things are the name and signature. Here a contract is not required.
- For **Inline**, the necessary things are the name, signature as well as the contract. Because ideally, when calling an inline function we want to `assert` and `assume` the pre+post condition of the inlined function. 

## Changes in the ProgramConverter
To mitigate the recursion problem the `callable` map was introduced. This stores all the functions that can be called. Furthermore, instead of having the pure and impure functions, we just have `fullFunctionSignatures` which stores functions that have their contract already embedded. The pure vs impure separation is not necessary, since every signature has a boolean flag.


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -25,7 +24,22 @@ import org.jetbrains.kotlin.formver.core.names.CatchLabelName
...
+    val signature: S,
+    val returnTarget: ReturnTarget
+) {
+    /**
+     * Refines the signature by applying the given action to the current signature.
+     * The return target is preserved.
+     *
+     * It should be used to make a signature more specific, e.g., to add pre+post conditions to an existing signature
+     */
+    fun <T : S> refineSignature(action: (SignatureWithTarget<S>) -> T): SignatureWithTarget<T> =
```
jesyspa: I will try to do a full review as soon as I can, but I wanted to mention that this makes `SignatureWithTarget` a comonad (namely, the environment comonad: https://hackage-content.haskell.org/package/comonad-5.0.10/docs/Control-Comonad-Env.html )

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -207,52 +208,24 @@ class ProgramConverter(
...
-        for ((name, special) in specialFunctions) {
-            // Fully-special functions never produce a Viper method; partially-special ones produce a
-            // header only when their base embedding has been initialised.
-            val callable = (special as? PartiallySpecialKotlinFunction)?.baseEmbedding?.callable ?: continue
-            callable.toViperMethodHeader(typeResolver)?.let { linearizedBodyResolver.storeMethod(name, it) }
-        }
-        for ((name, embedding) in impureFunctions) {
-            val callable = embedding.callable
-            val source = callable.declarationSource
+        val (pure, impure) = fullSignatures.entries.partition { it.value.isPure }
```
jesyspa, line 208: Maybe worth just splitting the bodies of the two loops into their own functions, so you can delegate to the correct one?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+        val isExtensionReceiverBorrowed = symbol.receiverParameterSymbol?.isBorrowed(session) ?: false
+
+        val returnType = embedType(symbol.resolvedReturnType)
+
+        val returnTarget = when {
+            symbol.isPure(session) -> ReturnTarget.createForPureFunction(returnType)
+            else -> returnTargetProducer.getFresh(returnType)
+        }
+
+        val signature = object : FunctionSignature {
```
jesyspa: I thought a recent goal was to avoid creating signature objects like this.  What makes this case different?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+                override val name: SymbolicName = name
+                override val symbol: FirFunctionSymbol<*> = symbol
+                override val labelName: String
+                    get() = super<NamedFunctionSignature>.labelName
+            }
+        }
+
+
+    /**
+     * Embeds an inline function. This also includes to embed the provided contract.
```
jesyspa: nit: signature, also grammar is a bit wonky.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -25,7 +24,22 @@ import org.jetbrains.kotlin.formver.core.names.CatchLabelName
 import org.jetbrains.kotlin.formver.core.names.TryExitLabelName
 import org.jetbrains.kotlin.formver.viper.NameResolver
 
+data class SignatureWithTarget<S : FunctionSignature>(
```
jesyspa: `out`, I think?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+                current.signature.userFunctionPostcondition(typeResolver) + postcondition
+            )
+            if (!symbol.neverConvert(session)) {
+                fullSignatures.putIfAbsent(fullSignature.name, fullSignature)
+            }
+            fullSignature
+        }
+
+
+        return SignatureWithTarget(fullSignature, returnTarget)
```
jesyspa: This repackaging shouldn't be necessary.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+        if (callable.isPure && !symbol.neverConvert(session)) {
+            val declaration = symbol.fir as? FirSimpleFunction ?: throw SnaktInternalException(
+                symbol.source, "Expected FirSimpleFunction, got unexpected type ${symbol.fir.javaClass.simpleName}"
             )
+
+            val context = createBodyConversionContext(callable, returnTarget)
+            val body = context.convertPureBody(declaration)
+            convertedBodyResolver.storePure(callable.name, body)
+        }
+
```
jesyspa, line 363: nit: empty line

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+    }
+
+
+    /**
+     * Embeds the full function signature (e.g. with pre+post conditions). If necessary, the body of the function is stored as well.
+     */
+    private fun embedFullFunctionAndBody(symbol: FirFunctionSymbol<*>): CallableNamedSignature {
```
jesyspa: This is the so manyeth embedSignature function here.  Maybe that's a responsibility that can be delegated?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+
+        callable[symbol.embedName(this)]?.let { return it.signature }
+
+        val (callableEmbedding, returnTarget) = embedCallable(symbol)
+        // Inline functions, which should not be converted, don't need a signature.
```
jesyspa: Does it hurt to give them one anyway?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
 
-    override fun embedFunction(symbol: FirFunctionSymbol<*>): CallableEmbedding {
-        val lookupName = symbol.embedName(this)
-        specialFunctions[lookupName]?.let { existing ->
+    /**
+     * Returns the callable embedding if a matching special function exists.
+     */
+    fun embedSpecialFunction(symbol: FirFunctionSymbol<*>): CallableEmbedding? {
+        val name = symbol.embedName(this)
+        return specialFunctions[name]?.let { existing ->
```
jesyspa: I think `also` would make more sense here

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,74 +278,263 @@ class ProgramConverter(
...
+            existing
         }
     }
 
-    override fun embedPureFunction(symbol: FirFunctionSymbol<*>): PureUserFunctionEmbedding {
-        val (returnTarget, signature) = embedFullSignature(symbol)
-        return embedPureUserFunction(symbol, signature, returnTarget)
-    }
+    /**
+     * This method should be used to lookup a callable embedding for a function symbol.
```
jesyspa: I would prefer a clearer statement here.   Do you mean that this is the public interface?  You have lots of functions that do embeddings, I assume all should be used in some context, so what makes this the one that "should" be used?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
```diff
@@ -36,8 +38,57 @@ interface FullNamedFunctionSignature : NamedFunctionSignature {
     val postconditions : List<ExpEmbedding>
 
     val declarationSource: KtSourceElement?
+
+    override fun toViperMethodHeader(ctx: TypeResolver): Method? {
```
jesyspa: I don't like this :( you end up with duplicate logic between this and the function header below, and you make the caller figure out which function should be called.  Maybe the two can just be unified?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
```diff
@@ -36,8 +38,57 @@ interface FullNamedFunctionSignature : NamedFunctionSignature {
...
+            null,
+        )
+    }
+}
+
+fun FullNamedFunctionSignature.toViperFunction(
+    ctx: TypeResolver,
+    body: Exp?,
+): UserFunction {
+    postconditions.forEach { postcondition ->
```
jesyspa: By the way, I think we should consider checking this way, way earlier.  Not a blocker here, though.

jesyspa (changes requested):
General thoughts: I think this is going in the right direction, but it introduces more cases where we special-case pure vs impure things.  Could you look into whether this gap can be closed at least somewhat?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
The-Ray-Man: Technically yes, but in the `refineSignature` function below, we have `T: S`.  And there `S` appears in a `in` position.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
The-Ray-Man: Not really, they will just show up in the final viper program.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
The-Ray-Man: I totally agree

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullNamedFunctionSignature.kt
The-Ray-Man: Yes, I agree this is not elegant at all. This setup comes from the ``RichCallableEmbedding`` and I am not sure how we can get rid of it. There are multiple problem points:
- The construction of a ``UserMethod`` and `UserFunction` are similar but still different. The Method takes a declaration as return, where the Function only takes the type. 
- The body, is once an sequence statement and in the other case an expression.
- The viper ``Method`` and `Function` do not share a usable common supertype.
Generally, I feel like that the our viper package is very close to the silver implementation so I am not sure how much we should change there.

The-Ray-Man: I just realized I could just remove the ``toViperFunctionHeader`` because it is never used.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -25,9 +26,13 @@ import org.jetbrains.kotlin.formver.core.names.CatchLabelName
 import org.jetbrains.kotlin.formver.core.names.TryExitLabelName
 import org.jetbrains.kotlin.formver.viper.NameResolver
 
+
```
jesyspa: nit: is there a need for these extra lines?  I think it makes things harder to read

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
```diff
@@ -41,14 +46,14 @@ interface ProgramConversionContext : ErrorCollectionContext {
...
 
-    fun embedFunction(symbol: FirFunctionSymbol<*>): CallableEmbedding
-    fun embedPureFunction(symbol: FirFunctionSymbol<*>): PureUserFunctionEmbedding
     fun embedAnyFunction(symbol: FirFunctionSymbol<*>): CallableEmbedding
-    fun embedFunctionSignature(symbol: FirFunctionSymbol<*>): Pair<ReturnTarget,FunctionSignature>
     fun embedType(type: ConeKotlinType): TypeEmbedding
     fun embedFunctionPretype(symbol: FirFunctionSymbol<*>): FunctionTypeEmbedding
     fun embedType(exp: FirExpression): TypeEmbedding = embedType(exp.resolvedType)
     fun embedProperty(symbol: FirPropertySymbol): PropertyEmbedding
+    fun embedProvidedContract(
```
jesyspa: nit: What do you mean by `provided` here?  Why not just `embedContract`?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -140,37 +142,18 @@ class ProgramConverter(
         adts = typeResolver.adtTypeEmbeddings().map { it.toViper() },
     )
 
+
+    // START PIPELINE
```
jesyspa: nit: region would probably be helpful.  That said, I don't think methods should really be structured into a pipeline: that tends to be a sign there's implicit dependencies that you'd ratehr not have.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -199,6 +182,26 @@ class ProgramConverter(
         }
     }
 
+
+    private fun linearizePure(name: SymbolicName, signature: CompleteFunctionSignature) {
+        val converted = convertedBodyResolver.lookupPure(name)
+        val linearized = converted?.let { linearizePureBody(signature.declarationSource, it) }
+        linearizedBodyResolver.storeFunction(name, signature.toViperFunction(typeResolver, linearized))
+
```
jesyspa: nit: extra line

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConverter.kt
```diff
@@ -305,73 +276,147 @@ class ProgramConverter(
...
+    ): Pair<List<ExpEmbedding>, List<ExpEmbedding>> {
+        val kotlinContractPostcondition = embedKotlinContract(symbol, signature)
+        val userContract = embedFormverContract(symbol, signature, returnTarget)
+        return Pair(userContract.first, kotlinContractPostcondition + userContract.second)
     }
 
+    // END CONTRACTS
+
+
+    // START CLASSES (also ADT)
```
jesyspa: I've removed the ADT-related code since it wasn't contributing anything; I think the rebase should be mechanical, though perhaps not automatic.

jesyspa (approved):
I think this change is a big improvement, but the splitting of the programconverter makes me wonder: is grouping all those things together into a single class really the right call?  Or should we split it up more thoroughly?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/ProgramConversionContext.kt
The-Ray-Man: In my mind ``Contract`` refers to the full contract a function has. Including permissions, type information, etc. This function, however, only returns the contract that was provided by the user (so the union of the kotlin contract and our formver contract.)
But I think ``embedContract`` works well too. I will change it.

The-Ray-Man:
> is grouping all those things together into a single class really the right call? Or should we split it up more thoroughly?

I thought about that too. I actually tried to split it once and here is what my experience was. The split I tried was:
- `GlobalConverter`: Manages the registration, body conversion etc. This also contains most of the registration things (freshVariableProvder etc.)
- `TypeConverter`: This would convert types and classes as well.
- `PropertyConverter`
- `SignatureConverter`

The, to some extend, annoying thing is that most of them are dependent on each other. `SignatureConverter` needs access to the `TypeConverter`. The `TypeConverter` needs access to the `PropertyConverter`, because if we embed the Class we also need to embed the corresponding properties. `PropertyConverter`  needs access to the `SignatureConverter` because of getter and setter methods.

So, I guess what one would come up with is a structure where they all hold references to each other, or references to the parent and the parent holds references to the "sub-converters". But this is in my opinion does not differ that much from just having one class, where the methods are grouped into regions.

PR #170, opened 2026-05-22
Visitor-ify ExpEmbedding validity; SAMify PurityContext

komiputer:
## Summary

Move per-node validity dispatch out of `ExpEmbedding`:

- Replace `ExpEmbedding.isValid` (one non-default override on `Assert`) with `ValidityVisitor : ExpVisitor<Boolean>` next to `ExprPurityVisitor` in `core/purity/`.
- `PurityContext` becomes a `fun interface`; `DefaultPurityContext` is deleted and constructed inline as a lambda at its sole call site in `checkValidity`.
- Move `preorder` and `expressionSource` from `core/purity/ExprUtils.kt` to `core/embeddings/expression/Meta.kt`, next to `WithPosition` and `SharingContext`. The purity package no longer owns generic tree-walks; `FullNamedFunctionSignature` imports `preorder` from its new home.

`ExpEmbedding` no longer depends on the purity package, and the purity package owns its own dispatch.

## Notes

- `checkValidity` keeps the nullable `source: KtSourceElement?` parameter. Tightening it to non-null would force `ProgramConverter.validateAll` to compute a fallback for `declarationSource == null`, which it currently doesn't need. The internal `checkNotNull(src)` is preserved (same behavior, same error message); the brief offered either option.
- The only validity rule today (Assert condition must be pure) keeps its current semantics; no test gold changes expected.

## Test plan

- [x] `./gradlew :formver.compiler-plugin:detekt`
- [x] `./gradlew :formver.compiler-plugin:untilConversion`


jesyspa:
Closing: the main target of this refactor (`FullNamedFunctionSignature.kt`) was deleted by the callable-hierarchy refactor in #169, and the ExpVisitor landscape has since changed. Redoing the idea from scratch is cheaper than rebasing this.

PR #172, opened 2026-05-22
Restrict TypeInvariantHolder to TypeEmbedding; flatten TypeEmbeddingFlags

komiputer:
## Summary

Two coupled cleanups in the type-embedding layer:

**Restrict `TypeInvariantHolder` to `TypeEmbedding`.** Both `PretypeEmbedding` and `TypeEmbedding` implemented `TypeInvariantHolder` with same-named methods but different semantics. The pretype side returned raw invariants; the type side wrapped them via flag-aware `adjust*` helpers. Calling on a `PretypeEmbedding` directly silently stripped nullability handling and produced wrong Viper output.

`PretypeEmbedding` is now a `sealed interface` that no longer extends `TypeInvariantHolder`. The raw building blocks move to internal extension functions (`rawAccessInvariants`, `rawPureInvariants`, `rawUniquePredicateAccessInvariant`) that dispatch via `when` on the sealed hierarchy and are only callable from inside the `types/` package. The public invariant API is reached solely through `TypeEmbedding`, which delegates to the raw helpers and wraps results in `IfNonNullInvariant` when nullable. The redundant `subTypeInvariant()` override on `PretypeEmbedding` is removed entirely.

**Flatten `TypeEmbeddingFlags`.** The single-Boolean wrapper is gone; `TypeEmbedding` now has a plain `nullable: Boolean` field. The four `adjust*` helpers it carried were only used by `TypeEmbedding` itself and become private members or inline expressions. The 2024 TODO about "potentially other flags" never materialised; if a real second flag is ever needed the wrapper can be reintroduced then.

No behaviour change is expected. `./gradlew :formver.compiler-plugin:detekt :formver.compiler-plugin:untilConversion` passes (no golden-file diffs in conversion output).

## Test plan

- [x] `./gradlew :formver.compiler-plugin:detekt`
- [x] `./gradlew :formver.compiler-plugin:untilConversion`

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassPredicateBuilder.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/types/ClassPredicateBuilder.kt
@@ -34,7 +34,7 @@ internal class ClassPredicateBuilder private constructor(
         ): Predicate {
             val typeEmbedding = ctx.lookupClassTypeEmbedding(name)!!
             val builder = ClassPredicateBuilder(
-                TypeEmbedding(typeEmbedding, TypeEmbeddingFlags(nullable = false)),
+                TypeEmbedding(typeEmbedding, nullable = false),
                 ctx.lookupClassFields(name),
                 ctx.lookupSuperTypes(name)
             )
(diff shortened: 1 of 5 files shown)


jesyspa:
Not helpful: the wrapper is still useful for uniqueness and locality, I'd say.

PR #173, opened 2026-05-22
Delete dead Trafos AST metadata

komiputer:
## Summary

Every Viper AST node carried a `trafos: Trafos = Trafos.NoTrafos` constructor argument. `Trafos` had exactly one inhabitant (`NoTrafos`) and no code path ever constructed another value — the field plumbed through ~50 data classes and was passed to `toSilver` purely as the fixed Silver `NoTrafos\$.MODULE\$` literal.

- Drop the parameter from every Kotlin Viper AST data class (`Exp`, `Stmt`, `Declaration`, `Method`, `Function`, `Domain`, `Field`, `Predicate`, `PermExp`, `AdtDecl`, `Program`, `AccessPredicate`, `Label`, `OperatorFunctions`).
- Drop the matching parameter from `core/` plumbing helpers: `Applicable.toFuncApp`, `NamedFunctionSignature.{toMethodCall,toFuncApp}`, `VariableEmbedding.{toLocalVarDecl,toLocalVarUse}`, `OperatorExpEmbeddingBuilder.ViperCallData`, and the `viperImplementation { ... }` blocks in `OperatorExpEmbeddings`.
- Delete `Trafos.kt`. Replace it with a single `internal val silverNoTrafos: ErrorTrafo` in a new `SilverNoTrafos.kt`, which each `toSilver` body feeds inline to its Silver constructor at the boundary.

No behavior change is intended; Viper output is unchanged.

## Test plan

- [x] `./gradlew :formver.compiler-plugin:detekt` passes
- [x] `./gradlew :formver.compiler-plugin:untilConversion` passes (golden-file diffs surface here for output-preserving refactors; none observed)


jesyspa:
I think this is reasonable, we don't use this and don't plan to anytime soon.

The-Ray-Man (approved):

PR #174, opened 2026-05-22
Extract FieldAccessLowering for hierarchy-predicate unfolding

komiputer:
## Summary

Five sites duplicated the "lower a `FieldAccess` through the class
hierarchy's unique-predicate stack" operation: each one walked
`typeResolver.hierarchyPathTo(receiver, field)`, built a
`predicateAccess` per class, and emitted in its own shape (imperative
`Stmt.Unfold`, SSA predicate-access invariants, or nested
`Exp.Unfolding`). The underlying walk was identical at all four
linearizer call sites, but the `predicateAccess` helper on
`ClassTypeEmbedding` demanded an `ExpEmbedding` receiver, forcing each
caller to wrap an already-lowered Viper `Exp` in `ExpWrapper`.

This change introduces `FieldAccessLowering` (in
`core/linearization/`, next to the consumers) with two methods:

- `pathTo(receiverType, field): Sequence<ClassTypeEmbedding>` --- the
  hierarchy walk.
- `predicateAccessFor(receiverViper, classOnPath): Exp.PredicateAccess`
  --- the per-class predicate access, taking a raw `Exp` directly
  rather than going through `fillHole`/`pureToViper`.

The three linearizers (`Linearizer`, `LinearizationVisitor`'s
`visitFieldModification`, `PureFunBodyLinearizer`, `PureExpLinearizer`)
now each walk the path and pick their own per-class emit shape. The
old `predicateAccess` extension on `ClassTypeEmbedding` is deleted ---
it had no callers outside these four sites.

`ExpWrapper(receiverViper, receiverType)` construction sites: 6 → 2.
The two remaining sites in `LinearizationVisitor.kt` are unrelated
(equality lowering, type-cast folding); eliminating them is Slice H.

## Test plan
- [x] `./gradlew :formver.compiler-plugin:detekt` --- passes
- [x] `./gradlew :formver.compiler-plugin:untilConversion` --- passes;
  golden Viper output unchanged for every existing test.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
@@ -7,37 +7,36 @@ package org.jetbrains.kotlin.formver.core.linearization
 
 import org.jetbrains.kotlin.KtSourceElement
 import org.jetbrains.kotlin.formver.core.asPosition
-import org.jetbrains.kotlin.formver.core.conversion.TypeResolver
 import org.jetbrains.kotlin.formver.core.embeddings.properties.FieldEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.types.ClassTypeEmbedding
 import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
 import org.jetbrains.kotlin.formver.viper.ast.Exp
 import org.jetbrains.kotlin.formver.viper.ast.PermExp
+import org.jetbrains.kotlin.formver.viper.ast.Stmt
 
 /**
- * Owns the path-and-predicate computation for a [FieldAccess][org.jetbrains.kotlin.formver.core.embeddings.expression.FieldAccess]
- * on a class hierarchy.
- *
- * To reach a field declared on a superclass we must walk the chain of unique-predicates
- * from the receiver's runtime type down to the declaring class, and for each class on
- * that path emit one predicate-access expression. The three linearizers consume that
- * sequence in their preferred shape:
- *  - imperative linearization emits a `Stmt.Unfold` per access;
- *  - SSA linearization collects them as invariants on the assignment;
- *  - pure-expression linearization nests them as `Exp.Unfolding` around the body.
+ * The unique-predicate access on [classOnPath] that must be unfolded to reach a field on a superclass
+ * through [receiver]. Takes a raw [Exp] rather than going through `fillHole`/`pureToViper`.
  */
-class FieldAccessLowering(
-    private val typeResolver: TypeResolver,
-    private val source: KtSourceElement?,
-) {
-    fun pathTo(receiverType: TypeEmbedding, field: FieldEmbedding): Sequence<ClassTypeEmbedding> =
-        typeResolver.hierarchyPathTo(receiverType.pretype, field)
+fun hierarchyPredicateAccess(
+    receiver: Exp,
+    classOnPath: ClassTypeEmbedding,
+    source: KtSourceElement?,
+): Exp.PredicateAccess =
+    Exp.PredicateAccess(
+        classOnPath.uniquePredicateName,
+        listOf(receiver),
+        PermExp.FullPerm(),
+        source.asPosition,
+    )
 
-    fun predicateAccessFor(receiverViper: Exp, classOnPath: ClassTypeEmbedding): Exp.PredicateAccess =
-        Exp.PredicateAccess(
-            classOnPath.uniquePredicateName,
-            listOf(receiverViper),
-            PermExp.FullPerm(),
-            source.asPosition,
-        )
+/**
+ * Emits a `Stmt.Unfold` for each unique-predicate on the hierarchy path from [receiverType]
+ * down to the class declaring [field]. Shared by the imperative linearizer and the
+ * `FieldModification` visitor; both want the same imperative unfold shape.
+ */
+fun LinearizationContext.unfoldHierarchyPredicates(receiver: Exp, receiverType: TypeEmbedding, field: FieldEmbedding) {
+    for (classOnPath in typeResolver.hierarchyPathTo(receiverType.pretype, field)) {
+        addStatement { Stmt.Unfold(hierarchyPredicateAccess(receiver, classOnPath, source), source.asPosition) }
+    }
 }
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
@@ -31,12 +31,25 @@ fun hierarchyPredicateAccess(
     )
 
 /**
+ * The unique-predicate accesses to unfold to reach [field] through [receiver], in walk order
+ * (outermost class first). Returned ordered top-down: callers that need bottom-up nesting
+ * (e.g. [Exp.Unfolding] fold) reverse via `foldRight` or `toList().asReversed()`.
+ */
+fun LinearizationContext.hierarchyPredicateAccesses(
+    receiver: Exp,
+    receiverType: TypeEmbedding,
+    field: FieldEmbedding,
+): Sequence<Exp.PredicateAccess> =
+    typeResolver.hierarchyPathTo(receiverType.pretype, field)
+        .map { hierarchyPredicateAccess(receiver, it, source) }
+
+/**
  * Emits a `Stmt.Unfold` for each unique-predicate on the hierarchy path from [receiverType]
  * down to the class declaring [field]. Shared by the imperative linearizer and the
  * `FieldModification` visitor; both want the same imperative unfold shape.
  */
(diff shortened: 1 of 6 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+
+/**
+ * Owns the path-and-predicate computation for a [FieldAccess][org.jetbrains.kotlin.formver.core.embeddings.expression.FieldAccess]
+ * on a class hierarchy.
+ *
+ * To reach a field declared on a superclass we must walk the chain of unique-predicates
+ * from the receiver's runtime type down to the declaring class, and for each class on
+ * that path emit one predicate-access expression. The three linearizers consume that
```
jesyspa: Please remove this list, it doesn't help with understanding the function.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
```diff
@@ -0,0 +1,43 @@
...
+ * sequence in their preferred shape:
+ *  - imperative linearization emits a `Stmt.Unfold` per access;
+ *  - SSA linearization collects them as invariants on the assignment;
+ *  - pure-expression linearization nests them as `Exp.Unfolding` around the body.
+ */
+class FieldAccessLowering(
```
jesyspa: Why should this be a class rather than just two functions?  The usages all seem to be create+call a function

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
```diff
@@ -126,9 +122,12 @@ data class Linearizer(
...
                     if (field.unfoldToAccess) {
-                        val receiverWrapper = ExpWrapper(receiverViper, receiverType)
-                        val hierarchyPath = typeResolver.hierarchyPathTo(receiverType.pretype, field)
-                        hierarchyPath.unfoldHierarchyPath(receiverWrapper, this)
+                        val lowering = FieldAccessLowering(typeResolver, source)
+                        for (classOnPath in lowering.pathTo(receiverType, field)) {
+                            addStatement {
+                                Stmt.Unfold(lowering.predicateAccessFor(receiverViper, classOnPath), source.asPosition)
+                            }
+                        }
```
jesyspa: This seems like a common pattern.  It doesn't deserve a helper function?

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
komiputer: Removed in 741fc886. KDoc trimmed to the rationale only.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
komiputer: Agreed — the class held no real state. In 741fc886 the class is gone:

- `pathTo` was a thin delegation to `typeResolver.hierarchyPathTo(receiverType.pretype, field)`, so I inlined it at the three pure call sites (`PureExpLinearizer`, `PureFunBodyLinearizer`, and the new helper below).
- `predicateAccessFor` survives as a free function `hierarchyPredicateAccess(receiver, classOnPath, source)` — it still earns its keep by taking a raw `Exp` and bypassing `fillHole`/`pureToViper`.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/Linearizer.kt
komiputer: Done in 741fc886 as `LinearizationContext.unfoldHierarchyPredicates(receiver, receiverType, field)`. Both the imperative `Linearizer.addFieldAccessStoringIn` and `LinearizationVisitor.visitFieldModification` use it. The two remaining call sites in `PureExpLinearizer` and `PureFunBodyLinearizer` consume the path in different shapes (`Exp.Unfolding` fold and SSA invariants), so they still walk the path inline.

One small behaviour change worth flagging: the helper passes `source.asPosition` to `Stmt.Unfold`. The pre-refactor visitor site emitted with no position, which looked like an oversight next to the matching unfolds in `Linearizer`. Golden Viper output is unchanged, so positions evidently aren't printed for `Stmt.Unfold` in the snapshot.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.jetbrains.kotlin.formver.core.embeddings.properties.FieldEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.types.ClassTypeEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.types.TypeEmbedding
+import org.jetbrains.kotlin.formver.viper.ast.Exp
+import org.jetbrains.kotlin.formver.viper.ast.PermExp
+import org.jetbrains.kotlin.formver.viper.ast.Stmt
+
+/**
+ * The unique-predicate access on [classOnPath] that must be unfolded to reach a field on a superclass
+ * through [receiver]. Takes a raw [Exp] rather than going through `fillHole`/`pureToViper`.
```
jesyspa: "rather than going" is a historical note that isn't helpful; cut.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
```diff
@@ -0,0 +1,55 @@
...
+        classOnPath.uniquePredicateName,
+        listOf(receiver),
+        PermExp.FullPerm(),
+        source.asPosition,
+    )
+
+/**
+ * The unique-predicate accesses to unfold to reach [field] through [receiver], in walk order
+ * (outermost class first). Returned ordered top-down: callers that need bottom-up nesting
+ * (e.g. [Exp.Unfolding] fold) reverse via `foldRight` or `toList().asReversed()`.
```
jesyspa: Remove details about callers.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
```diff
@@ -0,0 +1,55 @@
...
+    receiver: Exp,
+    receiverType: TypeEmbedding,
+    field: FieldEmbedding,
+): Sequence<Exp.PredicateAccess> =
+    typeResolver.hierarchyPathTo(receiverType.pretype, field)
+        .map { hierarchyPredicateAccess(receiver, it, source) }
+
+/**
+ * Emits a `Stmt.Unfold` for each unique-predicate on the hierarchy path from [receiverType]
+ * down to the class declaring [field]. Shared by the imperative linearizer and the
```
jesyspa: Remove details about callers (here and elsewhere - this is not appropriate).

jesyspa (changes requested):

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
komiputer: Cut in 7c864e59.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
komiputer: Cut in 7c864e59. I kept the ordering guarantee but restated it as a property of the return value rather than instructions to callers: "ordered top-down: the class of `receiverType` first, the class declaring `field` last." Callers that want the other order can work that out themselves.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/FieldAccessLowering.kt
komiputer: Cut in 7c864e59. On "elsewhere": I grepped the diff for added comment lines, and this file is the only one in the PR that adds any, so the other three sites had nothing to remove.

komiputer:
Rebased onto current `main` (clean, no conflicts) and addressed the three KDoc comments in 7c864e59.

Re-ran on the rebased branch:
- `:formver.compiler-plugin:detekt` — passes
- `:formver.compiler-plugin:untilConversion` — passes, golden Viper output unchanged

jesyspa (approved):

PR #178, opened 2026-05-29
Support `old` expressions in contracts

The-Ray-Man:
This PR allows the user to use old in the contracts

The comment below is not true.
### Why is `old` not embedded as a `FullySpecialKotlinFunction` like `verify`?
In the Framework of the `FullySpecialKotlinFunction` we need to provide the resulting type of the expression in the signature. We can not do this, because `old` has a generic return type.

### DIFF
--- a/formver.compiler-plugin/testData/diagnostics/verification/old.kt
+++ b/formver.compiler-plugin/testData/diagnostics/verification/old.kt
@@ -15,6 +15,7 @@ fun <!VIPER_TEXT!>test<!>(@Unique @Borrowed c: C) {
         c.field == 43
     }
     inc(c)
+    verify(c == old(c))
 }
 
 @NeverConvert
--- a/formver.compiler-plugin/testData/diagnostics/verification/old.kt
+++ b/formver.compiler-plugin/testData/diagnostics/verification/old.kt
@@ -18,6 +18,7 @@ fun <!VIPER_TEXT!>test<!>(@Unique @Borrowed c: C) {
     verify(c == old(c))
 }
 
+// TODO: Remove the @NeverConvert once we have uniqueness information.
 @NeverConvert
 fun inc(@Unique @Borrowed c: C) {
     postconditions<Unit> {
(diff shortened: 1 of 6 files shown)


jesyspa:
Let's see what we settle on with the forAll PR: my gut feeling is that this should still go under partially special kotlin function and we should just refactor that, but maybe it's more complicated than I imagine.

The-Ray-Man:
> Let's see what we settle on with the forAll PR: my gut feeling is that this should still go under partially special kotlin function and we should just refactor that, but maybe it's more complicated than I imagine.

Actually It is even simpler than that. We can treat it as a `FullyspecialKotlinFunction`. My comment above is wrong, because when we embed the call to `old`, we add the old ExpEmbedding which has the correct type embedding. So the wrong function type is never used in a "type-checking-context" but just for the name. So having it as nullableAny works fine (as long as we embed generic types as nullableAny)

formver.compiler-plugin/testData/diagnostics/verification/old.kt
```diff
@@ -0,0 +1,27 @@
...
+    }
+    postconditions<Unit> {
+        c.field == 43
+    }
+    inc(c)
+    verify(c == old(c))
+}
+
+@NeverConvert
+fun inc(@Unique @Borrowed c: C) {
```
jesyspa, line 23: This is quite a strange test: it is never convert, but also has a postcondition.  Why?

jesyspa (changes requested):

formver.compiler-plugin/testData/diagnostics/verification/old.kt
The-Ray-Man: This is a workaround to test the old thing without having uniqueness information. At the moment we assume that every object is shared, so if I would convert this method then the ` c.field = c.field + 1` would be havoc'ed (just vanish) and the postcondition would not verify. 
So by having the `inc` as a helper function and calling it from the `test` function, I can test in the postcondition of `test` that the old has the effect it should, without every updating a field (which would require uniqueness information)

jesyspa: Please add a TODO for removing the `@NeverConvert` in that case

jesyspa (approved):

PR #180, opened 2026-05-29
Allow `forAll` expression as `verify` parameter

The-Ray-Man:
This PR allows user to write `verify(forAll<Int> { i -> i == i })` (and similar).

### Note
I had to add a second `forAll` method to the builtins. I don't like this, but I don't see another way to do it. It would be possible to only have one forAll but then we need to make sure that verify also runs an `InvariantBuilder`. But then the user would need to write 
```
verify {
  forAll<Int> { i -> i == i}
}
```
which seems strange.

### DIFF
--- a/formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
+++ b/formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
@@ -17,23 +17,18 @@ fun verify(@Suppress("UNUSED_PARAMETER") vararg predicates: Boolean) = Unit
 
 infix fun Boolean.implies(other: Boolean) = !this || other
 
-fun loopInvariants(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.() -> Unit) = Unit
+fun loopInvariants(@Suppress("UNUSED_PARAMETER") body: () -> Unit) = Unit
 
-fun preconditions(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.() -> Unit) = Unit
+fun preconditions(@Suppress("UNUSED_PARAMETER") body: () -> Unit) = Unit
 
-fun <T> postconditions(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit) = Unit
+fun <T> postconditions(@Suppress("UNUSED_PARAMETER") body: (T) -> Unit) = Unit
 
 
-fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: (T) -> Unit): Boolean =
+fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit): Boolean =
     throw FormverFunctionCalledInRuntimeException("forAll")
 
-/**
- * This class is designed as a receiver for lambda blocks of `loopInvariants`, `preconditions` and `postconditions`.
- */
-class InvariantBuilder {
-    fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: (T) -> Unit): Boolean =
-        throw FormverFunctionCalledInRuntimeException("forAll")
 
+class InvariantBuilder {
     /**
      * Specifies trigger expressions for quantifiers.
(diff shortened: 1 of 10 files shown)


formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
```diff
@@ -23,6 +23,10 @@ fun preconditions(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.() -> Uni
 
 fun <T> postconditions(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit) = Unit
 
+
+fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: (T) -> Unit): Boolean =
```
jesyspa: I see two ways of avoiding this:
- moving `forAll` out of the `InvariantBuilder` (it's not like it's needed, it was just to make forAll easier to find + harder to misuse)
- adding a version of `verify` that takes a function object

I think the second makes a lot of sense, really, I don't think the curly braces look unnatural and it makes it makes it clear the code shouldn't execute at runtime.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -327,6 +322,25 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
                 )
                 return data.insertForAllFunctionCall(forAllArg.symbol, forAllBody)
             }
+
+            symbol.isFormverFunctionNamed("verify") -> {
+                val callee = data.embedAnyFunction(symbol)
+                val arguments = data.withNoScope {
+                    functionCall.functionCallArguments.withVarargsHandled(this, callee)
+
```
jesyspa: nit: extra line

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -305,17 +306,11 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
-                    data,
-                    data.embedType(functionCall.resolvedType),
-                )
-            }
+        val forAllLambda =
+            functionCall.extractFormverFirBlock { isInvariantBuilderFunctionNamed("forAll") || isFormverFunctionNamed("forAll") }
+        when {
 
-            else -> {
+            forAllLambda != null -> {
```
jesyspa: I'm not super happy about this being a special case here, and this change adds another special case below.  Would it be possible to avoid this?

jesyspa (changes requested):
More suggestions than necessities, but I think this could be a bit cleaner ^^

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
The-Ray-Man: Currently I have implemented option 1. 
I think a very confusing thing for the user would be that with approach 2 they can write `verify(a && b)`, but if they want to use a `forAll` they need to use `verify {forAll...}`. Also for `verify` we allow for `varargs` which would not mix with the `forAll`.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
The-Ray-Man: I was able to get rid off the `verify` special case distinction. At the moment I don't see how the `forAllLambda` could be resolved.
I think it is a similar situation as the `preconditions` function. But the important difference is that the output of the `forAll` should be an `ExpEmbedding`.  Maybe we could do some tricks with the partially implemented kotlin functions

jesyspa (approved):

PR #183, opened 2026-06-02
Generalize abstract type checking

### LINKED ISSUE #150: Ignore `LocalityAttribute`s placed on inferred generic types
Currently the following code reports an `INVALID_LOCALITY_TYPE_TARGET` error:
```kotlin
fun `pass local with with as global property context argument`(x: @Borrowed A) {
    <!INVALID_LOCALITY_TYPE_TARGET!>with<!>(<!LOCALITY_VIOLATION!>x<!>) {
        val y = <!LOCALITY_VIOLATION!>globalContextPropertyA<!>
    }
}
```

The cause is in in the signature of `with`:

```kotlin
public inline fun <T, R> with(receiver: T, block: T.() -> R): R
```

Since the receiver is bound to `x: local A`, the generic type `T` is inferred to be `local A` as well. This triggers the error because the locality attribute is not allowed on generic parameters.

One solution would be to ignore the attribute if it wasn't introduced manually by the user. The resolver will ignore the locality attributes placed on the generic type eitherway.

### LINKED ISSUE #164: Generalize abstract type checking
Many aspects of locality and uniqueness can be generalized to a form of abstract-type checking in which we simply check for basic specification conformance. For instance, in the locality checker we want to ensure that a `local` variable cannot be assigned to a `global`. Similarly, in the uniqueness checker we want to disallow assigning moved variables. We also need to perform type-checking of function contracts for reasoning about lambdas taking local or unique inputs.

For these reasons it would be desirable to include generic classes for performing simple type-conformance checks.

paganma:
This PR generalizes the locality-checking logic currently implemented in locality.plugin by extracting it into an abstract type-checking module in type.plugin.

The new module introduces a set of extended type checker classes that can be reused to verify type conformance in locality-sensitive contexts, such as assignments and related checks, without duplicating the underlying logic. This generalization can be used to implement the locality checking on lambda arguments, as well as uniqueness checking.

This PR also includes a small but important fix for inferred generic types: locality attributes are now ignored in those cases, which prevents a number of spurious verification failures.


formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/Locality.kt
```diff
@@ -31,3 +39,18 @@ val LocalityRenderer = Renderer<Locality> { locality ->
...
+object LocalityUnifier : TypeUnifier<Locality> {
+    override fun join(left: Locality, right: Locality): Locality {
+        return left.join(right)
+    }
+}
+
+fun Locality.meet(other: Locality): Locality =
+    maxOf(this, other)
+
+object LocalityIntersector : TypeIntersector<Locality> {
```
ligee: This seems unused. If you still need it for some reason, please add a comment with an explanation

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/SymbolTypeResolver.kt
```diff
@@ -0,0 +1,17 @@
...
+
+package org.jetbrains.kotlin.formver.type.plugin
+
+import org.jetbrains.kotlin.fir.analysis.checkers.context.CheckerContext
+import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
+
+/**
+ * Resolves the declared type of a symbol.
+ */
+fun interface SymbolTypeResolver<Type, in Symbol : FirBasedSymbol<*>> {
```
ligee: The name is confusing to my mind. It is difficult to think about "Locality" as a "Type". Maybe some other naming is needed.

ligee (changes requested):
Architecturally it looks good, although a bit overdesigned in the absence of the "Type" other than "Locality". 
But the use of the word "Type" is not justified to my mind. I suggest to think about some other naming. We can discuss it, if you like.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/Locality.kt
paganma: I added in comment that these operations will be used later for locality-contract checking.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/SymbolTypeResolver.kt
paganma: In the end I went with `TypeFact` instead of the generic `Type`. While the actual intent of these classes is to compute values like a type-resolver would, I do agree that `Type` seems to be a bit too general. I also tried:

- `Fact`: This is actually the most concise option but may be a bit too general. The good thing is that it references "facts" as in "analysis facts", but I'm not sure whether this would make sense for something that can be resolved from an expression, rather than a CFG. This said, a `Fact` can also be associated with `join` and `intersect` operations, so it may actually be a good match.
- `Info`: This is the same as Fact but maybe even more general.
- `TypeQualifier`: This is a bit too long and doesn't play nicely when there is a QualifiedAccess in the same name.

Overall `TypeFact` works fine with the only con that it's much more verbose than Type. Besides `TypeFact` the other name I would consider is just "Fact". In the uniqueness checker we use these generic classes to infer not only qualifier-like values but also more complex values such as `AccessState`, a trie data structure that keeps track of the variable-field accesses in an expression.

paganma:
> Architecturally it looks good, although a bit overdesigned in the absence of the "Type" other than "Locality". But the use of the word "Type" is not justified to my mind. I suggest to think about some other naming. We can discuss it, if you like.

Although I do agree that for only one Locality "Type" this may seem overkill, these classes have been proved quite useful for implementing locality-contracts and most of the uniqueness checker. About the name I added one final small discussion as a reply to your second comment. Let me know whether you have any opinion.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/SymbolTypeResolver.kt
ligee: Let's leave it as `TypeFact` now.

ligee (approved):

PR #186, opened 2026-06-12
Add locality-contract checking

paganma:
This PR adds checking for locality annotations (currently `@Borrowed`) on function-typed values. Whenever a function value flows into a position with an expected function type (e.g. assignments, returns, etc.) the locality contracts of the two function types must agree. For example:
```kotlin
val g: (Any) -> Unit
fun h(f: (@Borrowed Any) -> Unit) { ... }
h(g) // LOCALITY_CONTRACT_MISMATCH
```

Two new diagnostics are reported:
  - `LOCALITY_CONTRACT_MISMATCH` for assignments, returns, arguments, receivers, and parameter defaults;
  - `CONTEXT_LOCALITY_CONTRACT_MISMATCH` for context parameters.

The check is implemented by reusing the generic `*TypeFactChecker` classes from `type.plugin`. Additionally:
  - `TypeRefLocalityAttributeChecker` is reworked so locality attributes are validated in the correct position within function types;

New test data lives under `testData/diagnostics/locality_checker/contract/` and covers assignments, plain and anonymous calls, context receivers, varargs, default arguments, and returns.


formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
```diff
@@ -8,34 +8,51 @@ package org.jetbrains.kotlin.formver.locality.plugin
...
-    private fun FirElement.isValidLocalityTarget(): Boolean =
-        this is FirValueParameter ||
-                this is FirReceiverParameter ||
-                this is FirProperty && isLocal ||
-                source?.kind is KtFakeSourceElementKind.ImplicitTypeArgument
-
-    private fun CheckerContext.isValidLocalityTarget(): Boolean {
-        val target = containingElements.dropLast(1).lastOrNull()
+    private fun CheckerContext.isValidLocalityTypeTarget(): Boolean {
+        val containingElements = containingElements.asReversed()
```
ligee: the variable name is confusing. It would be just better to inline it here

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
```diff
@@ -8,34 +8,51 @@ package org.jetbrains.kotlin.formver.locality.plugin
...
+        return when(target) {
+            is FirValueParameter, is FirReceiverParameter -> true
+            is FirProperty -> target.isLocal
+            else -> target.source?.kind is KtFakeSourceElementKind.ImplicitTypeArgument
+        }
+    }
 
-        return target?.isValidLocalityTarget() ?: false
+    private fun CheckerContext.isValidLocalityContractTarget(): Boolean {
+        val containingElements = containingElements.asReversed()
```
ligee: name is confusing. Please add some suffix/prefix to distinguish it from the base one.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/CallTypeFactChecker.kt
```diff
@@ -25,25 +23,24 @@ import org.jetbrains.kotlin.fir.types.resolvedType
...
+ * @param callParameterTypeFactsResolver for resolving the type-facts of the call's parameters.
  * @param contextDiagnosticFactory the diagnostic factory to use for reporting type-fact incompatibility in context
  *  parameters.
  */
 class CallTypeFactChecker<TypeFact>(
     kind: MppCheckerKind,
     private val typeFactJudgment: TypeFactJudgment<TypeFact>,
     private val expressionTypeFactResolver: ExpressionTypeFactResolver<TypeFact>,
-    private val valueParameterTypeFactResolver: SymbolTypeFactResolver<TypeFact, FirValueParameterSymbol>,
+    private val callParameterTypeFactsResolver: CallParameterTypeFactsResolver<TypeFact>,
```
ligee: Why this rename is important? It introduces a confusion to my mind: the entity "call parameter" is not used in the Kotlin compiler, but the "value parameter" is used.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/CallParameterTypeFactsResolver.kt
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Resolves the types of the parameters of a call.
+ *
+ * @param TypeFact the type-fact class of the parameters.
+ * @param declaredParameterTypeFactResolver the resolver for resolving the declared type of a call parameter.
+ * @param invokeParameterTypeFactsResolver the resolver for resolving the types of the parameters of an invoke call.
+ */
+class CallParameterTypeFactsResolver<TypeFact>(
```
ligee: Same comment as around the use site: in the Kotlin compiler the notion "value parameter" is used.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/contract/plugin/FunctionTypeFact.kt
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2010-2026 JetBrains s.r.o. and Kotlin Programming Language contributors.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the license/LICENSE.txt file.
+ */
+
+package org.jetbrains.kotlin.formver.type.contract.plugin
+
+/**
+ * Recursive contract type-fact representation based on [TypeFact].
```
ligee: Please add more comment on how TypeContractFact differs from the TypeFact.
I maybe have some understanding now, but it took me some time to figure out.
Also if my understanding is correct, this could be a name-conflict with the entity "contract" as it is used in the compiler. So it may make sense to think about naming again.

formver.compiler-plugin/testData/diagnostics/locality_checker/invalid.kt
```diff
@@ -19,8 +19,6 @@ class `reject locality attribute on member property type` {
     val x: <!INVALID_LOCALITY_TYPE_TARGET!>@Borrowed Any<!> = Any()
 }
 
-// Note: the assigned `listOf` expression flags an error because the locality attribute is implicitly added on the
-// inferred generic parameter.
 val `reject locality attribute on generic type parameter`: List<<!INVALID_LOCALITY_TYPE_TARGET!>@Borrowed Any<!>>
```
ligee, line 22: Why the removed comment is not needed anymore?

ligee (changes requested):

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/CallTypeFactChecker.kt
paganma: The rationale for changing the name is that `callParameterTypeFactsResolver` doesn't resolve the type-fact of a value parameter, but it resolves the type-fact for every arguments of a call. This is necessary because sometimes we may not be able to extract the type-fact from just the value parameter, for example when calling a lambda through `invoke`. In those cases the type-fact will not be resolved on the value-parameter but from the contract instead. I'm currently looking for what name the compiler uses for types inferred from the lambda type instead of from the value parameter.

formver.compiler-plugin/testData/diagnostics/locality_checker/invalid.kt
paganma: This is something I forgot to remove from the previous [PR #183](https://github.com/JetBrains/SnaKt/pull/183). This PR added a fix that disables inferring the locality annotation within a generic parameter, which was causing errors like the one I was addressing with this comment. That PR fixed this behavior, hence this comment is no longer necessary.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/CallParameterTypeFactsResolver.kt
paganma: After thinking about it would it be acceptable to rename this class (and corresponding locals) to ` CallArgumentToTypeFactMapper`. We can also change the name of the method to `mapArgumentsOf`. I think this makes the scope of the class clearer.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/contract/plugin/FunctionTypeFact.kt
paganma: I think the closest thing we could rename this as is `FunctionTypeFact`. I think I originally didn't use this because it's a little bit longer but I think it may be clearer and can probably avoid the collisions to the `*Contract*` classes in the compiler.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
paganma: I take it that this comment refers to `target`, right?

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
paganma: I guess that `target0` should be `currentElement` and `target1` should be `containerElement`. I think that `target2` in the first `when` branch should always be a `FirFunctionTypeRef`, so I may call it directly `functionTypeRef` and replace the instance check with an `error` case if the element cannot be converted to a `functionTypeRef`.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
ligee: No, I meant `containigElements`

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
ligee: Here I'm talking about `containingElements` too. `targets0`, etc. are not ideal, but not confusing either.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/TypeRefLocalityAttributeChecker.kt
ligee: You're shadowing the original `containigElements` with the one with reversed order, this is the source of confusion here.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/contract/plugin/FunctionTypeFact.kt
ligee: Yes, that would be much clearer.

ligee (approved):

PR #187, opened 2026-06-23
Ramon/int array as seq

The-Ray-Man:


jesyspa:
Closing in favour of #191, which pursues the newer domain-model approach to int arrays.

PR #188, opened 2026-06-24
Added Manual Folds + Unfolds

The-Ray-Man:
Comment: I needed the `Fold` and `Unfold` `ExpEmbeddings` for the `intArray` stuff, so I thought I could also add the manual folds and unfolds to it. I am not a 100% sure, if we want to keep this.

This PR introduces manual permission management focused on folding and unfolding predicates. While this will eventually be rendered obsolete by the uniqueness system, having the option to manage permissions manually in the meantime is a valuable addition.

## Core Changes

* **Class-Level `@Manual` Annotation:** The `@Manual` annotation can now be applied to an entire class. This specifies that all folding and unfolding for this type must be handled manually. Specifically:
  * Predicates are still added to pre- and post-conditions as usual (based on uniqueness/locality annotations).
  * The `Unique` predicate is still created.
  * All paths not annotated as manual continue to be handled as they were previously.
  * Fields of a manual class are never havoced. (this only applies to the current type, if a superclass is annotated as manual but the actual type is not, then it is still havoced).

## Syntax

```kotlin
val x: A = A()
unfold(UniquePred(x))
```

This will unfold the unique predicate of type `A` for `x`. Because a supertype's predicate is part of the subtype's predicate, you may also need to unfold the supertype predicate. This can be achieved as follows:

```kotlin
unfold(UniquePred(x as B))
```

Note on Generics & Naming:  Specifying the type via casting is currently required. While using a generic type parameter inside UniquePred would be much cleaner, implementing it that way right now would be highly hacky due to our current lack of support for generics.

Additionally, the predicate is named `UniquePred` rather than `Unique` to avoid conflicts with the existing `@Unique` annotation.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
@@ -130,19 +130,28 @@ object SpecialKotlinFunctions {
 
         withCallableType(booleanBooleanToBooleanType) {
             addFunction(
-                booleanBooleanToBooleanType, SpecialPackages.kotlin, className = "Boolean", name = "and"
+                booleanBooleanToBooleanType,
+                SpecialPackages.kotlin,
+                className = "Boolean",
+                name = "and"
             ) { args, _ ->
                 And(args[0], args[1])
             }
 
             addFunction(
-                booleanBooleanToBooleanType, SpecialPackages.kotlin, className = "Boolean", name = "or"
+                booleanBooleanToBooleanType,
+                SpecialPackages.kotlin,
+                className = "Boolean",
+                name = "or"
             ) { args, _ ->
                 Or(args[0], args[1])
             }
 
             addFunction(
-                booleanBooleanToBooleanType, SpecialPackages.kotlin, className = "Boolean", name = "xor"
+                booleanBooleanToBooleanType,
+                SpecialPackages.kotlin,
+                className = "Boolean",
+                name = "xor"
             ) { args, _ ->
                 Xor(args[0], args[1])
             }
@@ -206,11 +215,13 @@ object SpecialKotlinFunctions {
         addFunction(forAllCallableType, SpecialPackages.formver, name = "forAll") { args, ctx ->
             val arg = args.first()
             val lambda = arg.ignoringMetaNodes() as? LambdaExp ?: throw SnaktInternalException(
-                null, "First argument of forAll function must be a lambda."
+                null,
+                "First argument of forAll function must be a lambda."
             )
             val param = lambda.function.valueParameters.first()
             val body = lambda.function.body ?: throw SnaktInternalException(
-                null, "Lambda body of forAll function must be present."
+                null,
+                "Lambda body of forAll function must be present."
             )
             ctx.insertForAllFunctionCall(param.symbol, body)
         }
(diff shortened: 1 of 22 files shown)


formver.compiler-plugin/testData/diagnostics/verification/manualFolding.viper.diag.txt
jesyspa, line 1: Should this be here?

formver.compiler-plugin/testData/diagnostics/verification/classes/acc_precondition.fir.diag.txt
```diff
@@ -1,6 +1,10 @@
 /acc_precondition.kt: info: Generated Viper text for test_acc_precondition:
 field a: Ref
 
+predicate Permission_unique(this: Ref) {
```
jesyspa, line 4: This is a bit noisy, can we easily avoid it somehow?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -201,47 +206,88 @@ object SpecialKotlinFunctions {
...
+                is PermissionLit -> perm
+                else -> throw SnaktInternalException(
+                    null, "Second argument of `acc` must be `read()` or `write()`."
+                )
+            }
+        }
+
+        fun extractPredicate(exp: MethodCall, permissions: ExpEmbedding?): PredicateAccessPermissions {
+
+            val permissions = extractPermission(permissions)
```
jesyspa: Why plural?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -201,47 +206,88 @@ object SpecialKotlinFunctions {
...
+
+            return when (exp.type.pretype.name) {
+                uniquePredTypeName -> {
+                    val arg = exp.args.firstOrNull()!!
```
jesyspa: I think there's a `first` that is like firstOrNull()!! but a bit more descriptive

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -316,13 +362,48 @@ object SpecialKotlinFunctions {
 
         val stringIntToCharType = buildFunctionPretype {
             withDispatchReceiver { string() }
-            withParam { int() }
+            withParam {
+
```
jesyspa: nit: extra line

jesyspa (approved):
Nice, looks good!  Some minor comments

formver.compiler-plugin/testData/diagnostics/verification/manualFolding.viper.diag.txt
The-Ray-Man: Nope, good catch. I think in general we should not allow to have empty files in the test data directory.

formver.compiler-plugin/testData/diagnostics/verification/classes/acc_precondition.fir.diag.txt
The-Ray-Man: Hmm, I think we kind of have this problem for all our builtin things. I will look into it

TheDeveloper101 (commented):
looks good

PR #190, opened 2026-06-25
Resolved Pure Function Permission Issue

The-Ray-Man:
This PR fixes a bug introduced by #169 
In a pure function if a parameter was annotated as unique and borrowed in the postcondition the unique predicate was added. Viper does not allow to have permissions in the postcondition of pure functions.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/ContractBuilder.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/ContractBuilder.kt
@@ -58,7 +58,7 @@ class FunctionConditionBuilder(
             args {
                 accessInvariants()
                 pureInvariants()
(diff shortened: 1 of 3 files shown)


jesyspa (approved):

jesyspa:
I think it would be better to remove it from the postcondition, but this is okay as a temporary fix; maybe add a TODO?

The-Ray-Man:
> I think it would be better to remove it from the postcondition

I don't fully understand. Are you suggesting that we add the predicates anyway and remove it afterwards?

jesyspa:
The semantics of access permissions are different in methods and functions.  In methods, passing a permission consumes it, so you need it in the postcondition in order to have access to it.  In functions, permissions are only ever borrowed, so you can put acc(a.x, write) in the precondition and then you get it back after the function is executed.

jesyspa:
Ahh, sorry, I think I see where I was misunderstanding; I took it to remove the conditions fully (pre and post).  This code looks correct.

PR #193, opened 2026-06-26
Refactor Stdlib Matching

The-Ray-Man:
In my opinion the way we matched a function to stdlib functions to add various pre+post conditions was more complicated then it needs to be and it did not allow us to match for constructors.

This PR cleans it up and supports matching constructors.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
@@ -30,56 +30,56 @@ private fun VariableEmbedding.increasedSize(amount: Int): ExpEmbedding = EqCmp(
 /**
  * Matches a stdlib function against a set of AND-combined criteria.
  *
- * [pkg] + [receiverInherits]: the dispatch receiver must be a subtype of `pkg.receiverInherits`.
+ * [pkg] + [receiverSubtypeName]: the dispatch receiver must be a subtype of `pkg.receiverInherits`.
  * [pkg] is not re-checked against the function's own package in this case — overrides in
  * user-defined classes match just as well as the original declarations.
  *
  * [pkg] + [noReceiver] / [functionName] / [constructorOf]: the function itself must live in [pkg].
  * Use this for top-level functions (e.g. `emptyList`) and constructors.
  *
- * [receiverInherits] and [noReceiver] are mutually exclusive.
+ * [receiverSubtypeName] and [noReceiver] are mutually exclusive.
  * [functionName] and [constructorOf] are mutually exclusive.
  */
-data class StdLibFunctionSpec(
+class StdLibFunctionSpec(
     val pkg: List<String>? = null,
-    val receiverInherits: String? = null,
+    val receiverSubtypeName: String? = null,
     val noReceiver: Boolean = false,
     val functionName: String? = null,
     val constructorOf: String? = null,
 ) {
     init {
-        require(!(receiverInherits != null && noReceiver)) { "receiverInherits and noReceiver are mutually exclusive" }
-        require(receiverInherits == null || pkg != null) { "pkg must be set when receiverInherits is set" }
+        require(!(receiverSubtypeName != null && noReceiver)) { "receiverInherits and noReceiver are mutually exclusive" }
+        require(receiverSubtypeName == null || pkg != null) { "pkg must be set when receiverInherits is set" }
         require(constructorOf == null || functionName == null) { "constructorOf and functionName are mutually exclusive" }
     }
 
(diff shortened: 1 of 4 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
```diff
@@ -14,71 +14,109 @@ import org.jetbrains.kotlin.formver.core.embeddings.expression.OperatorExpEmbedd
...
+    val pkg: List<String>? = null,
+    val receiverInherits: String? = null,
+    val noReceiver: Boolean = false,
+    val functionName: String? = null,
+    val constructorOf: String? = null,
+) {
+    init {
+        require(!(receiverInherits != null && noReceiver)) { "receiverInherits and noReceiver are mutually exclusive" }
+        require(receiverInherits == null || pkg != null) { "pkg must be set when receiverInherits is set" }
+        require(constructorOf == null || functionName == null) { "constructorOf and functionName are mutually exclusive" }
```
jesyspa: There seem to be a lot of options here that are mutually exclusive.  Would decomposing this a bit to avoid the invalid possible states make sense?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
```diff
@@ -14,71 +14,109 @@ import org.jetbrains.kotlin.formver.core.embeddings.expression.OperatorExpEmbedd
...
+ *
+ * [pkg] + [noReceiver] / [functionName] / [constructorOf]: the function itself must live in [pkg].
+ * Use this for top-level functions (e.g. `emptyList`) and constructors.
+ *
+ * [receiverInherits] and [noReceiver] are mutually exclusive.
+ * [functionName] and [constructorOf] are mutually exclusive.
+ */
+data class StdLibFunctionSpec(
+    val pkg: List<String>? = null,
+    val receiverInherits: String? = null,
```
jesyspa: I'm not a fan of this name... `receiverSupertypeName` maybe?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
```diff
@@ -14,71 +14,109 @@ import org.jetbrains.kotlin.formver.core.embeddings.expression.OperatorExpEmbedd
...
-        }
-            ?: false
-}
+        return matchesNameScope(function.name)
+    }
 
-data object CollectionInterface : PresentInterface {
-    override val interfaceName = "Collection"
-}
+    private fun matchesReceiverInheritance(
```
jesyspa: Same here: this sounds like the receiver has an inheritance and this needs to match it.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
```diff
@@ -14,71 +14,109 @@ import org.jetbrains.kotlin.formver.core.embeddings.expression.OperatorExpEmbedd
...
-    override val interfaceName = "List"
-}
+        val receiverPretype = function.callableType.dispatchReceiverType?.pretype ?: return false
+        return typeResolver.isInheritorOf(receiverPretype, pkg!!, receiverInherits)
+    }
 
-data object MutableListInterface : PresentInterface {
-    override val interfaceName = "MutableList"
-}
+    private fun needsNameCheck(): Boolean {
```
jesyspa: I'm not sure I understand.  What does this mean?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StdLibConverter.kt
```diff
@@ -14,71 +14,109 @@ import org.jetbrains.kotlin.formver.core.embeddings.expression.OperatorExpEmbedd
...
+ * [pkg] is not re-checked against the function's own package in this case — overrides in
+ * user-defined classes match just as well as the original declarations.
+ *
+ * [pkg] + [noReceiver] / [functionName] / [constructorOf]: the function itself must live in [pkg].
+ * Use this for top-level functions (e.g. `emptyList`) and constructors.
+ *
+ * [receiverInherits] and [noReceiver] are mutually exclusive.
+ * [functionName] and [constructorOf] are mutually exclusive.
+ */
+data class StdLibFunctionSpec(
```
jesyspa: Having helpers for the common cases may be useful (and would make it harder to create an invalid one)

jesyspa (changes requested):
Generally looks good, a few comments on how it could maybe be cleaner

PR #196, opened 2026-06-27
Enable anonymous and inner functions to capture `local` variables by resolving them as `local`

paganma:
This PR addresses the current inability of an anonymous/inner function to capture `local` variables in its outer scope that are not outlived by the function. 

To achieve this, we introduce new locality resolution rules: if the body of an anonymous/inner function references a `local` symbol that was defined outside of its scope, then the function itself resolves to `local`.

This PR also reorganizes the locality tests according to the language feature that they exercise (e.g. `cast.kt`) or explicit examples (e.g. `connection_pool.kt`).


formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/PropertyAccessLocalityChecker.kt
```diff
@@ -9,47 +9,60 @@ import org.jetbrains.kotlin.diagnostics.DiagnosticReporter
...
+context(context: CheckerContext)
+private fun FirFunctionSymbol<*>.declares(symbol: FirBasedSymbol<*>): Boolean {
+    val graph = resolvedControlFlowGraphReference?.controlFlowGraph
+        ?: return false
+
+    return symbol in graph.resolveDeclaredSymbols()
+}
+
+private val FirFunctionSymbol<*>.supportsLocalityCapture: Boolean
+    get() = this is FirAnonymousFunctionSymbol || !isNonLocal
```
ligee: `isNonLocal` is obsolete and already removed from the compiler in master.
We need to discuss how to implement it a future-proof way.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/GraphDeclaredSymbolsResolver.kt
```diff
@@ -0,0 +1,72 @@
...
+                declaration.valueParameters.mapTo(declarations) { it.symbol }
+            }
+
+            else -> { }
+        }
+
+        graph.nodes
+            .mapNotNull { node ->
+                (node as? VariableDeclarationNode)?.fir?.symbol ?:
+                (node as? LocalFunctionDeclarationNode)?.fir?.symbol
```
ligee: Shouldn't we process local classes here too?

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/LocalityErrorMessages.kt
```diff
@@ -29,7 +29,8 @@ object LocalityErrorMessages : BaseDiagnosticRendererFactory() {
                 )
                 map.put(
                     LocalityErrors.INVALID_LOCALITY_CAPTURE,
-                    "Unable to capture outer local declaration from ''{0}''.",
+                    "Unable to capture local {0} crossing {1}.",
```
ligee: It would be nice to have somewhat clearer message here, "crossing" is too parsimonous.

ligee (changes requested):
I hope I got it now. The comments are mostly minor.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/PropertyAccessLocalityChecker.kt
paganma: Instead of using `isNonLocal` I have directly inlined the case for `CallableSymbol`.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/GraphDeclaredSymbolsResolver.kt
paganma: Added local class symbols to the list of declared symbols.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/locality/plugin/LocalityErrorMessages.kt
paganma: Switched to "across". Also for `{1}` I've implemented a custom renderer that makes the diagnostic clearer when the local is captured from anonymous declarations.

ligee (approved):

PR #207, opened 2026-07-06
Implement existential quantifier

sachok42:
Implements exists<T>  as the symmetric construct to forAll<T>. Like its counterpart, can be used in preconditions, postconditions and loop invariants


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
```diff
@@ -243,6 +243,31 @@ fun StmtConversionContext.insertForAllFunctionCall(
     }
 }
 
+/**
+ * Insert `ExistsEmbedding` where `exists` function call was encountered.
+ */
+fun StmtConversionContext.insertExistsFunctionCall(
```
jesyspa, line 249: dedup?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -226,73 +213,49 @@ object SpecialKotlinFunctions {
             ctx.insertForAllFunctionCall(param.symbol, body)
         }
 
+        addFunction(forAllCallableType, SpecialPackages.formver, name = "exists") { args, ctx ->
```
jesyspa: generalize forAllCallableType?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/expression/Quantifiers.kt
jesyspa, line 1: Maybe rename ForAllEmbedding.kt -> Quantifiers.kt and put it there?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -618,3 +629,18 @@ data class LinearizationVisitor(
...
+// Exclusive upper bound of the Unicode code-point range a Kotlin `Char` occupies.
+private const val CHAR_CODE_POINT_UPPER_BOUND = 65536
+
+/**
+ * A `Char` quantifier variable is lowered to a Viper `Int`, so without a bound it
+ * ranges over every integer including negative code points. Constrain it to the
+ * Unicode code-point range `[0, 65536)` so `forAll<Char>`/`exists<Char>` don't
+ * need a manual guard. Returns `null` for non-`Char` variables.
+ */
+private fun VariableEmbedding.charDomainBoundOrNull(): Exp? =
```
jesyspa: This is not the place to put this: it should be handled by assuming correct invariants on the type being quantified over.  That said, since the issue is preexisting (already in forAll) it's probably better done in a separate pull request.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
```diff
@@ -0,0 +1,32 @@
+// FULL_JDK
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+
+// `m` is the maximum character of `s` and actually occurs in it: the existential
+// quantifier is the dual of `forAll` and is required to state that the maximum is
+// attained. The loop invariant carries the witness forward so the postcondition holds.
+fun <!VIPER_TEXT!>maxCharacter<!>(s: String): Char {
```
jesyspa: Great example, but should be in the slow/e2e tests, here we would prefer minimal cases to test everything works.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists_char_bound.kt
jesyspa, line 1: out of scope

pipeline/meta/f6-agent-ratings.md
jesyspa, line 1: This and other files in pipeline don't belong here

formver.compiler-plugin/build.gradle.kts
jesyspa, line 1: I think this conflicts with how we solved this on main

jesyspa (changes requested):

jesyspa:
Please sync your changes + make a clear PR name and description

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -504,14 +487,42 @@ data class LinearizationVisitor(
...
+        }
+    }
+
+    override fun visitExistsEmbedding(e: ExistsEmbedding): Linearizable = object : OnlyToBuiltinLinearizable(e, this@LinearizationVisitor) {
+        override fun toViperBuiltinType(ctx: LinearizationContext): Exp {
+            val conjunction = e.conditions.pureToViper(true, ctx.typeResolver, ctx.source).toConjunction()
+            val viperTriggers = e.triggerExpressions.map { triggerExpr ->
+                Exp.Trigger(listOf(triggerExpr.linearize().toViperBuiltinType(ctx)))
+            }
+            val charBound = e.variable.charDomainBoundOrNull()
```
jesyspa: The charbound stuff doesn't belong here, though a separate PR with proper support for char would be great.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
jesyspa, line 1: I think unifying the exists files would be better; a file per test gets pretty hard to find stuff in.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
```diff
@@ -0,0 +1,11 @@
+// FULL_JDK
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+fun <!VIPER_TEXT!>simpleExists<!>(s: String): Int {
```
jesyspa: Could you check that this test fails if you remove the precondition?  If not, please add `@AlwaysVerify` (and we should fix our verification defaults...)

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists_double_index.kt
```diff
@@ -0,0 +1,15 @@
+// FULL_JDK
+
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+// Indexing the same String at two distinct variables under one `exists` reaches
+// a Silver info kind the error-reporting path does not handle. The unprovable
+// postcondition degrades to a normal Viper verification error.
```
jesyspa: This comment isn't meaningful without context; please rewrite to express what is being tested.

formver.compiler-plugin/viper/src/org/jetbrains/kotlin/formver/viper/ast/Info.kt
```diff
@@ -22,7 +22,11 @@ sealed class Info : IntoSilver<viper.silver.ast.Info> {
         fun fromSilver(info: viper.silver.ast.Info): Info = when (info) {
             `NoInfo$`.`MODULE$` -> NoInfo
             is Wrapper -> Wrapped(info.wrappedValue)
-            else -> TODO("Unreachable")
+            // Silicon may attach its own info kinds (e.g. auto-generated trigger
+            // metadata) to the nodes it reports errors on. Those carry no SnaKt
+            // source role, so treat them as absent rather than crashing: this lets
+            // the error surface as a normal Viper verification error.
+            else -> NoInfo
```
jesyspa, line 29: Please verify that this is what happens.

.gitignore
jesyspa, line 1: Change doesn't belong in this repository.

jesyspa (changes requested):
Please update PR name + description to be a meaningful description of the change.

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
```diff
@@ -28,6 +28,32 @@ fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit)
...
+ * ```kotlin
+ * exists<Int> { i -> 0 <= i && i < n && pred(i) }
+ * ```
+ *
+ * will not verify even when the body is trivially satisfiable and a trigger is present. This is
+ * confirmed at the Silicon configuration level and is not a gap in the lowering.
+ *
+ * Workaround: place the witness in a **precondition** or carry it through a **loop invariant**
+ * rather than asserting it as a postcondition. Enabling MBQI at the two `SiliconFrontend`
+ * call sites is tracked as a maintainer follow-up.
```
jesyspa, line 51: While this is very interesting, I don't think it belongs here.  The question should be: is this useful for a user?

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
```diff
@@ -28,6 +28,32 @@ fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit)
     throw FormverFunctionCalledInRuntimeException("forAll")
 
 
+/**
+ * Asserts that there exists a value of type [T] satisfying the predicate in [body].
+ *
+ * Lowered to a Viper `exists` quantifier. The [body] runs in an [InvariantBuilder] receiver scope,
+ * allowing [InvariantBuilder.triggers] to supply SMT trigger hints.
```
jesyspa, line 35: Probably not relevant to the user, an the parts that are should be phrased for the user's benefit and not duplicate the code.

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/ViperPoweredDeclarationChecker.kt
```diff
@@ -109,6 +109,9 @@ class ViperPoweredDeclarationChecker(private val session: FirSession, private va
                     reporter.reportVerifierError(source, err, config.errorStyle)
                 }
 
+                // TODO: emptyList() leaves z3config.smt2's `smt.mbqi false` in effect, which
+                //  prevents exists<T> from discharging postcondition existential witnesses. Pass
+                //  Silicon args that enable MBQI here once that change is scoped and approved.
                 val verifier = SiliconFrontend(emptyList())
```
jesyspa, line 115: Again, probably not something that should be included this way.  If you want to illustrate an issue, a test would do better.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists_list_get_crash.kt
```diff
@@ -0,0 +1,18 @@
...
+
+// `List.get` inside a quantifier body goes through a method-call (non-pure) embedding path,
+// unlike `String`'s indexing operator. The quantifier lowering's fresh-anonymous-variable
+// machinery assumes a pure context, so this currently crashes the compiler instead of producing
+// a verification diagnostic. Confirmed to reproduce identically with `forAll<Int>` in place of
+// `exists<Int>` here, so this is a pre-existing defect in quantifier lowering generally, not
+// something introduced by `exists<T>`; pinning it via `exists` since that is this PR's scope.
+// A real fix needs the linearizer to handle a non-pure quantifier body, which is more than a
+// minimal patch, so this is landed as a documented crash rather than fixed here.
+<!INTERNAL_ERROR!>fun existsListGetCrash(l: List<Int>, res: Int): Int {
```
jesyspa, line 13: Testing for an internal error seems pretty strange.  Maybe just file an issue/have an agent fix it?

jesyspa (changes requested):

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
```diff
@@ -28,6 +28,16 @@ fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit)
     throw FormverFunctionCalledInRuntimeException("forAll")
 
 
+/**
+ * Asserts that there exists a value of type [T] satisfying the predicate in [body].
+ *
+ * **Known limitation:** `exists` in a postcondition will not verify even when the predicate
+ * is trivially satisfiable. Place the witness in a precondition or loop invariant instead.
```
jesyspa: For my understanding, why is this?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -32,6 +31,18 @@ data class LinearizationVisitor(
      */
     private fun ExpEmbedding.linearize(): Linearizable = accept(this@LinearizationVisitor)
 
+    private fun quantifierParts(
```
jesyspa: A function should generally be given a verb as its name.  What is the verb here?

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
```diff
@@ -28,6 +28,24 @@ fun <T> forAll(@Suppress("UNUSED_PARAMETER") body: InvariantBuilder.(T) -> Unit)
...
+ * in a postcondition must be *proven*: it verifies when the solver can find a witness. This
+ * succeeds when the body contains a trigger term the solver can match on (e.g. a `s[i]` access)
+ * and the witness is established in context — for example carried forward by a loop invariant of
+ * the same shape (see `max_character.kt` in the `expensive_verification` tests).
+ *
+ * It may fail for an existential over bare arithmetic with no such trigger (e.g. `exists { it == 0 }`),
+ * because finding that witness needs model-based quantifier instantiation, which is not enabled.
+ * Such a failure is reported as a verification warning ("might not hold"); note this signals only
+ * that the witness could not be found, not that the existential is false.
+ */
```
jesyspa, line 33: This kind of comment is not something we should merge, for a number of reasons:
- The second paragraph is almost entirely a repeat of things that are not about this function.
  - The developer should know that preconditions are assumed and postconditions proven (in the context of one function; of course, the picture is reversed when calling the function, which makes the whole aside even more pointless).
  - The second sentence does not give enough detail to understand the material, it only makes sense if the reader already knows how triggers work, in which case it is useless.
  - It points to examples that may drift and that don't actually help.
- The third paragraph is similarly problematic.
  - Whether MBQI is enabled isn't determined here, and hard-coding it in a comment like this is unlikely to be the right call.
  - How the error is reported is not something this comment should care about.
  - That verification failing isn't the same as a counterexample is again something that the user should know (or at least, this isn't the place to tell them about it).

The reason for LLMs making that kind of error is that they still have quite poor theory of mind, so they don't understand what the reader does/doesn't know.  This is something humans are much better at, so it's worth doing it manually.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
```diff
@@ -39,6 +40,24 @@ val SpecialKotlinFunction.callableId: CallableId
...
 fun SpecialKotlinFunction.embedName(): ScopedName = callableId.embedFunctionName(callableType)
 
+private fun insertQuantifierFunctionCall(
+    functionName: String,
+    args: List<ExpEmbedding>,
+    ctx: StmtConversionContext,
+    buildEmbedding: (VariableEmbedding, List<ExpEmbedding>, List<ExpEmbedding>) -> ExpEmbedding,
+): ExpEmbedding {
+    val lambda = args.first().ignoringMetaNodes() as? LambdaExp ?: throw SnaktInternalException(
+        null,
```
jesyspa, line 50: Why null here?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -31,6 +31,18 @@ data class LinearizationVisitor(
...
```
jesyspa: A function name should generally contain a verb to indicate what it does.  What is the correct verb for this function?

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
```diff
@@ -510,6 +519,23 @@ data class LinearizationVisitor(
...
+        override fun toViperBuiltinType(ctx: LinearizationContext): Exp {
+            val (conjunction, viperTriggers) = quantifierParts(e.conditions, e.triggerExpressions, ctx)
+            return Exp.Exists(
+                variables = listOf(e.variable.toLocalVarDecl()),
+                triggers = viperTriggers,
+                exp = if (e.variable.isOriginallyRef) Exp.And(
+                    e.variable.toViperExp(ctx).isOf(e.variable.type.runtimeType),
+                    conjunction
+                )
+                else conjunction,
```
jesyspa, line 532: Perhaps this should be a helper, given there's duplication with the above?

formver.compiler-plugin/testData/diagnostics/expensive_verification/algorithms/max_character.kt
```diff
@@ -0,0 +1,32 @@
+// FULL_JDK
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+
+// `m` is the maximum character of `s` and actually occurs in it: the existential
+// quantifier is the dual of `forAll` and is required to state that the maximum is
+// attained. The loop invariant carries the witness forward so the postcondition holds.
```
jesyspa, line 8: Is all the detail here worthwhile?

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
```diff
@@ -0,0 +1,41 @@
+// FULL_JDK
+
+import org.jetbrains.kotlin.formver.plugin.*
+
+// Green path: an `exists` in a *precondition* is assumed at method entry, so it
+// verifies without the solver having to construct a witness. (An `exists` in a
+// postcondition must instead be proven; that verifies when a witness is reachable
+// — see max_character.kt in the expensive_verification tests.)
```
jesyspa: Please apply similar guidance to what I said before to comments here.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.fir.diag.txt
```diff
@@ -0,0 +1,32 @@
+/exists.kt: info: Generated Viper text for simpleExists:
+method simpleExists() returns (v_ret_0: Ref)
+  requires (exists anon: Int :: anon == 0)
```
jesyspa, line 3: I find it a bit surprising here that we lower it this way, rather than `exists anon: Ref :: intFromRef(anon) == 0`, interesting.  That actually explains what the agent was talking about.  Interesting question.

jesyspa (changes requested):

formver.annotations/src/org/jetbrains/kotlin/formver/plugin/Builtins.kt
sachok42: Removed most of the comment, left only the basic definition part, as I think that an unfamiliar developers may fail to understand by themselves that exists means ∃

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/embeddings/callables/FullySpecialKotlinFunction.kt
sachok42: It replicates the way it's done in other places around this (and possibly other) files. The source of the exception, which is what the first argument stands for, is lost in a few other places as well. Seems like a thing to fix in a separate PR later on

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
sachok42: renamed locally to getQuantifierParts, for some reason git doesn't let me push right now. Will be fixed with the next push

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/linearization/LinearizationVisitor.kt
sachok42: I'd say it's not worth is, as these two contain some pieces of code are quite short and contain some important differences and while they can be represented with one helper function, it would be a clunky and confusing one, which, in my opinion, would hurt readability

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+    val c = s[0]
+    var i = 0
+    while (i < s.length) {
+        loopInvariants {
+            0 <= i && i <= s.length
+            exists<Int> { 0 <= it && it < s.length && s[it] == c }
+        }
+        i += 1
+    }
```
jesyspa, line 34: What does the presence of the loop contribute to the test?

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
```diff
@@ -0,0 +1,63 @@
...
+        exists<Int> { 0 <= it && it < s.length && s[it] == s[res] }
+    }
+    return 0
+}<!>
+
+// Incompleteness path: this existential (`it == 0`) is plainly true, but its body is bare
+// arithmetic with no trigger term for the solver to match on. Without model-based quantifier
+// instantiation the witness cannot be constructed, so the postcondition is reported as a
+// verification warning. This pins "witness not found" — not "existential is false" — and is the
+// counterpart to max_character.kt, where a `s[i]` trigger plus a loop invariant let it verify.
```
jesyspa, line 56: Please remove reference to other file; this is just going to get lost over time, and it's not a helpful comparison anyway.

jesyspa (approved):
Two minor points, then should be good to merge.

formver.compiler-plugin/testData/diagnostics/verification/user_invariants/exists.kt
sachok42: the loop provides the test with triggers

PR #208, opened 2026-07-07
Integrate uniqueness checker within compiler infrastructure

paganma:
This PR integrates the uniqueness checker as a standalone FIR compiler-plugin extension and replaces the old experimental uniqueness checker with a checker stack that raises normal compiler diagnostics.

The change adds:

- `@Unique` as a type-position attribute (much like in the locality-checker).
- A uniqueness FIR extension registrar, expression resolvers, CFG state resolver, additional checkers, diagnostics, and renderers.
- Forward-flow uniqueness state analysis over FIR control-flow graphs.
- Uniqueness type-fact checkers and more specialized checkers for checking uniqueness consistency.
- More tests focusing aliasing, calls, constructors, context parameters, control flow, nullable/safe calls, operators, receivers, returns, throws, loops, and `when`/`try` expressions.

## Uniqueness Analysis

The analysis tracks the uniqueness state for values and corresponding subpaths through the FIR CFG. The current uniqueness domain is:

- `Unique`: the value/path is uniquely owned.
- `Unknown`: a local/borrowed value whose uniqueness cannot be assumed.
- `Shared`: the value/path is shared.
- `Moved`: a unique or unknown value/path has been consumed and can no longer be accessed.

The analysis is a forward dataflow analysis over FIR CFG nodes. It initializes function parameters and receiver parameters from their declared uniqueness, propagates a `UniquenessState` through transitions that depend on the individual nodes. Example of such transitions:

- Variable declarations and assignments transplant the right-hand side uniqueness substate onto the left-hand path, initialize the assigned path to its declared uniqueness, and move the right-hand access paths.
- Function-call entry moves explicit receivers, context receivers, and arguments because those values may escape to the callee.
- Function-call exit re-initializes arguments and receivers whose required locality is local, modeling borrowed/local calls that do not consume their inputs.
- Return and throw expressions move the returned/thrown access paths.
- Merged control-flow paths conservatively join per-path uniqueness states, so any path that may have been moved remains visible to later diagnostics.

## Analysis and checking models

The implementation uses two models based on immutable tries over FIR symbols:

- `AccessState = PathTrie<Access>` records the paths touched by an expression. A terminal node represents a concrete access such as `box.item`, and terminal nodes may also have children when both a path and one of its subpaths are accessed.
- `UniquenessState = PathTrie<Uniqueness>` records the current uniqueness value for each tracked path prefix. Missing nodes are materialized lazily from the declared uniqueness of the symbol, joined with the parent state.

Expression resolvers build these models from FIR:

- `ExpressionAccessStateResolver` extracts access paths from qualified accesses, receivers, safe calls, and composed expressions.
- `ExpressionUniquenessResolver` computes the declared/terminal uniqueness of expressions, using `@Unique`, locality attributes, constructors, primitive/Nothing values, receivers, property accesses, and function result types.
- `GraphUniquenessStatesResolver` caches CFG analysis results and exposes input/output uniqueness states to the function-level checkers.

The `@Unique` annotation is represented as a FIR type attribute through `UniquenessAttributeExtension`, and `TypeRefUniquenessAttributeChecker` restricts it to supported declaration, which currently include: function parameters, receiver parameters, properties (both local and fields), and function return types. Note that uniqueness contracts haven't been implemented yet.

## Uniqueness checking

The checker stack turns the analysis state and type facts into FIR diagnostics:

- Assignment, call, qualified-access, return, throw, property, and type-ref checkers report uniqueness mismatches where a site requires a stronger uniqueness than the expression provides.
- `FunctionMovedAccessChecker` reports invalid reads of paths whose flow state is `Moved`.
- `FunctionEscapeUniquenessConsistencyChecker` reports values or context parameters that escape while containing moved unique fields.
- `FunctionExitUniquenessConsistencyChecker` reports borrowed/local function values that still contain moved fields at function exit.
- `ExpressionArgumentUniquenessCollisionChecker` rejects passing the same unique access twice, or passing overlapping unique paths, including through receivers and context arguments.

The diagnostics now distinguish mismatched uniqueness, invalid moved accesses, escaping moved fields, invalid duplicate unique arguments, overlapping unique arguments, and invalid `@Unique` targets.


formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/Uniqueness.kt
```diff
@@ -0,0 +1,68 @@
...
+    Unique,
+    Unknown,
+    Shared,
+    Moved
+}
+
+/**
+ * Resolves the uniqueness assumed for values of this type when no parameter-specific rule applies.
+ */
+val ConeKotlinType.defaultUniqueness: Uniqueness
```
ligee: Is there a reason it is called the "default" uniquiness? Seems misleading to me.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessAttribute.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.jetbrains.kotlin.fir.types.ConeAttributes
+import kotlin.reflect.KClass
+
+data object UniquenessAttribute : ConeAttribute<UniquenessAttribute>() {
+    override fun union(other: UniquenessAttribute?): UniquenessAttribute =
+        this
+
+    override fun intersect(other: UniquenessAttribute?): UniquenessAttribute? =
+        other
+
```
ligee, line 27: I'm not sure it is correct. Something like this should be checked, to make sure such arithmetic is not making it unsound:
```
fun f(val a: A, @Unique val b: B) {
    if (a is B) {
       // here a will have an intersection type, those attributes are intersected as well
...
```

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
```diff
@@ -0,0 +1,188 @@
...
+ *     \__b*
+ *     \  \__c*
+ *     \__d*
+ *
+ *   t1:                paths(t1) = { [f], [g] }
+ *     \__f*
+ *     \__g*
+ *
+ *   t0.append(t1):     paths = { [b, f], [b, g], [b, c, f], [b, c, g], [d, f], [d, g] }
+ *     \__b
```
ligee, line 49: Do I understand it correctly, that the `Access` is not changed by this op, so `b` and `c` remain being `Terminal` in the results? 
If so, the diagram appears slightly incorrect.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
```diff
@@ -0,0 +1,188 @@
...
+ *     \  \__g*
+ *     \  \__c
+ *     \     \__f*
+ *     \     \__g*
+ *     \__d
+ *        \__f*
+ *        \__g*
+ * ```
+ */
+fun AccessState.append(other: AccessState): AccessState {
```
ligee, line 64: The only use site I see performs much simpler version of the algorithm, adding a sigle child to every path. Maybe it makes sense to have less generic but simpler `appendLeaf` or something, instead.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/UnifyingExpressionTypeFactResolver.kt
```diff
@@ -43,6 +44,11 @@ fun FirExpression.collectTails(): Sequence<FirExpression> =
             sequence {
                 yieldAll(expression.branches.map(FirWhenBranch::result))
             }
+        is FirElvisExpression ->
```
ligee, line 51: BTW, don't we need to consider binary boolean ops the same way due to short-circuiting?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/QualifiedAccessPathReceiverResolver.kt
```diff
@@ -0,0 +1,29 @@
...
+        when(this) {
+            is FirPropertyAccessExpression -> {
+                val symbol = this.calleeReference.symbol ?: return null
+
+                if (symbol !is FirPropertySymbol) {
+                    return null
+                }
+
+                if (symbol.backingFieldSymbol != null) {
+                    explicitReceiver ?: extensionReceiver ?: dispatchReceiver
```
ligee: There could be multiple receivers, shouldn't we merge paths from all of them?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionUniquenessResolver.kt
```diff
@@ -0,0 +1,122 @@
...
+            val receiver = receiver
+            val receiverUniqueness = receiver.resolveUniqueness()
+
+            receiverUniqueness.join(resolveAccessUniqueness())
+        }
+
+        else -> {
+            val resolvedType = resolvedType
+
+            if (resolvedType.isNothingOrNullableNothing || resolvedType.isPrimitive) {
```
ligee, line 78: Is `Unique` on `Nothing(?)` could be useful in any sense?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/GraphUniquenessStatesAnalyzer.kt
```diff
@@ -0,0 +1,226 @@
...
+    override fun mergeInfo(
+        a: UniquenessStateFlow,
+        b: UniquenessStateFlow,
+        node: CFGNode<*>
+    ): UniquenessStateFlow =
+        a.merge(b) { leftState, rightState ->
+            leftState.join(rightState)
+        }
+
+    private fun UniquenessStateFlow.ensure(): UniquenessState =
```
ligee: I'd suggest to rename it to `getOrDefault` or so.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/PathTrie.kt
```diff
@@ -0,0 +1,112 @@
...
+
+import kotlinx.collections.immutable.PersistentMap
+import kotlinx.collections.immutable.persistentMapOf
+import org.jetbrains.kotlin.fir.symbols.FirBasedSymbol
+import org.jetbrains.kotlin.formver.type.plugin.TypeFactIntersector
+import org.jetbrains.kotlin.formver.type.plugin.TypeFactUnifier
+
+data class PathTrie<Type>(
+    val data: Type,
+    val children: PersistentMap<FirBasedSymbol<*>, PathTrie<Type>> = persistentMapOf(),
```
ligee, line 28: It would be nice to have some KDoc here or on the class itself, explaining, e.g., expected children relationships in the class and in the corresponding FIR.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/GraphUniquenessStatesAnalyzer.kt
```diff
@@ -0,0 +1,226 @@
...
+    ): PathAwareUniquenessStateFlow {
+        val call = node.fir
+
+        with(context) {
+            return data.transformValues { data ->
+                var newUniquenessState = data.ensure()
+
+                // NOTE: `allReceiverExpressions` also includes context arguments.
+                for (receiver in call.allReceiverExpressions) {
+                    newUniquenessState = receiver.resolveAccessState().move(newUniquenessState)
```
ligee, line 151: Do I understand it correctly, that you postponed the solution of the problem with expressions as arguments in function calls?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionArgumentUniquenessCollisionChecker.kt
```diff
@@ -0,0 +1,120 @@
...
+ * Two unique arguments collide when their access paths are identical, or when one access path is a prefix of the other.
+ * For example, passing both `x` and `x.f` to unique parameters is invalid because the accesses overlap.
+ *
+ * @param Statement the FIR expression kind handled by this checker.
+ * @param argumentUniquenessMapper the mapper for resolving required uniqueness of argument-like expressions.
+ * @param checks the predicate deciding whether this checker handles a particular expression instance.
+ */
+class ExpressionArgumentUniquenessCollisionChecker<Statement : FirStatement>(
+    private val argumentUniquenessMapper: ArgumentUniquenessesMapper<Statement>,
+    private val checks: (Statement) -> Boolean = { true }
```
ligee: `shouldBeChecked` would be clearer

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionArgumentUniquenessCollisionChecker.kt
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+            break
+        }
+    }
+
+    return commonPrefix
+}
+
+context(context: CheckerContext, reporter: DiagnosticReporter)
+private fun reportUniquenessCollision(
```
ligee: `checkUniquinessCollision` would be fairer.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionArgumentUniquenessCollisionChecker.kt
```diff
@@ -0,0 +1,120 @@
...
+    ownerElement: FirElement,
+    leftArgument: FirExpression,
+    rightArgument: FirExpression
+) {
+    for (leftPath in leftArgument.resolveAccessState().enumeratePaths()) {
+        for (rightPath in rightArgument.resolveAccessState().enumeratePaths()) {
+            val commonPrefix = leftPath.computeCommonPrefix(rightPath)
```
ligee: Adding "length" syntax to names of vars and funs would simplify understanding. (Or "size", but "lenght" looks better imo.)

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessErrorMessages.kt
```diff
@@ -0,0 +1,66 @@
...
+                    PathRenderer
+                )
+                map.put(
+                    UniquenessErrors.EXIT_UNIQUENESS_INCONSISTENCY,
+                    "Borrowed local value has moved field at exit: ''{0}''.",
+                    PathRenderer
+                )
+                map.put(
+                    UniquenessErrors.INVALID_DUPLICATE_UNIQUE_ARGUMENT,
+                    "Invalid attempt to pass the same unique argument ''{0}'' twice.",
```
ligee: "error: Invalid attempt" is excessive. Justt "Attempt" will suffice imo.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessErrorMessages.kt
```diff
@@ -0,0 +1,66 @@
...
+                    PathRenderer
+                )
+                map.put(
+                    UniquenessErrors.INVALID_OVERLAPPING_UNIQUE_ARGUMENTS,
+                    "Invalid attempt to pass unique argument ''{0}'' with overlapping unique argument ''{1}''.",
```
ligee: ditto

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionEscapeUniquenessConsistencyChecker.kt
```diff
@@ -0,0 +1,94 @@
...
+object FunctionEscapeUniquenessConsistencyChecker : FirFunctionChecker(MppCheckerKind.Common) {
+    context(context: CheckerContext, reporter: DiagnosticReporter)
+    override fun check(declaration: FirFunction) {
+        val graph = declaration.controlFlowGraphReference?.controlFlowGraph ?: return
+        val uniquenessStateFlows = graph.resolveUniquenessStateFlows()
+
+        for (node in graph.nodes) {
+            if (node.isDead) continue
+
+            val inputUniquenessState = uniquenessStateFlows.readInputUniquenessStateOf(node) ?: EmptyUniquenessState
```
ligee: maybe should be lazy to avoid calculating for all non-escaping cases

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionEscapeUniquenessConsistencyChecker.kt
```diff
@@ -0,0 +1,94 @@
...
+
+            for (escapingExpression in node.resolveEscapes()) {
```
ligee, line 84: Do I understand correctly, that interpretation of "escape" is pretty wide?
Do we have tests for throw jump nodes other than return?

formver.compiler-plugin/testData/diagnostics/uniqueness_checker/call.kt
```diff
@@ -0,0 +1,350 @@
...
+fun `borrow multiple unique subproperties`(z: @Unique B) {
+    borrow(z.y.x)
+    borrow(z.y.w)
+}
+
+// Borrowing partially-inconsistent subproperties
+
+fun `borrow partially moved`(z: @Unique B) {
+    consume(z.y)
+    borrow(<!ESCAPE_UNIQUENESS_INCONSISTENCY!>z<!>)
```
ligee, line 260: It is unclear why the report here is about escape, rather than partially moved?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionUseAfterMoveChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.jetbrains.kotlin.fir.declarations.FirFunction
+import org.jetbrains.kotlin.fir.expressions.FirExpression
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.CFGNode
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.ExitSafeCallNode
+import org.jetbrains.kotlin.fir.resolve.dfa.cfg.QualifiedAccessNode
+import org.jetbrains.kotlin.fir.resolve.dfa.controlFlowGraph
+import org.jetbrains.kotlin.formver.uniqueness.plugin.UniquenessErrors.INVALID_MOVED_ACCESS
+
+/**
+ * Resolves expressions that read paths from the uniqueness state at this CFG node.
```
ligee, line 17: It would be nice to explain why only these two.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionMovedAccessChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+/**
+ * Checks that expressions do not read paths that have already been moved.
+ */
+object FunctionMovedAccessChecker : FirFunctionChecker(MppCheckerKind.Common) {
+    context(context: CheckerContext, reporter: DiagnosticReporter)
+    override fun check(declaration: FirFunction) {
+        val graph = declaration.controlFlowGraphReference?.controlFlowGraph ?: return
+        val uniquenessStateFlows = graph.resolveUniquenessStateFlows()
```
ligee: lazy

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionMovedAccessChecker.kt
```diff
@@ -0,0 +1,48 @@
...
+    when (this) {
+        is QualifiedAccessNode -> sequenceOf(fir)
+        is ExitSafeCallNode -> sequenceOf(fir)
+        else -> emptySequence()
+    }
+
+/**
+ * Checks that expressions do not read paths that have already been moved.
+ */
+object FunctionMovedAccessChecker : FirFunctionChecker(MppCheckerKind.Common) {
```
ligee: Name is misleading, suggesting that we report "moved functions".

ligee:
Do we hava a list of generally unsupported patterns of code? Or the ones where we're not tested/not sure? I feel like it would be useful.

ligee (changes requested):

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/Uniqueness.kt
paganma: I changed the name to `scopeUniqueness`, meaning that the uniqueness of this type corresponds to a reference in the current scope. This is to differentiate it from `parameterUniqueness`, which instead resolves the uniqueness of the type of a value parameter.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
paganma: The `Access` of a particular trie node can actually be changed by `append`. What we are effectively saying here is that every path in `t1` is being appended to every path in `t0`. Since `t1` is not empty, and each of its paths are effectively extending the paths in `t0`, every terminal in `t0` will effectively become an intermediate node. I added a comment clarifying this behavior.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessAttribute.kt
paganma: From what I've observed, `intersect` determines how an attribute is inferred in cases like:
```kotlin
fun f(val x: X) {
    if (x is @Unique A && x is B) {
        // Here `x`'s attributes are intersected
...
```
In this case, `x`'s attribute will be inferred to be `null` (meaning `shared`) which is consistent with the uniqueness lattice. Following this logic, the `intersect` function simply returns the attribute (which may be null).

Furthermore, that these `union` and `intersect` implementations of the `UniquenessAttribute` are actually not relevant to how we compute the uniqueness of a symbol (like `x`), and are only used by the kotlin type resolver to decide whether a type should have the attribute or not.

For resolving the uniqueness of a symbol we instead find the declaration and only consider the attribute that was specified there or the initializer, in the case of initialized properties. 

Finally, having `@Unique` on a type is also only possible if the type refers to a function parameter or local property. So cases such as the example that I just specified above are not allowed.

I added documentation to `UniquenessAttribute` to clarify this.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
paganma: I am not sure that `appendLeaf` would actually be simpler. Currently the only place where this function is used is:

```
receiverState.append(EmptyAccessState.putChild(symbol, AccessState(Access.Terminal)))
```

I think you are suggesting something like this:

```
receiverState.appendLeaf(symbol, Access.Terminal)
```

Note however that if we do this we still need to create a new child `AccessState` for `symbol`, so I think that the complexity would move to the implementation of `AccessState.append`.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/UnifyingExpressionTypeFactResolver.kt
paganma: Short circuiting boolean operators always produce primitive boolean values instead of references, therefore for our current use case we don't need to support them. This said, I do agree that boolean operator operands are technically also control-flow operations so handling them would be consistent with how `collectTails` is specified. In that case every OR or AND operands should be treated as independent tails.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionUniquenessResolver.kt
paganma: This rule is for resolving the uniqueness of `null` to `unique`. I added a small comment to clarify this.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/GraphUniquenessStatesAnalyzer.kt
paganma: Renamed to `getOrInitialized`

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionArgumentUniquenessCollisionChecker.kt
paganma: Renamed to `shouldBeChecked`.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionArgumentUniquenessCollisionChecker.kt
paganma: Renamed to `checkUniquenessCollision`.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/ExpressionArgumentUniquenessCollisionChecker.kt
paganma: Renamed `commonPrefix` to `commonPrefixLength`.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessErrorMessages.kt
paganma: Shortened the diagnostic.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessErrorMessages.kt
paganma: Shortened the diagnostic.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionEscapeUniquenessConsistencyChecker.kt
paganma: Wrapped `inputUniquenessState` initializer in a `lazy` block to compute it only if we need it.

formver.compiler-plugin/locality/src/org/jetbrains/kotlin/formver/type/plugin/UnifyingExpressionTypeFactResolver.kt
paganma: In the end I added a small comment describing the current limitation.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionEscapeUniquenessConsistencyChecker.kt
paganma: I added a few tests on `throw` and `try-catch`es. I also confirmed that currently throws that target local try-catch blocks are considered both as a potential exit-point and escape-point by the checker.  I added comments to describe these limitations. Handling local throws should be possible in principle but the change won't be as trivial as I initially thought. Hence I think this is work for another PR. The main challenge is that it is not always possible to statically resolve the type of an exception, especially if casting is involved.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/QualifiedAccessPathReceiverResolver.kt
paganma: After looking more carefully it seems that what we are actually interested in here is the `dispatchReceiver`, since that is the receiver that is actually considered at runtime for object references. Currently we do not support specifying extension properties as `@Unique`, so the path receiver can only be a `dispatchReceiver`. In the future we may want to change strategy for supporting extension properties and the likes. I added a few more tests and a TODO describing the current limitation.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/FunctionMovedAccessChecker.kt
paganma: Renamed to `FunctionUseAfterMoveChecker`.

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/PathTrie.kt
paganma: Added documentation.

paganma:
> Do we hava a list of generally unsupported patterns of code? Or the ones where we're not tested/not sure? I feel like it would be useful.

I added a small README in the uniqueness module where I describe the current limitations. The main untested feature is how the uniqueness checker should behave with closures.

ligee (approved):

PR #209, opened 2026-07-10
Treat exhaustive `when` fallthrough as unreachable

komiputer:
## Summary

An expression-position `when` that Kotlin accepts as exhaustive without an
`else` was lowered with a `UnitLit` fallthrough. On the
compiler-guaranteed-impossible fallthrough this fabricated a value of the wrong
type, and without exhaustiveness information Viper believed the path reachable —
so a total function over a sealed interface failed to verify.

This reads the frontend's exhaustiveness decision in `convertWhenBranches`:
when the `when` is `ExhaustivenessStatus.ProperlyExhaustive` and has no
syntactic `else`, the missing fallthrough emits the existing `ErrorExp`
(`inhale false`) path instead of `UnitLit`. Exhaustiveness is trusted from FIR,
matching SnaKt's existing smart-cast trust model — no `RuntimeTypeDomain`
changes.

Because exhaustiveness is read from FIR, the change is generic: sealed
interfaces, sealed classes, and exhaustive Boolean `when`s all flow through the
same path.

## Verification

Verified against the `pipeline/sealed-interfaces` branch of
`komiputer/snakt-usage-example`: a recursive sealed-interface `Expr` evaluator
(exhaustive `when`, no `else`) plus positive smart-cast narrowing, building with
0 Viper verification errors.

Golden diagnostic tests cover the sealed-interface total `when` (verifies) and
the still-checked branch body — an exhaustive `when` whose body violates a
contract still fails to verify. Trust applies to totality, not to branch bodies.

## Future work

The fix is generic over FIR exhaustiveness and would cover exhaustive enum
`when` for free, but converting an enum class declaration currently fails
earlier in SnaKt (pre-existing, unrelated). Enum exhaustiveness becomes
reachable once enum-declaration support lands, with no further work on this
path.

PRD: https://bot.jesyspa.dev/kotlin/snakt/prd/sealed-interfaces/

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
@@ -137,9 +137,14 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
     private fun convertWhenBranches(
         whenBranches: Iterator<FirWhenBranch>,
         type: TypeEmbedding,
+        fallthroughUnreachable: Boolean,
         data: StmtConversionContext,
     ): ExpEmbedding {
-        if (!whenBranches.hasNext()) return UnitLit
+        // When Kotlin proves the `when` exhaustive but there is no syntactic `else`, the missing
+        // fallthrough is unreachable. We trust that determination (as we trust smart-casts, see
+        // `visitSmartCastExpression`) and emit `ErrorExp` (`inhale false`) rather than fabricating a
+        // `UnitLit` of the wrong type.
+        if (!whenBranches.hasNext()) return if (fallthroughUnreachable) ErrorExp else UnitLit
 
         val branch = whenBranches.next()
 
@@ -149,7 +154,7 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
         } else {
             val cond = data.convert(branch.condition).withType { boolean() }
             val thenExp = data.withNewScope { convert(branch.result) }
-            val elseExp = convertWhenBranches(whenBranches, type, data)
+            val elseExp = convertWhenBranches(whenBranches, type, fallthroughUnreachable, data)
             If(cond, thenExp.withType(type), elseExp.withType(type), type)
         }
     }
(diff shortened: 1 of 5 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -137,9 +137,14 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
         type: TypeEmbedding,
+        fallthroughUnreachable: Boolean,
         data: StmtConversionContext,
     ): ExpEmbedding {
-        if (!whenBranches.hasNext()) return UnitLit
+        // When Kotlin proves the `when` exhaustive but there is no syntactic `else`, the missing
+        // fallthrough is unreachable. We trust that determination (as we trust smart-casts, see
+        // `visitSmartCastExpression`) and emit `ErrorExp` (`inhale false`) rather than fabricating a
+        // `UnitLit` of the wrong type.
+        if (!whenBranches.hasNext()) return if (fallthroughUnreachable) ErrorExp else UnitLit
```
jesyspa, line 147: ErrorExp isn't the right embedding to use here; we probably want an unreachable embedding or just an inhale(false) directly (I think we have an Inhale ExpEmbedding? if not, you'll need it for manual inhales/exhales anyway)

formver.compiler-plugin/testData/diagnostics/verification/control_flow/exhaustive_when.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import org.jetbrains.kotlin.formver.plugin.AlwaysVerify
+import org.jetbrains.kotlin.formver.plugin.verify
+
+sealed interface Expr
+class Const(val value: Int) : Expr
+class Neg(val operand: Const) : Expr
+
+// An exhaustive `when` over a sealed interface with no `else` is total: the missing fallthrough is
+// unreachable (`inhale false`), so the function verifies as always returning an Int.
```
jesyspa, line 11: Implementation details in comments are a bad pattern.

formver.compiler-plugin/testData/diagnostics/verification/control_flow/exhaustive_when.kt
```diff
@@ -0,0 +1,28 @@
...
+@AlwaysVerify
+fun <!VIPER_TEXT!>eval<!>(e: Expr): Int = when (e) {
```
jesyspa, line 13: The test is correct, because of how we translate things, but it's not obviously correct.  (Why there should be a third branch is confusing.)  Would perhaps be worth adding a more explicit test.

formver.compiler-plugin/testData/diagnostics/verification/control_flow/exhaustive_when.kt
```diff
@@ -0,0 +1,28 @@
...
+    is Const -> e.value
+    is Neg -> -e.operand.value
+}
+
+// Totality is trusted, but branch bodies are still checked. `r` may be negative (e.g. Const(-1)),
+// so the assertion below fails to verify even though the `when` is total.
```
jesyspa, line 19: Both comment and example don't make sense: it looks like a verification shortcoming.

jesyspa (changes requested):

PR #214, opened 2026-07-22
Resolve primitive expressions to `shared` instead of `unique`

### LINKED ISSUE #213: Disable resolving primitive values as `unique`.
Currently the uniqueness checker allows us to specify primitive variables as `unique`. In addition to this, primitive literals also resolve to `unique`. This allows us to write code such as:
```
val x: @Unique Int = 1
```

However this means that we can multiple primitive variables with the same value, which breaks the invariant stating that if something is `unique` then it is not aliased:
```
val x: @Unique Int = 1
val y: @Unique Int = 1

if (x == y) { /* shouldn't be possible */ }
```

Ultimately it would be better to simply disable specifying primitive variables and parameters as unique.

paganma:
This PR changes the behavior of the expression uniqueness resolver so that primitive expressions (including literals) resolve to shared instead of unique. It also adds tests covering the new cases the checker must handle.

### DIFF
--- a/formver.compiler-plugin/testData/diagnostics/uniqueness_checker/continue.kt
+++ b/formver.compiler-plugin/testData/diagnostics/uniqueness_checker/continue.kt
@@ -1,0 +1,10 @@
+// UNIQUE_CHECK_ONLY
+
+import org.jetbrains.kotlin.formver.plugin.Borrowed
+import org.jetbrains.kotlin.formver.plugin.Unique
+
+fun `continue conforms to unique`() {
+    while (true) {
+        val x: @Unique Nothing = continue
+    }
+}
--- a/formver.compiler-plugin/testData/diagnostics/uniqueness_checker/continue.kt
+++ b/formver.compiler-plugin/testData/diagnostics/uniqueness_checker/continue.kt
@@ -3,14 +3,28 @@
 import org.jetbrains.kotlin.formver.plugin.Borrowed
 import org.jetbrains.kotlin.formver.plugin.Unique
 
+fun makeUnique(): @Unique Any? = null
+
 fun `continue conforms to unique`() {
     while (true) {
         val x: @Unique Nothing = continue
     }
 }
 
+fun `continue in elvis conforms to unique`() {
+    while (true) {
+        val x: @Unique Any = makeUnique() ?: continue
+    }
+}
+
 fun `continue conforms to shared`() {
     while (true) {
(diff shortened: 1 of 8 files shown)


formver.compiler-plugin/testData/diagnostics/uniqueness_checker/continue.kt
```diff
@@ -0,0 +1,16 @@
+// UNIQUE_CHECK_ONLY
+
+import org.jetbrains.kotlin.formver.plugin.Borrowed
+import org.jetbrains.kotlin.formver.plugin.Unique
+
```
ligee, line 5: It seems to me that such a code is not useful at all, so why do we need to have a test about it?
If you have some scenario in mind, where the uniquiness of Nothing could be useful, then it is probably a good idea to show such scenario in the code.
Otherwise I'd rathere suggest to drop such tests.

formver.compiler-plugin/testData/diagnostics/uniqueness_checker/return.kt
```diff
@@ -51,3 +51,7 @@ fun `return unique-borrowed subproperty`(a: @Unique @Borrowed B): Any {
 fun `return shared from unique function`(a: Any) : @Unique Any {
     return <!UNIQUENESS_MISMATCH!>a<!>
 }
+
```
ligee, line 54: The same comment about usefulness applies here

ligee (changes requested):

formver.compiler-plugin/testData/diagnostics/uniqueness_checker/continue.kt
paganma: I added more tests for both `continue` and `break` that should represent the cases that I think would be better to handle i.e. those involving branching expressions.

formver.compiler-plugin/testData/diagnostics/uniqueness_checker/return.kt
paganma: Added some better examples for `return` as well.

jesyspa (approved):
I am a bit on the fence on the amount of tests.  On the one hand, I think it's good to have a comprehensive test suite, especially with more AI writing our code.  On the other, having a lot of routine tests makes the interesting ones harder to find... Overall I think it's okay here.

ligee (approved):

PR #219, opened 2026-08-03
Add Char domain bound [0, 65536)

sachok42:
A Char quantifier variable lowers to a Viper Int, which is unbounded. Without an explicit domain bound, forAll<Char> ranges over all integers including negative code points that no real Kotlin Char can hold. Constrain it to [0, 65536) so the verifier can discharge properties that hold for all valid chars without a manual guard.

Adds charDomainBoundOrNull() helper and forall_char_domain.kt regression test.


sachok42 (commented):
A small fix on for all modifier. Looked through, seems good

jesyspa (changes requested):
I don't think this makes sense to implement here: forall is not the only place where we reason about chars.  Rather, the TypeEmbedding should have this as a proven invariant and then it should be handled from there (including guaranteed by operations like +, ideally).

PR #220, opened 2026-08-04
Split developer docs and add tooling for the test loop

komiputer:
Developer material was split across README.md and dev-info.md, and the test loop
had to be rediscovered from the source each time.

## Docs

Developer docs move into `docs/`, split by who reads them:

- `docs/developing.md` — humans developing the plugin: test layout, the three test
  tasks, directives, and the checks CI runs.
- `docs/agents-dev.md` — depth behind AGENTS.md on the scripts.
- `docs/publish-silicon.md` — Silicon publishing, named for the operation.
- `README.md` — keeps the user-facing material.
- `AGENTS.md` — entry point for agents, with `CLAUDE.md` symlinked to it. Holds
  the essentials; anything needing a second reading lives in docs/agents-dev.md.

## Scripts

One driver, `agent-scripts/test.sh`:

    ./agent-scripts/test.sh [pattern]           # conversion only — the fast loop
    ./agent-scripts/test.sh --verify [pattern]  # full pipeline, including verification
    ./agent-scripts/test.sh --update-goldens [pattern] # regenerate goldens, then report what changed

A pattern can be spelled as the testData file is named (`assign_local`) or as
the generated method (`testAssign_local`).

Failure output carries the expected/actual diff directly. Gradle's cross-JVM
serialization strips it off golden-file assertions, so
`DumpAssertionDiffExtension` captures it inside the test JVM; it is registered
for every run via `formver.compiler-plugin/test-resources/` and stays inert
unless `SNAKT_TEST_DUMP_DIR` is set, which `test.sh` does. Nothing is written
into the source tree to arrange it.

`--update-goldens` regenerates goldens and then prints what each one now says,
because
regeneration records whatever the run produced: a function that fails
verification has that failure written into `<name>.viper.diag.txt` and passes
from then on. Reading that output is what distinguishes the two.

Supporting:

- `check-all.sh` — `./gradlew check`, `check-testdata.sh` and `pre-commit`,
  reporting each as passed, failed or skipped. Exit 1 is a real failure; exit 2
  means nothing failed but something did not run, which is what a missing
  `pre-commit` gives you, and is a state to fix rather than a pass.
- `check-testdata.sh` — golden files whose `.kt` source is gone, and empty
  golden files: goldens that assert nothing.
- `lib.sh` and `junit_first_failure.py` — shared helpers, with
  `agent-scripts/tests/run.sh` exercising the parser against fixture XML.

`check-testdata.sh` and the script tests need no build, so they run as
`pre-commit` hooks rather than in the Gradle workflow.

`--update-goldens` rather than `--update`, because `./gradlew update` is a test
mode: convert everything, verify only where conversion changed.


agent-scripts/tests/fixtures/error.xml
sachok42, line 1: Why is the standard timestamp specifically 08.05.2026? Seems very random

sachok42: Same question to the few changes below

sachok42 (commented):
Other than the one genuine question I asked, everything else seems good

agent-scripts/tests/fixtures/error.xml
jesyspa: Probably the date it was generated on, fixed.

sachok42 (approved):
No problems noticed by me

PR #227, opened 2026-08-10
Make the severity of failed proofs configurable

jesyspa:
A failed proof is reported at warning level, so that turning the plugin on cannot
break an adopter's build. Anyone who wants the opposite — a failed proof breaking
the build — had only `-Xwarning-level=VIPER_VERIFICATION_ERROR:error`, which is not
part of the plugin's surface and is documented nowhere.

This adds the switch:

```kotlin
formver {
    verificationErrorSeverity("error")  // or "warning", the default
}
```

and the corresponding compiler-plugin option `verification_error_severity`.

## Design

The Kotlin diagnostic API fixes a severity per `KtDiagnosticFactory`, so a
configurable severity takes one set of factories per severity. `VerificationErrors`
keeps the warning-level factories and `StrictVerificationErrors` holds error-level
ones; both implement `VerificationDiagnostics`, which is what the reporting path
now takes. Because a factory's name is its property name, the two sets carry the
*same* diagnostic names — so which set is in use is invisible to anything that
refers to a diagnostic by name: a `@Suppress`, an `-Xwarning-level`, a test's
expected tags. The shared interface is also what keeps the two sets from drifting:
a diagnostic added to one must be added to the other.

The switch covers every failed-proof diagnostic, not only
`VIPER_VERIFICATION_ERROR` — a build that broke on an unproven postcondition but
not on a possible index-out-of-bounds would be incoherent.

## Tests

`testData/diagnostics/verification/error_severity/` holds the same failing proof
twice, differing only in the `FAIL_ON_VERIFICATION_ERROR` directive. The
`.viper.diag.txt` goldens record the severity, so the pair asserts `warning:`
against `error:`.

The harness collects diagnostics rather than compiling, so it cannot assert that a
build fails. That was checked by hand against a consumer project resolving the
plugin from mavenLocal: on the default the failed proof is `w:` and the build
succeeds; with `verificationErrorSeverity("error")` it is `e:` and `:compileKotlin`
fails.

### DIFF
--- a/formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
+++ b/formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
@@ -7,6 +7,7 @@ package org.jetbrains.kotlin.formver.plugin.compiler
 
 import com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.diagnostics.*
+import org.jetbrains.kotlin.formver.common.VerificationErrorSeverity
 
 object PluginErrors : KtDiagnosticsContainer() {
     val VIPER_TEXT by info2<PsiElement, String, String>(SourceElementPositioningStrategies.DECLARATION_NAME)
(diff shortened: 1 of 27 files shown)


formver.gradle-plugin/src/org/jetbrains/kotlin/formver/gradle/model/impl/FormVerImpl.kt
sachok42, line 1: serial version of what?

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
```diff
@@ -27,13 +28,39 @@ object PluginErrors : KtDiagnosticsContainer() {
...
+    }
+}
 
-object VerificationErrors : KtDiagnosticsContainer() {
-    val CONDITIONAL_EFFECT_ERROR by warning2<PsiElement, String, String>()
-    val VIPER_VERIFICATION_ERROR by warning1<PsiElement, String>()
-    val POSSIBLE_INDEX_OUT_OF_BOUND by warning2<PsiElement, String, String>()
-    val UNEXPECTED_RETURNED_VALUE by warning1<PsiElement, String>()
-    val INVALID_SUBLIST_RANGE by warning2<PsiElement, String, String>()
+object VerificationErrors : KtDiagnosticsContainer(), VerificationDiagnostics {
```
sachok42, line 58: is multi-inheritance a must-have hear? Generally preferable to avoid it

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
```diff
@@ -45,3 +72,12 @@ object VerificationErrors : KtDiagnosticsContainer() {
...
 }
+
+object StrictVerificationErrors : KtDiagnosticsContainer(), VerificationDiagnostics {
+    override val CONDITIONAL_EFFECT_ERROR by error2<PsiElement, String, String>()
+    override val VIPER_VERIFICATION_ERROR by error1<PsiElement, String>()
+    override val POSSIBLE_INDEX_OUT_OF_BOUND by error2<PsiElement, String, String>()
+    override val UNEXPECTED_RETURNED_VALUE by error1<PsiElement, String>()
+    override val INVALID_SUBLIST_RANGE by error2<PsiElement, String, String>()
+    override fun getRendererFactory() = FormalVerificationPluginErrorMessages
+}
```
sachok42, line 83: same

sachok42 (changes requested):
highlighted some minor issues, overall seems good

jesyspa:
Superseded by #272.

PR #228, opened 2026-08-10
Add formver.annotations as an automatic dependency of the Gradle plugin

jesyspa:
## Summary
- `FormVerGradleSubplugin.applyToCompilation` now adds `formver.annotations` (coordinates already computed by `BuildConfig.ANNOTATIONS_LIBRARY_COORDINATES`) as an `implementation` dependency on each compilation's default source set, following the standard `KotlinCompilerPluginSupportPlugin` pattern.
- README no longer tells consumers to manually add the `formver.annotations` dependency.

Found while standing up a fresh consumer project from `mavenLocal`: applying the Gradle plugin didn't give access to the annotations without the extra manual dependency line.

## Test plan
- Compiled `formver.gradle-plugin` (no deprecation warnings; uses `KotlinCompilation.defaultSourceSet.dependencies { implementation(...) }`).
- `./gradlew publishToMavenLocal`, then built a throwaway consumer project applying only `id("org.jetbrains.kotlin.formver")` with no manual `formver.annotations` dependency — `compileKotlin` succeeded and `dependencies --configuration compileClasspath` showed `formver.annotations` resolved automatically.
- `./agent-scripts/check-all.sh` (gradle check, check-testdata.sh, pre-commit) all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,13 +84,8 @@ unless you run `gradle` with the `--info` flag.
 ### Annotations
 
 The plugin provides a number of annotations to add specifications to your code.
-To access these, add a dependency to `formver.annotations`:
-
-```kotlin
-dependencies {
-    implementation("org.jetbrains.kotlin.formver:formver.annotations:0.1.0-SNAPSHOT")
-}
-```
+Applying the Gradle plugin automatically adds a dependency on `formver.annotations`,
+so no further setup is needed to access them.
 
 ### Running from the command line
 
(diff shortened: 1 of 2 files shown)


sachok42 (approved):

sachok42:
actually, maybe there is a reason why you initially didn't include it into the dependencies?

jesyspa:
Nah, just an oversight on our part - I think we didn't realise it worked like this.

PR #230, opened 2026-08-10
Treat a SnaKt spec block as a contract for target selection

jesyspa:
## The gap

Under the default `targets_with_contract` selection, "contract" meant a Kotlin `contract { }` block
with resolved effects and nothing else. A function whose only specification was a SnaKt spec block
was therefore not a target — but its callers still got its postconditions, because `embedContract`
extracts a callee's spec whether or not the callee is itself verified.

So this compiled green:

```kotlin
fun alwaysPositive(x: Int): Int {
    postconditions<Int> { it > 0 }
    return x                        // never checked against the postcondition
}

@AlwaysVerify
fun caller() {
    val positive = alwaysPositive(-1)
    verify(positive > 0)            // proved, from the lie
}
```

Adding `@AlwaysVerify` to `alwaysPositive` made it fail as it should. Without the annotation every
spec block was an axiom nothing checked.

## The fix

`TARGETS_WITH_CONTRACT` now selects a function if it has a Kotlin contract **or** a SnaKt
specification block. The semantics: *a function has a contract when callers may assume something of
it that its own body has to establish.* That is what the selection is for, and the
Kotlin-contract-only reading was narrower both than the option's name and than its purpose.

Detection mirrors `extractFirSpecification` exactly — `hasFirSpecification` sits next to it so the
two stay in sync. The body's first statement being a `preconditions` or `postconditions` call is
precisely when conversion finds a spec, so selection and semantics cannot drift apart. It does not
touch the lambdas, so it is safe to ask about a body conversion may yet reject.

A `preconditions`-only function is selected too. It promises callers nothing, but its own body still
has to hold up under the precondition it declares, and nothing was checking that either.

`loopInvariants` and `verify` deliberately do not select: neither is visible to callers.

## Other entry points for the same trap

Checked against the plugin rather than guessed at; both confirmed by probe.

- **`@NeverConvert` is a live hole, and this PR does not close it.** The annotation stops the
  function being converted as a target, but `embedAnyFunction` reads it only to decide whether to
  inline a *pure* function's body — the callee's signature, postcondition included, is still
  embedded into the caller's program. A `@NeverConvert` helper with a false postcondition still
  hands that postcondition to its callers, with nothing anywhere checking it. Unlike the default
  selection this is at least explicitly asked for, but it is asked for with an annotation whose
  documented meaning is "skip Viper conversion entirely", not "trust me".
- **`@NeverVerify` is the same shape, but only inside the test harness.** The production checker
  verifies everything it converts and never reads `shouldVerify`; only `VerificationFacade` does. So
  `@NeverVerify` — and `verification_targets_selection` with it — has no effect on a real compile
  today. In a test run the callee is converted, left unverified, and its postcondition assumed by
  callers. Worth deciding deliberately if that flag is ever wired up in production.
- **Cross-module callees.** `embedFormverContract` bails when the body is not visible, so a spec
  from another module is neither assumed nor honoured. Not a hole; a known limitation.
- **`verify(...)` in an unselected function** is silently never checked. Not unsound for callers,
  but the same flavour of false assurance. Left alone here.

The shared root is that assuming a callee's spec and checking that callee are decided independently.
A diagnostic at the point where a spec is assumed of a function nothing verifies would cover all of
these at once; that seems the right follow-up rather than patching each entry point.

## Test

`testData/diagnostics/verification/target_selection/spec_block_is_a_contract.kt`, under a new
`DEFAULT_SELECTION` directive — the existing suite runs everything at `all_targets`, so no test
could see this. On `main` it fails all three goldens. Here `alwaysPositive` reports its postcondition
violation and `atLeastTen` its failing body assertion, while the unspecified function is still left
alone.

No existing golden changed: nothing else in the suite uses `targets_with_contract`.

## Docs

`SPECIFICATIONS.md` said the default verifies "only Kotlin `contract { }` blocks" and presented
`@AlwaysVerify` as the way to get a spec block verified at all; both updated, with a note that
suppressing verification of a specified function turns its spec into an unchecked assumption.
`README.md` says what counts as a contract for the CLI option. `docs/developing.md` documents the
new directive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ The plugin accepts a number of command line options which can be passed via
   `only_warnings`).
 - Option `error_style`: permitted values `user_friendly`, `original_viper` and `both` (default: `user_friendly`).
 - Options `conversion_targets_selection` and `verification_targets_selection`: permitted values `no_targets`,
-  `targets_with_contract`, `all_targets` (default: `targets_with_contract`).
+  `targets_with_contract`, `all_targets` (default: `targets_with_contract`). A function counts as having a
+  contract if it has a Kotlin `contract { }` block or a SnaKt `preconditions`/`postconditions` block.
 - Option `unsupported_feature_behaviour`: permitted values `throw_exception`, `assume_unreachable` (default:
   `throw_exception`).
 
--- a/SPECIFICATIONS.md
+++ b/SPECIFICATIONS.md
@@ -4,17 +4,25 @@ SnaKt translates Kotlin code with formal specifications to [Viper](https://www.p
 
 ## Verification Control
 
-By default, SnaKt only verifies functions with Kotlin `contract { }` blocks. To verify functions with SnaKt specifications:
+By default, SnaKt verifies the functions that have a contract: a Kotlin `contract { }` block, or a
+SnaKt `preconditions { }` or `postconditions<T> { }` block. Everything else is left alone:
 
 ```kotlin
 import org.jetbrains.kotlin.formver.plugin.*
 
-@AlwaysVerify  // Enables verification for this function
 fun divide(numerator: Int, denominator: Int): Int {
-    preconditions { denominator != 0 }
+    preconditions { denominator != 0 }   // enough on its own to make this a target
     return numerator / denominator
 }
+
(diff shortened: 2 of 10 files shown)


jesyspa:
Superseded by #274.

PR #231, opened 2026-08-10
Allow field reads in specifications

jesyspa:
Evaluating `list.size` inside any spec block — `preconditions`, `postconditions`, `loopInvariants` — crashed conversion:

```
PureLinearizer used to convert non-pure ExpEmbedding; operation freshAnonVar is not supported in a pure context.
Embedding debug view: (InhaleInvariants(Var(par$l).Field(p$size), type = Int), Int(0))
```

A field read is wrapped in `InhaleInvariants`, which stores its result in a temporary and then inhales what is known about it. A specification has to come out as a single Viper expression, so `PureExpLinearizer` can supply neither the temporary nor the statement position.

`LinearizationContext` now says whether it can hold an `inhale`. Where it cannot, the invariants are dropped and the bare expression is emitted, which withholds information from the verifier but never adds any.

The loop case needed a second fix: a while loop carried only the proven invariants of the variables in scope, so it gave up the permissions to their fields. Any invariant mentioning `l.size` was then ill-formed for lack of `acc(l.size)`. The access invariants now go into the loop invariant too.

Together these make an indexing loop over a list verifiable, which it was not before — the loop havocs the heap, so nothing could relate the index to the size across it.

## Tests

`testData/diagnostics/stdlib/list/size_in_specs.kt` covers a size in a postcondition, a size in a precondition, and a while loop that indexes the list while carrying accumulator and size invariants. All three verify; no `.viper.diag.txt` is recorded.

`./agent-scripts/test.sh --verify` is green at 169 tests, and `./agent-scripts/check-all.sh` passes. No existing golden changed.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
@@ -336,6 +336,9 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
         val invariants = buildList {
             data.retrievePropertiesAndParameters().forEach {
                 addAll(it.provenInvariants())
+                // Without these the loop gives up the permissions to the fields of everything in scope, so
+                // no invariant may mention such a field and nothing may be read from one after the loop.
+                addAll(it.accessInvariants(data.typeResolver))
             }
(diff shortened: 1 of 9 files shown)


sachok42 (approved):

jesyspa (commented):
I want to take a closer look at this - it looks rather suspicious, since it's unclear where to get the permissions otherwise.

jesyspa:
Superseded by #276.

PR #233, opened 2026-08-10
Report ignored preconditions/postconditions/loopInvariants blocks

jesyspa:
## Summary

Spec-block extraction (\`extractFirSpecification\`/\`extractLoopInvariants\` in \`FormverFirBlock.kt\`) only ever looks at fixed statement positions: \`preconditions\` must be the first statement of a function body, \`postconditions\` must be first (if there's no preconditions) or immediately after it, and \`loopInvariants\` must be the first statement of a loop body. Meanwhile the rest of body conversion resolves these calls by name *anywhere* they appear in the block and silently turns them into no-ops (they're registered as no-op special functions). A spec block outside its required position was therefore dropped with no diagnostic — same source, opposite verification outcome.

- Added \`ConversionErrors.IGNORED_SPEC_BLOCK\`, following the existing \`PURITY_VIOLATION\`/\`MINOR_INTERNAL_ERROR\` pattern in that file.
- \`extractFirSpecification\`/\`extractLoopInvariants\` now take a \`reportMisplaced\` callback and scan the full statement list for any spec-block call outside the position that's actually used, reporting it at its exact source.
- Wired into \`ProgramConverter.embedFormverContract\` and \`StmtConversionVisitor.visitWhileLoop\`.
- Added a golden diagnostics test (\`testData/diagnostics/verification/user_invariants/misplaced_spec_blocks.kt\`) covering preconditions-after-postconditions, postconditions-not-immediately-after-preconditions, and loopInvariants-not-first-in-loop.

## Test plan

- [x] \`./agent-scripts/test.sh --verify user_invariants\` — 13/13 passed
- [x] \`./agent-scripts/test.sh --verify\` (full suite) — 169/169 passed
- [x] \`./agent-scripts/check-all.sh\` — passed

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/FormverFirBlock.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/FormverFirBlock.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.kotlin.formver.core.conversion
 
+import org.jetbrains.kotlin.KtSourceElement
 import org.jetbrains.kotlin.fir.declarations.FirAnonymousFunction
 import org.jetbrains.kotlin.fir.expressions.*
 import org.jetbrains.kotlin.fir.symbols.impl.FirFunctionSymbol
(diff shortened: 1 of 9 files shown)


sachok42 (changes requested):
While this PR's goal is to report incorrectly located SnaKt blocks instead of ignoring them, wrongfully nested blocks still get ignored

By wrongfully nested blocks, I mean something like

`
fun f(x: Boolean){
    if (x){
        preconditions { x }
    }
}
`

jesyspa:
Superseded by #274.

PR #234, opened 2026-08-10
Pass an argument for every parameter of a call

jesyspa:
Reported as a string bug: after `a.startsWith(b)`, `verify(false)` was accepted, so
everything downstream of such a call was vacuously proved on a green build.

## Root cause

Not the string model — `startsWith`/`endsWith` have no model at all, and neither does
any other unmodelled stdlib function. They are emitted as body-less Viper methods whose
only contract is the type invariants of their arguments and result, which is sound.

The hole is in how a call passes its arguments. `String.startsWith` is
`startsWith(prefix: String, ignoreCase: Boolean = false)`, and conversion took the call's
arguments straight from the FIR argument list, which does not contain an argument for a
parameter left at its default. The emitted call was therefore

    method startsWith(v_this_extension: Ref, prefix: Ref, ignoreCase: Ref) returns (ret_1: Ref)
      requires isSubtype(typeOf(ignoreCase), boolType())
      ...
    anon := startsWith(a, b)          // two arguments, three formals

Silicon neither rejects this nor treats it as a consistency error; it leaves the third
formal unbound and the state after the call is inconsistent, so `assert false` succeeds.

Nothing about this is string-specific. It fires for any call that omits a defaulted
argument — a plain user function, a `@Pure` function, a constructor, and a parameter
skipped in the middle via named arguments all reproduce it.

## The fix

`StmtConversionVisitor.visitFunctionCall` now lines the call's arguments up against the
callee's value parameters via `resolvedArgumentMapping`, and fills a parameter left at
its default with a fresh value constrained only by the parameter's type:

    var anon_0: Ref
    inhale isSubtype(typeOf(anon_0), boolType())
    anon_1 := startsWith(a, b, anon_0)

That is a sound under-approximation: the caller learns nothing the default would have
told it, and the callee's precondition on that parameter has to hold for an arbitrary
value of its type, so a call can now fail to verify where it previously "verified"
vacuously. Access permissions are deliberately not inhaled for the fresh value —
that would conjure permission to a reference nobody handed us.

Only callees that match arguments against their own parameters go through this —
`NonInlineCallable` and `InlineNamedFunction`. The plugin's own specification functions
(`acc`, `unfold`, …) interpret their arguments themselves and must keep receiving them as
written; filling a gap for those put a variable declaration inside a precondition.

## Tests

- `testData/diagnostics/stdlib/string/prefix_suffix.kt` — the reported case:
  `verify(false)` after `startsWith`/`endsWith` must fail. Fails on `main`.
- `testData/diagnostics/verification/default_arguments.kt` — the general case: a trailing
  default, a default skipped in the middle by a named argument, an inline function, a
  constructor default, and a `@Pure` function showing that the default value itself is not
  assumed.

## Audit — other models, and what is still suspect

- The collection models in `StdLibConverter.kt` (`get`, `subList`, `isEmpty`, `emptyList`,
  `add`) are internally consistent: sizes are related to `old(size)` or to the arguments,
  and no two conditions contradict. None of those functions has a defaulted parameter, so
  none was hit by this bug.
- Inline functions take a different path (`insertInlineFunctionCall` → `getInlineFunctionCallArgs`,
  which `zip`s arguments against formals and so also truncates). There the omitted parameter
  left the substitution map short rather than producing a malformed Viper call, so it failed
  loudly with `INTERNAL_ERROR` instead of verifying vacuously. Filling the gap before the call
  fixes that path too, and the new test covers it.
- `ViperPoweredDeclarationChecker.check` reports an internal exception with
  `reporter.reportOn(e.source, …)`. When the exception carries no source this throws
  `IllegalArgumentException: source must not be null` out of the declaration checker, so an
  internal error surfaces as a compiler crash rather than a diagnostic. Pre-existing and not
  touched here; `e.source ?: declaration.source` would fix it.
- Worth a follow-up: using the parameter's actual default value where FIR exposes it would
  recover the precision this trades away. It is not a drop-in — a default expression may
  reference other parameters, so it cannot simply be converted in the caller's scope.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
@@ -23,6 +23,8 @@ import org.jetbrains.kotlin.formver.common.SnaktInternalException
 import org.jetbrains.kotlin.formver.common.UnsupportedFeatureBehaviour
 import org.jetbrains.kotlin.formver.core.embeddings.LabelLink
 import org.jetbrains.kotlin.formver.core.embeddings.callables.CallableEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.callables.InlineNamedFunction
+import org.jetbrains.kotlin.formver.core.embeddings.callables.NonInlineCallable
 import org.jetbrains.kotlin.formver.core.embeddings.callables.insertCall
 import org.jetbrains.kotlin.formver.core.embeddings.callables.isVerifyFunction
 import org.jetbrains.kotlin.formver.core.embeddings.expression.*
@@ -282,16 +284,52 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
             }
         }
 
+    /**
+     * The call's value arguments in the order the callee declares its parameters, `null` where
+     * the call leaves a parameter to its default value.
+     *
+     * `null` when the mapping is unavailable, in which case the arguments have to be taken as
+     * they come.
+     */
+    private fun FirFunctionCall.argumentsPerParameter(symbol: FirFunctionSymbol<*>): List<FirExpression?>? {
+        val mapping = resolvedArgumentMapping ?: return null
+        val byParameter = mapping.entries.associate { (argument, parameter) -> parameter.symbol to argument }
+        return symbol.valueParameterSymbols.map { byParameter[it] }
+    }
+
     override fun visitFunctionCall(functionCall: FirFunctionCall, data: StmtConversionContext): ExpEmbedding {
         val symbol = functionCall.toResolvedCallableSymbol() as? FirFunctionSymbol<*>
             ?: throw NotImplementedError("Only functions are expected as callables of function calls, got ${functionCall.toResolvedCallableSymbol()}")
 
         val callee = data.embedAnyFunction(symbol)
-        return callee.insertCall(
-            functionCall.functionCallArguments.withVarargsHandled(data, callee),
-            data,
-            data.embedType(functionCall.resolvedType),
-        )
+        val returnType = data.embedType(functionCall.resolvedType)
+        // Only a call that matches its arguments against the callee's own parameters needs this;
+        // the plugin's specification functions interpret the arguments they were given themselves.
+        val matchesParameters = callee is NonInlineCallable || callee is InlineNamedFunction
+        val arguments = if (matchesParameters) functionCall.argumentsPerParameter(symbol) else null
+        if (arguments == null || arguments.none { it == null }) {
+            return callee.insertCall(
+                functionCall.functionCallArguments.withVarargsHandled(data, callee),
+                data,
+                returnType,
+            )
+        }
+
+        // A default value is not available at the call site, and a Viper call that simply omits the
+        // argument leaves the callee's formal unbound, which makes the verification state
+        // inconsistent and proves anything afterwards. Pass a value of the parameter's type instead
+        // and assume nothing else about it: that loses what the default would have told us, but
+        // only that.
+        val declarations = mutableListOf<Declare>()
+        val receivers = listOfNotNull(functionCall.dispatchReceiver, functionCall.extensionReceiver)
+        val args = receivers.map { data.convert(it) } +
+                arguments.zip(symbol.valueParameterSymbols) { argument, parameter ->
+                    if (argument != null) return@zip data.convert(argument)
+                    val (declaration, value) = data.unconstrainedValue(data.embedType(parameter.resolvedReturnType))
+                    declarations.add(declaration)
+                    value
+                }
+        return (declarations + callee.insertCall(args, data, returnType)).toBlock()
     }
 
     override fun visitImplicitInvokeCall(
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionContext.kt
@@ -149,6 +149,17 @@ fun StmtConversionContext.embedPropertyAccess(accessExpression: FirPropertyAcces
     }
 
 
+/**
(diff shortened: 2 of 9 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -282,16 +284,52 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
         val callee = data.embedAnyFunction(symbol)
-        return callee.insertCall(
-            functionCall.functionCallArguments.withVarargsHandled(data, callee),
-            data,
-            data.embedType(functionCall.resolvedType),
-        )
+        val returnType = data.embedType(functionCall.resolvedType)
+        // Only a call that matches its arguments against the callee's own parameters needs this;
+        // the plugin's specification functions interpret the arguments they were given themselves.
+        val matchesParameters = callee is NonInlineCallable || callee is InlineNamedFunction
```
sachok42, line 308: What if further into development we get more than two of such kinds of calees? Lambdas in example. 
Can we do this less case specific? As far as I understand the current code, it would break to the original bug if we reproduce it with some expression that doesn't fall into the specific listed categories. Probably this should work for all kinds of calees, not these two specifically

sachok42 (changes requested):

jesyspa:
Reviewed the fix and checked a few of its claims against the emitted Viper. It is correct and well-targeted, and the goldens do pin the regression (`Assert might fail. Assertion false might not hold.` for all four omission cases and both stdlib string cases). Four things below; the first is the concern already raised in review.

## 1. The callee whitelist: the enumeration is complete, but its shape is wrong

The list is exhaustive today. From `visitFunctionCall`, `embedAnyFunction` can only produce `FullySpecialKotlinFunctionImpl`, `StringPlusAnyFunction`, `NonInlineCallableImpl`, `NonInlineFunctionSignature`, or `InlineNamedFunction`, and none of the excluded ones has defaulted parameters. The lambda case cannot arise at all: `LambdaExp` is reachable only from `visitImplicitInvokeCall`, and Kotlin function types have no default parameter values.

So this is not incomplete. But the failure mode is worth designing against: `matchesParameters = callee is NonInlineCallable || callee is InlineNamedFunction` makes the unsound path the *default*. A new `CallableEmbedding` inherits the vacuity bug silently, and the symptom is a green build in which everything downstream is proved — the worst signal to debug from.

Suggest moving the decision onto `CallableEmbedding` as a member (a property, or better, letting the callable fill its own gaps), so a new implementation has to state its answer rather than getting the unsound one by omission. Behaviour is unchanged.

## 2. The fill path bypasses `withVarargsHandled`

Arguments on the new path are converted with plain `data.convert`. For a callee with both a vararg parameter and an omitted default, the `FirVarargArgumentsExpression` reaches `visitElement`/`handleUnimplementedElement` instead of the deliberate `SnaktInternalException("Vararg arguments are currently supported for `verify` function only.")`. Latent today, but it downgrades a precise diagnostic to a generic one.

Related, in `argumentsPerParameter`:

```kotlin
val byParameter = mapping.entries.associate { (argument, parameter) -> parameter.symbol to argument }
```

`associate` silently keeps only the last entry if several arguments ever map to one parameter. Also unverified: a vararg parameter given zero arguments likely has no mapping entry at all, so it would be filled with an unconstrained `Ref` rather than an empty array. Worth a test either way.

## 3. Two argument-list builders now have to stay in sync

```kotlin
val receivers = listOfNotNull(functionCall.dispatchReceiver, functionCall.extensionReceiver)
```

re-implements the first half of `functionCallArguments`, which itself has to match `FunctionSignature.formalArgs` — three places, one convention. Collapsing to a single builder that takes `List<FirExpression?>` and applies vararg handling to the non-null entries would remove the early return in `visitFunctionCall` and take points 2 and 3 with it.

## 4. The doc comment on `unconstrainedValue` overstates its guarantee

"Access permissions are deliberately left out" holds for what `unconstrainedValue` emits, but is not a property the value carries: on the inline path, `insertInlineFunctionCall` -> `getInlineFunctionCallArgs` -> `argumentDeclaration` re-decorates every argument with `withNewTypeInvariants { proven = true; access = true }`. I checked the emitted Viper for a class-typed defaulted parameter and no `acc` appears, so this is comment accuracy rather than a bug — but the comment should describe what the function emits, not a property of the resulting value.

## Test coverage

Covered: trailing default, middle default via named arguments, inline, constructor, `@Pure` "default value is not assumed", plus the two stdlib string cases.

Not covered, in priority order:

- **A reference-typed defaulted parameter.** Adding `class Cell(var value: Int)` with a non-inline and an inline `f(c: Cell = Cell(0))` and running the verification pipeline: both emit `inhale isSubtype(typeOf(anon), Cell())` and `assert false` correctly fails. So the behaviour is right — but nothing pins it, and this is exactly the case the permission decision in point 4 is about.
- A vararg parameter combined with an omitted default (see point 2).
- A defaulted parameter of function or nullable type.

formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
jesyspa: Checked this. The list is exhaustive today, but your concern about its shape holds.

Reachable from `visitFunctionCall` via `embedAnyFunction`: `FullySpecialKotlinFunctionImpl`, `StringPlusAnyFunction`, `NonInlineCallableImpl`, `NonInlineFunctionSignature`, `InlineNamedFunction`. The two excluded kinds are the plugin's own spec functions and `String.plus(Any)`, neither of which has defaulted parameters. Lambdas specifically cannot hit this: `LambdaExp` is only reachable from `visitImplicitInvokeCall`, and Kotlin function types have no default parameter values.

So nothing currently falls through — but if something did, it would fall through into the unsound path, silently, and the symptom would be a green build proving everything downstream. Better to move the decision onto `CallableEmbedding` as a member so a new implementation has to answer the question instead of inheriting the wrong answer by omission. Same behaviour, and the next callable kind cannot regress it by accident.

jesyspa:
Superseded by #273.

PR #259, opened 2026-08-20
Add uniqueness checker coverage for linked lists and binary trees

jesyspa:
Adds `linked_list.kt` and `binary_tree.kt` to `testData/diagnostics/uniqueness_checker/`, exercising the uniqueness checker on recursive data structures (`Node` with `@Unique next`, `Tree` with `@Unique left/right`).

Seven scenarios behave as intended: push-front, use-after-move of the old head, tail detach, escape with a moved field, leaf insert, child swap via a temporary, and duplicate-unique-argument detection on a property path (previously only covered for bare locals).

Three gaps are documented in-file rather than fixed:
- Borrowed traversal of a `@Unique` next chain is impossible: advancing the cursor moves the cursor's own `next`, and the loop-head join keeps it moved.
- Safe-call assignment (`pivot?.right = root`) is not recognized as restoring a moved field, so a rotation on a nullable pivot gets a spurious escape diagnostic; the non-null and smart-cast forms check clean.
- Iterative list reversal makes the analysis hang (committed commented out, with a comment narrowing the trigger to writing back a unique local carried over from the previous iteration).

Note for reviewers: the inline `<!DIAG!>` markers in this suite are not asserted — `AbstractPhasedDiagnosticTest` compares only the offset-free `.fir.diag.txt` golden. The markers here were verified by hand against emitted offsets.

### DIFF
--- a/formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/PhasedDiagnosticTestGenerated.java
+++ b/formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/PhasedDiagnosticTestGenerated.java
@@ -171,6 +171,12 @@ public class PhasedDiagnosticTestGenerated extends AbstractPhasedDiagnosticTest 
     }
 
     @Test
+    @TestMetadata("binary_tree.kt")
+    public void testBinary_tree() {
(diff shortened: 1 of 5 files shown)


jesyspa:
Superseded by #278.

Issue #138, opened 2026-05-06
Move locality tests in locality module


paganma:
Currently locality tests are residing in `formver.compiler-plugin/testData`. It would be better if they were bundled within the `locality` module.

PR #260, opened 2026-08-20
Consume uniqueness and locality results in verification

jesyspa:
The uniqueness and locality checkers ran as an opt-in side channel: the verification
plugin never registered them, and conversion read `@Unique`/`@Borrowed` off
annotations of its own rather than off what the checkers compute. Verification could
therefore encode an annotation that nothing had validated. This makes verification
consume them.

Built from the abandoned `marco/add-uniqueness-locality-checkers-to-viper-plugin-checkers`
(40 commits, 17 unsquashed fixups), rebased onto `main` and reworked where noted below.

## What changed

**Registration.** `FormalVerificationPluginExtensionRegistrar` now registers the FIR
type-attribute extensions and the uniqueness/locality resolvers unconditionally, so
`@Unique`/`@Borrowed` on a type always reach conversion. The separate registration
from the CLI registrar and from the compiler-plugin test fixture is gone.

**Conversion reads results, not annotations.** `ProgramConverter` and
`SignatureCreation` read `resolvedType.scopeUniqueness` and `.locality`.

**`@Unique`/`@Borrowed` are type-position-only**, which is what a type attribute can
see. The verification testData moves to that form, and off `Int` — the annotation on
a primitive never produced a meaningful predicate. The `uniqueness_checker` testData
was already written this way on `main`.

**Specification calls are uniqueness-neutral.** A `verify` or `preconditions` block
never executes, so it must neither consume nor mutate what it mentions, but the CFG
analysis saw a call and moved the arguments. Specification DSL entry points now carry
`@SpecificationHelper`; the analysis skips calls a `UniquenessNeutralCallPredicate`
accepts, at both call-enter and call-exit, so a neutral call can neither move a path
nor restore one. `asSpecificationAware()` additionally stands the flow-sensitive
checkers down inside a specification argument.

**New test.** `uniqueness_checker/specification.kt` pins all of that: a unique value
named in `verify`, `acc`, `old`, `UniquePred`/`fold`/`unfold`, `preconditions`,
`postconditions`, `loopInvariants` and `forAll` stays usable afterwards, and a value
moved *before* a specification call is still `INVALID_MOVED_ACCESS` after it. Dropping
the neutrality predicate makes it fail, which I checked. The case that carries the
guard is `acc(a.x)`: it is the only DSL entry point in the file that takes the unique
value itself and is not `@Borrowed`, so without the guard `a.x` stays moved and the
following use reports an escaping moved field.

## Decisions I made that the source branch left open

- **`checkUniqueness`/`checkLocality` default to `true`**, in one place
  (`PluginConfiguration`) that the CLI registrar reads. The source branch had them at
  `true` in `PluginConfiguration` but still `false` at the CLI registrar, so tests ran
  with the checkers on and real compilations ran with them off. On is the right
  default: the encoding consumes the annotations whether or not the checkers run, so
  off-by-default leaves them unvalidated. No existing verification golden gained a
  diagnostic from the switch.
- **Dropped the `UniquePred` DSL TODO.** `fold`/`unfold` take `Predicate`, and the
  converter matches a predicate *constructor* call; turning `UniquePred` into a
  function would remove the extension point `Predicate` exists to be.
- **Dropped two commits that cancelled out** (an import removed and re-added) and one
  whitespace-only change to an untouched file.
- Checker registration is duplicated between `PluginAdditionalCheckers` and the
  modules' own `UniquenessAdditionalCheckers`/`LocalityAdditionalCheckers`. That is
  deliberate: the plugin's copy wraps some checkers in `asSpecificationAware()`, which
  the modules cannot do.

## Things a reviewer may want to overrule

- `UniquenessExtensionRegistrar` and `UniquenessAdditionalCheckers` now have no
  in-tree caller. I left them; locality keeps its equivalents for its own tests.
- `UNIQUE_CHECK_ONLY` and `LOCALITY_CHECK_ONLY` are now synonyms — both only take
  conversion and verification out of the way, since both checkers always run. I
  updated their stale descriptions rather than collapsing them into one directive.
- Reading a *moved* value inside a specification is reported for `verify(a === b)` but
  not for `preconditions { a.n == 0 }`: the lambda body is a separate declaration that
  `asSpecificationAware()` covers. Both behaviours are asserted by the new test and
  the asymmetry is written up in the uniqueness README's limitations.

## Testing

`./agent-scripts/check-all.sh` green, including `./agent-scripts/test.sh --verify`.

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -15,8 +15,12 @@ diagnostics, alongside golden files holding the diagnostic text:
 The test runners are generated from the testData tree as part of
 `compileTestKotlin`, so a new file is picked up on the next build.
 
-The pipeline splits into conversion (uniqueness checking, conversion, purity
-checking) and verification (Viper consistency checking and verification):
+Locality and uniqueness checking happens in FIR checkers that run in every task
+below and report into `.fir.diag.txt`; the plugin's `check_locality` and
+`check_uniqueness` options switch those diagnostics off, and tests leave them on.
+`@Unique` and `@Borrowed` reach conversion as type attributes either way. Past
+that the pipeline splits into conversion (conversion proper and purity checking)
+and verification (Viper consistency checking and verification):
 
 | Task                       | Conversion | Verification                |
 |:---------------------------|:-----------|:----------------------------|
@@ -43,11 +47,11 @@ Which checks run:
 
 - `NEVER_VALIDATE` — convert but do not verify. Consistency checking still runs.
   This is how a test that is not meant to reach the verifier says so.
-- `UNIQUE_CHECK_ONLY` — uniqueness checking, with locality first. No conversion.
-- `LOCALITY_CHECK_ONLY` — locality checking alone, uniqueness off. No conversion.
-- `ALWAYS_VALIDATE` — verify every target. Verification is already the default,
-  so this changes nothing on its own; it earns its place by overriding the two
-  `*_CHECK_ONLY` directives above.
+- `UNIQUE_CHECK_ONLY`, `LOCALITY_CHECK_ONLY` — neither convert nor verify, so the
+  goldens hold the locality and uniqueness diagnostics alone. The two have the
+  same effect; both checkers run either way.
+- `ALWAYS_VALIDATE` — verify every target. Tests verify every target already, so
+  this changes nothing on its own.
 
 What the diagnostic contains:
 
--- a/formver.common/src/org/jetbrains/kotlin/formver/common/PluginConfiguration.kt
+++ b/formver.common/src/org/jetbrains/kotlin/formver/common/PluginConfiguration.kt
@@ -11,10 +11,17 @@ data class PluginConfiguration(
     val behaviour: UnsupportedFeatureBehaviour,
     val conversionSelection: TargetsSelection,
     val verificationSelection: TargetsSelection,
-    val checkLocality: Boolean,
-    val checkUniqueness: Boolean,
-    val dumpUniquenessCFG: Boolean,
+    val checkLocality: Boolean = DEFAULT_CHECK_LOCALITY,
+    val checkUniqueness: Boolean = DEFAULT_CHECK_UNIQUENESS,
+    val dumpUniquenessCFG: Boolean = false,
 ) {
+    // The Viper encoding reads `@Unique` and `@Borrowed` off types whether or not the checkers run, so leaving them
+    // off gives a program whose annotations nothing validates.
+    companion object {
+        const val DEFAULT_CHECK_LOCALITY = true
+        const val DEFAULT_CHECK_UNIQUENESS = true
+    }
+
     init {
         require(conversionSelection >= verificationSelection) {
             "Conversion options may not be stricter than verification options; converting $conversionSelection but verifying $verificationSelection."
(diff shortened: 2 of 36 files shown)


docs/developing.md
```diff
@@ -43,11 +47,11 @@ Which checks run:
 
 - `NEVER_VALIDATE` — convert but do not verify. Consistency checking still runs.
   This is how a test that is not meant to reach the verifier says so.
-- `UNIQUE_CHECK_ONLY` — uniqueness checking, with locality first. No conversion.
-- `LOCALITY_CHECK_ONLY` — locality checking alone, uniqueness off. No conversion.
-- `ALWAYS_VALIDATE` — verify every target. Verification is already the default,
-  so this changes nothing on its own; it earns its place by overriding the two
-  `*_CHECK_ONLY` directives above.
+- `UNIQUE_CHECK_ONLY`, `LOCALITY_CHECK_ONLY` — neither convert nor verify, so the
```
paganma, line 50: I think that we could have a simpler `VERIFICATION_ONLY` flag instructing the test driver to skip uniqueness and locality checking altogether. I don't really think that this should be a blocking issue as later we can simply move uniqueness-only tests to the `uniqueness` submodule, which will remove the need for the `UNIQUE_CHECK_ONLY` flag. The `LOCALITY_CHECK_ONLY` flag is already superfluous as all the locality-only tests have already been moved to the `locality` submodule (#138).

formver.common/src/org/jetbrains/kotlin/formver/common/PluginConfiguration.kt
```diff
@@ -11,10 +11,17 @@ data class PluginConfiguration(
...
+    val checkLocality: Boolean = DEFAULT_CHECK_LOCALITY,
+    val checkUniqueness: Boolean = DEFAULT_CHECK_UNIQUENESS,
+    val dumpUniquenessCFG: Boolean = false,
 ) {
+    // The Viper encoding reads `@Unique` and `@Borrowed` off types whether or not the checkers run, so leaving them
+    // off gives a program whose annotations nothing validates.
+    companion object {
+        const val DEFAULT_CHECK_LOCALITY = true
+        const val DEFAULT_CHECK_UNIQUENESS = true
+    }
```
paganma, line 23: Since every instantiation of `PluginConfiguration` already provides arguments for `checkLocality` and `checkUniqueness`, I would just remove these default constants.

formver.compiler-plugin/cli/src/org/jetbrains/kotlin/formver/cli/FormalVerificationPluginComponentRegistrar.kt
```diff
@@ -40,21 +38,19 @@ class FormalVerificationPluginComponentRegistrar : CompilerPluginRegistrar() {
...
-        val checkUniqueness = configuration.get(FormalVerificationConfigurationKeys.CHECK_UNIQUENESS, false)
-        val checkLocality = configuration.get(FormalVerificationConfigurationKeys.CHECK_LOCALITY, false)
+        val checkUniqueness = configuration.get(
+            FormalVerificationConfigurationKeys.CHECK_UNIQUENESS,
+            PluginConfiguration.DEFAULT_CHECK_UNIQUENESS
+        )
+        val checkLocality = configuration.get(
+            FormalVerificationConfigurationKeys.CHECK_LOCALITY,
+            PluginConfiguration.DEFAULT_CHECK_LOCALITY
+        )
```
paganma, line 48: Inlining `true` at the default parameter of `get` should be enough.

paganma (commented):
Overall this seems in line with what I submitted in #216. The comments are on how we should handle the test directives and related plugin configuration.

docs/developing.md
jesyspa: There's a few open PRs about configuring verification levels, I think we should just get those merged first and hopefully it'll clarify things.

jesyspa:
Superseded by #279.

PR #261, opened 2026-08-20
Make the uniqueness checker usable from a Gradle build

jesyspa:
`formver { checkUniqueness(true) }` in a Gradle build died with an internal
compiler error on any source, including trivial ones with no `@Unique` anywhere:

    ClassCastException: org.jetbrains.kotlin.kotlinx...PersistentOrderedMap
    cannot be cast to kotlinx.collections.immutable.PersistentMap
      at GraphUniquenessStatesAnalyzer.visitNode

## Cause

Gradle runs the compiler out of `kotlin-compiler-embeddable`, which relocates the
libraries it bundles under `org.jetbrains.kotlin.*` — among them
`kotlinx.collections.immutable` and `com.intellij`. Both surface in the compiler
API this plugin builds on: FIR's control-flow visitors take `PersistentMap`
directly in the signatures the uniqueness analyzer overrides, and the diagnostic
factories mention `PsiElement`. Compiled against the plain compiler, the plugin
names classes the embeddable one does not have.

The shadow jar also shaded in the entire plain compiler (104 MB), which is what
made the mismatch a cast failure rather than a missing class: with the plugin's
own class loader supplying a second, unrelocated copy of those packages, the
plugin loaded its own `PersistentMap` and was handed the compiler's.

`checkLocality(true)` survived only because nothing on its path crosses one of
those types.

## Changes

- Publish `formver.compiler-plugin-embeddable`, a second shadow jar with those
  two packages rewritten to match the embeddable compiler, and resolve that
  artifact from the Gradle plugin. The plain artifact still targets `kotlinc`.
- Stop shading the compiler into the plugin jar — whichever compiler runs the
  plugin supplies it. Both jars drop from 104 MB to 41 MB.
- Disable the aggregator's plain `jar` task. It has no sources, so it produced an
  empty archive under the same name as `shadowJar`, leaving whichever ran last in
  `build/libs` — which can mean publishing an empty jar.

A second defect surfaced once the first was fixed: uniqueness checking resolves
the locality of expressions and symbols, but those resolvers are session
components only the locality extensions install, so `checkUniqueness` alone threw
`IllegalStateException: No ExpressionLocalityResolver in array owner` on the
first declaration that consults one. A trivial source never reaches that path.
The test harness already enables both checkers together and says so; the
compiler plugin now matches. Note the user-visible consequence: `checkUniqueness`
also reports locality diagnostics.

## Verification

End to end against `mavenLocal` in a scratch Gradle project, `--no-daemon`, clean
build each run, with a non-trivial source (`testData/diagnostics/uniqueness_checker/aliasing.kt`
with its diagnostic markers stripped): `checkUniqueness` alone, `checkLocality`
alone, and both together all compile without an internal error and report the
uniqueness diagnostics the golden file expects.

The `kotlinc` path was checked separately against the 2.3.0 compiler and reports
the same diagnostics. (The `kotlinc` on this machine is 2.4.10 and cannot load
the plugin at all — `FirExtensionRegistrarAdapter` changed supertype since 2.3.0.
That is unrelated to this change and predates it.)

### DIFF
--- a/formver.compiler-plugin/build.gradle.kts
+++ b/formver.compiler-plugin/build.gradle.kts
@@ -1,3 +1,5 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
 plugins {
     kotlin("jvm")
     `java-test-fixtures`
@@ -35,7 +37,10 @@ dependencies {
     implementation(project(":formver.compiler-plugin:plugin")) { isTransitive = false }
     implementation(project(":formver.compiler-plugin:locality")) { isTransitive = false }
     implementation(project(":formver.common")) { isTransitive = false }
-    implementation(kotlin("compiler"))
+    // Provided by whichever compiler runs the plugin. Shading a second copy in would
+    // put compiler classes in the plugin's own class loader next to the ones the
+    // running compiler uses, and the two sets do not interoperate.
+    compileOnly(kotlin("compiler"))
 
     testFixturesApi(kotlin("test-junit5"))
     testFixturesApi(kotlin("compiler-internal-test-framework"))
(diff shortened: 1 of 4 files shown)


formver.compiler-plugin/build.gradle.kts
```diff
@@ -157,14 +162,46 @@ fun Test.setLibraryProperty(propName: String, jarName: String) {
...
+// its archive — and it names that archive exactly what shadowJar names its own, which
+// leaves whichever ran last in build/libs.
+tasks.jar {
+    enabled = false
+}
+
+// The jar for `kotlinc`, which runs the plugin against the plain compiler.
 tasks.shadowJar {
     archiveClassifier.set("")
 }
```
sachok42, line 176:
```suggestion
tasks.assemble {
    dependsOn(embeddableJar)
}
```
So embeddableJar would be checked in CI

sachok42 (changes requested):

jesyspa:
Superseded by #271.

PR #262, opened 2026-08-20
Check loop-carried unique state

jesyspa:
Loop-carried unique state was mishandled in two connected ways. Both are fixed here.

Stacked on #259, which supplies the tests; the first four commits are that branch. Review from `Read an assignment's right-hand side against the state it enters with` onwards, or rebase once #259 lands.

## Advancing a cursor left a moved field behind

The transfer function for a declaration or assignment projected the RHS substate into the LHS path and only then moved the paths the RHS had read. By that point the LHS was already overwritten, so the move landed on whatever the target now pointed at rather than on what was read.

`current = current.next` therefore marked the `next` field of the node the cursor had just moved *to*. The loop head carried the mark round, the next read of `current.next` was an access to a moved reference, and the borrowed cursor was left with a moved field at exit — on a list the traversal never disturbed.

Project and move against the incoming state, and apply the move before the insert. Overwriting a variable then drops the moved marks under it, which is right when they describe a location the variable no longer reaches.

This is a narrow change: it only alters the outcome when the LHS is a prefix of an RHS path. Everywhere else the two orders commute, which is why no existing golden moved. The move still bites when someone else takes the field first — `drop the tail in a loop` passes `current.next` to `consume` before advancing, and the advance is still reported.

Note what it does *not* fix: the checker is alias-insensitive, so if another variable still reaches the node the cursor left, the field really is moved and nothing says so. That gap predates this change — moving `current.next` never marked `head.next` either.

## The fixed point did not terminate over recursive types

Assignment projects the RHS substate under the LHS path, so a loop over a recursive type deepens the trie once per iteration: `prev` picks up the spine of the node assigned to it, and that spine goes into `current.next` on the way round. `PathTrie.join` unions children, so the states grow without bound and the fixed point is never reached. In-place list reversal was the case that showed it; #259 had to leave it commented out because the check did not come back.

### The choice

Two directions were on the table: depth k-limiting, or widening at loop heads. I took a third that is really k-limiting with the bound chosen structurally rather than numerically — **collapse a path wherever a symbol comes round a second time.**

A fixed depth k is a number with no meaning in the domain: it truncates `a.b.c.d` in straight-line non-recursive code for the same reason it truncates `prev.next.next`, and picking k trades one against the other. Repetition is the thing that actually distinguishes them. Unbounded growth *has* to repeat a symbol, because a loop body can only ever append the finitely many components its own syntax spells out — so cutting at repeats bounds depth by the number of distinct symbols a function mentions, which is finite, while leaving every path over non-recursive data exactly as it was. That is why this lands with no golden churn outside the file under test.

Widening at loop heads would be more precise inside loop bodies, but it makes termination depend on every CFG cycle passing through a node the traversal merges at. Normalizing at every node costs a little precision at depth and buys a bound that holds whatever shape the graph has.

### The mechanism

A `PathTrie` node may now be a **summary node**: a leaf whose value is the join of everything that would otherwise be below it, standing for both its own prefix and every path under it. `join` lets a summary absorb the other side's children; `find` descending past one lands on the summary; `insert` into one is a weak update. `UniquenessState.summarizeRecursivePaths` collapses repeats, and the analyzer normalizes every state it records.

Summaries are approximate, and the README says so: a read past one is treated as touching everything it covers, and restoring a borrow through one restores the whole region rather than the one path borrowed. Neither is reachable without source that spells out a path of length three or more into a recursive spine.

## Result

- `reverse in place` is un-commented and checks with no diagnostics, in seconds.
- `sum values` and `sum a readonly spine in a loop` check clean — the priority case, a unique readonly spine with mutable values walked in idiomatic loop code, now costs nothing that the recursive form does not.
- `drop the tail in a loop` is new, guarding that the reorder did not blunt move detection.
- `./agent-scripts/check-all.sh` passes.

### DIFF
--- a/formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/PhasedDiagnosticTestGenerated.java
+++ b/formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/PhasedDiagnosticTestGenerated.java
@@ -171,6 +171,12 @@ public class PhasedDiagnosticTestGenerated extends AbstractPhasedDiagnosticTest 
     }
 
     @Test
+    @TestMetadata("binary_tree.kt")
+    public void testBinary_tree() {
+      runTest("formver.compiler-plugin/testData/diagnostics/uniqueness_checker/binary_tree.kt");
+    }
+
+    @Test
     @TestMetadata("break.kt")
     public void testBreak() {
       runTest("formver.compiler-plugin/testData/diagnostics/uniqueness_checker/break.kt");
--- a/formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/PhasedDiagnosticTestGenerated.java
+++ b/formver.compiler-plugin/test-gen/org/jetbrains/kotlin/formver/plugin/runners/PhasedDiagnosticTestGenerated.java
@@ -231,6 +231,12 @@ public class PhasedDiagnosticTestGenerated extends AbstractPhasedDiagnosticTest 
     }
 
     @Test
+    @TestMetadata("linked_list.kt")
+    public void testLinked_list() {
+      runTest("formver.compiler-plugin/testData/diagnostics/uniqueness_checker/linked_list.kt");
+    }
+
+    @Test
     @TestMetadata("loop.kt")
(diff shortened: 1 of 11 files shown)


jesyspa:
Superseded by #278.

PR #263, opened 2026-08-20
Fix setup docs and the internal-error report URL

jesyspa:
Setting up the plugin from scratch on a clean machine, following the README as
written, and then running the [usage example](https://github.com/jesyspa/snakt-usage-example).
Verification does work end to end; these are the things that pointed the wrong
way on the road there.

### `pluginManagement` snippet is wrong

It lists `mavenCentral()` and `mavenLocal()`, having dropped
`gradlePluginPortal()`. Copied as written, `kotlin("jvm")` no longer resolves,
because the Kotlin Gradle plugin is published to the portal rather than to
Maven Central. The usage example's `settings.gradle.kts` already does this
correctly, so the README was the only wrong copy.

### Z3 instructions assume root

`export Z3_EXE=/usr/bin/z3` + `sudo cp` needs both root and a Debian-ish
layout. Added a `~/.local/bin` variant, and stated that `Z3_EXE` points at the
binary and not at its directory, which the old comment (`or a different
directory in $PATH`) made ambiguous.

### Nothing said which JDK to use

On JDK 25 both this repository and the usage example fail with:

```
* What went wrong:
25.0.2
```

That is the whole message. `--stacktrace` shows an `IllegalArgumentException`
from `JavaVersion.parse` inside Gradle's embedded build-script compiler —
Gradle 8.14.3 does not support JDK 25. Nothing identifies the JDK as the cause,
so it reads like a corrupt install. I looked at guarding it in our build, but
the failure happens while Gradle compiles the build script, before any project
code runs, so there is nowhere to catch it. Documented instead.

### A missing `Z3_EXE` looks like a plugin bug

It reports `An internal error has occurred. Details: Cannot run prover at
location 'z3': not a file. Please report this at ...`, which is a configuration
problem rather than something to file. Noted in the Z3 section. It does at
least fail the build rather than quietly skipping verification.

Separately, `INTERNAL_ERROR` pointed at `https://github.com/jesyspa/kotlin`,
which is not where SnaKt is developed; it now points at this repository's
issue tracker. One golden held that string and was regenerated — the diff is
the URL line only.

## Testing

- `./agent-scripts/check-all.sh` — `gradle check` and `check-testdata.sh` pass.
  `pre-commit` could not be installed on this machine (the proxy blocks PyPI),
  so I ran its three hooks by hand instead: `check-testdata.sh` and
  `agent-scripts/tests/run.sh` pass, and the changed files satisfy
  `end-of-file-fixer`.
- Rebuilt the plugin and reran the usage example against it, confirming
  verification still reports the expected `Function may return a false value`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,25 @@ Once you've published to your local Maven repository, you can use the Gradle
 plugin to enable verification of your project.
 You can see an example setup at [jesyspa/snakt-usage-example](https://github.com/jesyspa/snakt-usage-example).
 
+### JDK
+
+Build and run with a JDK that your Gradle version supports; JDK 21 is a safe
+choice. The Gradle version this repository builds with (8.14.3) does not support
+JDK 25, and the failure is easy to misread: the build aborts with a message that
+is only the version number,
+
+```
+* What went wrong:
+25.0.2
+```
+
+which comes from Gradle's own build-script compiler rejecting the version string
+(`IllegalArgumentException` in `JavaVersion.parse`, visible under
+`--stacktrace`). It happens before any project code runs, so neither this
+repository nor your own build can catch it and report something friendlier. If
+you see it, point `JAVA_HOME` at an older JDK and stop the daemons
+(`./gradlew --stop`).
+
 ### Setup
 
 In your `settings.gradle.kts`, configure your Gradle plugin repositories to allow local plugins:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -28,19 +28,41 @@ Once you've published to your local Maven repository, you can use the Gradle
 plugin to enable verification of your project.
 You can see an example setup at [jesyspa/snakt-usage-example](https://github.com/jesyspa/snakt-usage-example).
 
+### JDK
+
+Build and run with a JDK that your Gradle version supports; JDK 21 is a safe
+choice. The Gradle version this repository builds with (8.14.3) does not support
+JDK 25, and the failure is easy to misread: the build aborts with a message that
+is only the version number,
```
sachok42, line 36: In parallel, there is a PR on implementing JDK 25 support. Maybe, it's better to merge it first and then delete this paragraph before merging?

sachok42 (changes requested):

README.md
jesyspa: Yeah, not worth pinning these details in the README anyway

jesyspa:
Superseded by #269.

PR #264, opened 2026-08-20
Build on JDK 25, pin the compile toolchain, and guard the Kotlin version

jesyspa:
Makes the build work on JDK 25, stops the published bytecode from depending on
the developer's JDK, and turns the "wrong Kotlin" failure into a message that
says what is wrong.

## What was broken

**The build could not run on JDK 25.** Two things reject the JDK, both for the
same reason: they embed a Kotlin compiler whose bundled IntelliJ
`JavaVersion.parse` throws `IllegalArgumentException: 25.0.4` on feature
versions above 24, and both run inside the Gradle daemon.

- Gradle 8.14.3 compiles the Kotlin build scripts with Kotlin 2.0.21, so the
  build died during script compilation before reaching any task.
- detekt 1.23.7 and 1.23.8 embed the same compiler. Forcing a newer
  `kotlin-compiler-embeddable` onto the `detekt` configuration clears the parse
  failure and is then rejected by detekt's own version guard.

Confirmed directly: `JavaVersion.parse("25.0.4")` throws under
`kotlin-compiler-embeddable` 2.0.21 and returns `25.0.4` under 2.3.0.

**The compile toolchain was whatever JDK launched Gradle.** Nothing pinned it,
so building on JDK 25 produced class file 69 in every published jar, unloadable
on the JDK 21 the project targets. This is not specific to 25 — building on 24
today already ships class file 68.

**A wrong Kotlin failed opaquely.** `kotlinc` 2.4.10 could not load the plugin
at all, failing with `ClassCastException: FirExtensionRegistrarAdapter$Companion
cannot be cast to ProjectExtensionDescriptor`. The message names compiler
classes and gives the user nothing to act on.

## Changes

- Gradle wrapper to 9.7.1, which runs on JDK 25.
- Pin the Kotlin JVM toolchain to 21 for all subprojects, and add the foojay
  resolver so that JDK is provisioned when it is not installed. `jdkVersion` in
  the root build is now the single place the number appears; detekt's
  `jvmTarget` reads it too.
- detekt to 2.0.0-alpha.6 (`dev.detekt`), the first release that runs on a
  JDK 25 daemon. Mechanical migration: plugin id, task type, two report-format
  keys, four renamed threshold properties in `detekt.yml`, and the
  `UnusedPrivateMember` → `UnusedPrivateFunction` rename. Zero findings on all
  ten subprojects on both JDKs.
- A version guard in `FormalVerificationPluginComponentRegistrar`: compare the
  running compiler's feature release against the one the plugin was built
  against, and report an error naming both instead of letting the linkage
  failure happen. The bound is one-sided — older feature releases still load
  and verify, so they are left alone.
- CI runs the matrix `[21, 25]`.
- README gains a Kotlin/JDK section; docs/developing.md gains a toolchain note.

## Verification

`./agent-scripts/check-all.sh` passes on JDK 21 and on JDK 25.

The verification pipeline was additionally run with the test JVM forced to
JDK 25 rather than the pinned toolchain: 139 tests, 0 failures, Silicon and Z3
included.

End to end against `mavenLocal`, with #261 merged in, using a source with a
known verification failure. Every combination reports the expected Viper
diagnostic:

| Path | Daemon JDK | `jvmToolchain` | Result |
|:--|:--|:--|:--|
| Gradle 9.7.1 | 21 | 21 | diagnostic reported |
| Gradle 9.7.1 | 21 | 25 | diagnostic reported |
| Gradle 9.7.1 | 25 | 21 | diagnostic reported |
| Gradle 9.7.1 | 25 | 25 | diagnostic reported, class file 69 |
| Gradle 8.14.3 | 21 | 25 | diagnostic reported, class file 69 |
| Gradle 8.14.3 | 25 | — | fails in script compilation |

`kotlinc` 2.3.0 reports the same diagnostic on JDK 21 and on JDK 25. `kotlinc`
2.4.10 fails on both, identically — the Kotlin axis and the JDK axis are
independent.

The guard was checked against both: `kotlinc` 2.4.10 now exits 1 with
"this build of SnaKt is compiled against Kotlin 2.3.0 and cannot run on Kotlin
2.4.10", and 2.3.0 is unaffected.

## Two things for the reviewer to weigh

**detekt 2.0.0-alpha.6 is an alpha.** It is the only thing standing between this
branch and a JDK 25 daemon, and detekt 1.23 is a dead line for this purpose.
Config errors under it are hard failures at configure time with the allowed keys
named, so a bad migration cannot pass silently. Worth re-pinning once 2.0.0
stable ships.

**The CI matrix doubles the build job**, verification tests included. Dropping
JDK 25 to the fast checks only, or to a scheduled run, is a reasonable trim.

Unrelated but noticed: `config/detekt/baseline.xml` is referenced by
`build.gradle.kts` and has never existed in the repo. Left alone here.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,4 @@
-import io.gitlab.arturbosch.detekt.Detekt
+import dev.detekt.gradle.Detekt
 import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
@@ -7,7 +7,7 @@ plugins {
     id("com.github.gmazzo.buildconfig") version "5.6.5"
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.16.3" apply false
     id("com.gradle.plugin-publish") version "1.3.1" apply false
-    id("io.gitlab.arturbosch.detekt") version "1.23.7"
+    id("dev.detekt") version "2.0.0-alpha.6"
 }
 
 // The JDK this project compiles against and targets. Pinned rather than
@@ -33,7 +33,7 @@ allprojects {
 }
 
 subprojects {
-    apply(plugin = "io.gitlab.arturbosch.detekt")
+    apply(plugin = "dev.detekt")
 
     detekt {
         buildUponDefaultConfig = true
@@ -59,9 +59,9 @@ subprojects {
         exclude("**/build/**")
         reports {
             html.required.set(true)
-            xml.required.set(true)
+            checkstyle.required.set(true)
             sarif.required.set(false)
-            md.required.set(false)
+            markdown.required.set(false)
         }
     }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import io.gitlab.arturbosch.detekt.Detekt
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
@@ -9,9 +10,20 @@ plugins {
     id("io.gitlab.arturbosch.detekt") version "1.23.7"
 }
 
+// The JDK this project compiles against and targets. Pinned rather than
+// inherited from the launching JVM so the bytecode does not depend on which
+// JDK a developer happens to run Gradle on.
+val jdkVersion = 21
+
 allprojects {
     group = "org.jetbrains.kotlin.formver"
     version = "0.1.0-SNAPSHOT"
+
+    pluginManager.withPlugin("org.jetbrains.kotlin.jvm") {
+        extensions.configure<KotlinJvmProjectExtension> {
+            jvmToolchain(jdkVersion)
+        }
+    }
 
     tasks.withType<KotlinCompile> {
         compilerOptions {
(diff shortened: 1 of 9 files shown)


build.gradle.kts
```diff
@@ -1,18 +1,30 @@
...
     id("com.gradle.plugin-publish") version "1.3.1" apply false
-    id("io.gitlab.arturbosch.detekt") version "1.23.7"
+    id("dev.detekt") version "2.0.0-alpha.6"
 }
 
+// The JDK this project compiles against and targets. Pinned rather than
+// inherited from the launching JVM so the bytecode does not depend on which
+// JDK a developer happens to run Gradle on.
+val jdkVersion = 21
+
```
sachok42, line 17: isn't the goal of this PR to move from JDK 21 to JDK 25?

build.gradle.kts
```diff
@@ -1,18 +1,30 @@
...
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     kotlin("jvm") version "2.3.0" apply false
     id("com.github.gmazzo.buildconfig") version "5.6.5"
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.16.3" apply false
     id("com.gradle.plugin-publish") version "1.3.1" apply false
-    id("io.gitlab.arturbosch.detekt") version "1.23.7"
+    id("dev.detekt") version "2.0.0-alpha.6"
```
sachok42, line 10: What is the reasoning behind using an alpha build?

sachok42 (changes requested):

jesyspa:
Superseded by #269.

PR #265, opened 2026-08-20
Lower a primary constructor call to an allocation and a fold

jesyspa:
A constructor was a bodyless Viper method whose postcondition promised `acc(C_unique(ret), write)` and nothing about what the predicate contains. Silicon gives such an inhaled predicate an unconstrained snapshot, so a heap-dependent function reading through a nested predicate is unrelated to the values the constructor was given. `length(new) == old(length(tail)) + 1` on a linked-list push was unprovable — not because it is false, but because the contract could not express it.

A primary constructor call now builds the object at the call site: allocate with Silver's `NewStmt`, whose reference is genuinely fresh; write the parameters that back properties; fold the class predicate, which ties its snapshot to the heap just written.

`pure_recursion` verifies clean as a result — `pure_recursion.viper.diag.txt` regenerated empty and was removed.

## Shape

```
obj := new(<every backing field of C and its supertypes>)
inhale isSubtype(typeOf(obj), C())
obj.bf_x := arg          // parameter backing a `var`
inhale g_y(obj) == arg   // parameter backing a `val` (permission-free accessor)
<residual assumptions>
fold acc(S_unique(obj), write)   // each supertype, innermost first
fold acc(C_unique(obj), write)
```

## What this assumes

Folding forces the predicate body to be *proven*, where the opaque method merely assumed it. Two conjuncts no argument establishes are inhaled so the fold goes through:

- the type of a backing field no parameter writes, since a Kotlin initializer in the class body is not converted today;
- the unique predicate of a `@Unique` property in the same position — emitted, but no test exercises it.

`con_C` assumed all of this wholesale in its postcondition, so this assumes no more than the call it replaces.

## Notes for review

- **Specification constructors are exempt.** `UniquePred(x)` is itself a primary constructor call, and `fold`/`unfold` pattern-match the method call it converts to. Constructors carrying `@SpecificationHelper` keep the call form.
- **The diamond's shared predicate is folded once per path, and that is required, not incidental.** `Diamond_unique` nests `LeftBranch_unique` and `RightBranch_unique`; each nests its own `Apex_unique`. Two instances are consumed, so one fold would leave the second branch short. It succeeds because a shared apex is necessarily an interface, and an interface predicate carries no permission — `Apex_unique(this)` is `isSubtype(typeOf(this), Apex())` alone. `construction_through_hierarchy.kt` renders these bodies so the claim is checkable by reading the golden.
- **Superclass constructors are still not run.** `B(fieldB) : A(C())` allocates and folds `A_unique` without evaluating `A(C())`, so `g_fieldNotOverride(b)` stays unconstrained. Identical to the opaque path — no regression — but the gap is now visible in the emitted Viper instead of hidden behind an assumed postcondition.
- **`con_C` is still emitted** although nothing calls it now. Removing it is a separate change.
- Most of the golden churn is anon-var renumbering: the block's result is a variable, so callers emit one extra hop.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/FirUtils.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/FirUtils.kt
@@ -105,9 +105,12 @@ val FirBasedSymbol<*>.isBorrowed: Boolean
  *
  * Such a function is never executed, so its arguments are neither consumed nor mutated.
  */
+fun FirFunctionSymbol<*>.isSpecificationFunction(session: FirSession): Boolean =
+    hasAnnotation(annotationId("SpecificationHelper"), session)
+
 context(context: CheckerContext)
 private fun FirFunctionSymbol<*>.isSpecificationFunction(): Boolean =
-    hasAnnotation(annotationId("SpecificationHelper"), context.session)
+    isSpecificationFunction(context.session)
 
 context(context: CheckerContext)
 fun FirStatement.isSpecificationCall(): Boolean =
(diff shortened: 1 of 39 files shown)


jesyspa:
Superseded by #281.

PR #266, opened 2026-08-20
Make recursive @Pure functions over @Unique structures verify

jesyspa:
Stacked on #260. Two changes that together let a recursive `@Pure` function over a uniquely-owned recursive structure reach Viper and verify.

## Unfold the framing predicate for a `@Unique` property read

A `val @Unique` property is embedded as a permission-free function, so `Link_unique(this)` nests `acc(Link_unique(next(this)))` behind a null guard. Passing `l.next` to a parameter that needs that predicate therefore requires an `unfolding`, which was never emitted — so a recursive pure function failed its own recursive call with "insufficient permission to access `Link_unique(next(l))`".

`visitFunctionCall` now wraps the emitted `FuncApp` in `Exp.Unfolding` for each `@Unique` formal whose argument is a `@Unique` property read, innermost receiver outermost — the only sound order, since unfolding `A_unique(a)` is what grants `acc(B_unique(b(a)))`. The `unfolding` lands inside whatever branch the user's own null check already guards, so no synthesised guards are needed and failures surface against the user's code.

Not handled: a `@Unique` property declared on a supertype, which needs `hierarchyPathTo` treatment. `PropertyEmbedding` carries no declaring class, so such a call gets no unfolding — exactly as before, no regression.

## Let a function returning `@Unique` have postconditions

Pre-existing bug, found while writing the test. `extractFormverReturnVar` compared the `postconditions` binder's type to the declared return type with `!=`, which includes the uniqueness and locality attributes. The declared type carries `@Unique`; the binder cannot, because `@Unique` on a type argument is rejected as an invalid target. So *any* function returning `@Unique T` failed with an internal error.

The binder denotes the return value, so it now inherits those attributes rather than restating them: types are compared modulo them, and a session component supplies the binder's uniqueness from the signature. That second half is load-bearing and was verified by registering the no-op default and watching `expected 'unique', actual 'shared'` return.

## Soundness note

Modelling a closed `val` as a permission-free uninterpreted function is sound: the path is taken only for `(property final || class final) && no custom accessor`, so the backing field is written once in the constructor and `next(x)` really is time-invariant. Open `val`s and custom getters fall through to impure methods, closing the override-`val`-with-`var` hole. Accessor names are class-qualified, so two unrelated classes with a same-named `val` do not collide. The remaining gap is incompleteness, not unsoundness: a body-declared `val a = 5` gets no constraint tying it to `5`.

## Test

`testData/diagnostics/verification/uniqueness/pure_recursion.kt`. `length` and `callLength` verify. `pushFront` still fails here on a separate constructor-snapshot problem, fixed in #265.

### DIFF
--- a/SPECIFICATIONS.md
+++ b/SPECIFICATIONS.md
@@ -46,6 +46,8 @@ fun abs(x: Int): Int {
 ```
 
 Multiple conditions are implicitly conjoined. The postconditions block receives the return value as its parameter.
+Its type argument spells the return type without `@Unique` or `@Borrowed`, which the parameter inherits from the
+function's signature.
 
 ## Loop Invariants
 
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/FormverFirBlock.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/FormverFirBlock.kt
@@ -6,11 +6,16 @@
 package org.jetbrains.kotlin.formver.core.conversion
 
 import org.jetbrains.kotlin.fir.declarations.FirAnonymousFunction
+import org.jetbrains.kotlin.fir.declarations.FirSimpleFunction
 import org.jetbrains.kotlin.fir.expressions.*
+import org.jetbrains.kotlin.fir.symbols.SymbolInternals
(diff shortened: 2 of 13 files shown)


jesyspa:
Superseded by #280.

PR #267, opened 2026-08-20
Prove termination of recursive @Pure functions over @Unique parameters

jesyspa:
Stacked on #265. Recursive Viper functions are unsound without a termination measure, and SnaKt had no `decreases` machinery at all.

## What lands

`decreases` AST nodes (`DecreasesTuple`, `DecreasesWildcard`) and Viper's well-founded-order domains, emitted into the `Program`. SnaKt builds a Silver AST directly and runs only `[ConsistencyCheck, Verification]`, so `beforeTranslate` never runs and nothing the parse path would `import` is present — the domains have to be emitted. **Five** are load-bearing: `WellFoundedOrder[T]`, the empty `PredicateInstance` domain, `PredicateInstancesNestedRelation`, `PredicateInstancesWellFoundedOrder`, and `BoolWellFoundedOrder`.

The measure is the uniqueness predicate itself, which is the point: a uniquely-owned structure is finite exactly because you hold a finite predicate, and the recursive call sits under the `unfolding` from #266, which is what supplies the nesting fact. `length` gets `decreases l != nullValue(), Link_unique(l)`.

The nullity component is required, not decoration. `decreases Link_unique(l)` alone fails — a predicate-instance term needs permission, and when `next(l)` is null we hold none. Verified both ways, with a negative control: recursing on `l` instead of `next(l)` is rejected.

Bodyless functions get `decreases _`. That assumes nothing: an uninterpreted function cannot recurse. It is required because any function called from a measured one must itself carry a clause, and `length` calls `next` and `plusInts`.

278 `decreases` lines across 45 goldens, no new verification error anywhere.

## Known limitation

A **recursive** measured function calling a **bodied unmeasured** one crashes silver with `ClassCastException: viper.silver.ast.Function cannot be cast to viper.silver.ast.Exp` inside the termination plugin. Nothing in the suite hits it, but a user could write it.

It is not recorded as a test: a crash is not a diagnostic, so regeneration would record the file as verifying clean while the test fails — a golden that lies. Documented in prose instead. Recording it needs the suite to be able to assert on an internal error, which does not exist today.

The fix is to give non-recursive bodied `@Pure` functions `decreases _` as well, on the same reasoning as bodyless ones — no recursion, so nothing is assumed. It must *not* be applied blanket: `user_invariants/factorial.kt` currently records a consistency error for a self-referencing postcondition, and a wildcard there would suppress a check we have. That needs cycle detection over the pure call graph, so it is follow-up rather than part of this PR.


jesyspa:
Superseded by #282.

PR #269, opened 2026-08-29
Build on JDK 25

komiputer:
Allows the build to run under JDK 25 while keeping the compilation toolchain
pinned to JDK 21, so published bytecode remains compatible with the project's
existing target.

- Upgrade the Gradle wrapper to 9.7.1.
- Migrate detekt to `dev.detekt:2.0.0-alpha.6` and its 2.0 configuration names.
- Pin Kotlin compilation and detekt's JVM target to 21.
- Provision the pinned toolchain through the foojay resolver when needed.
- Run CI under JDK 25.

The detekt upgrade is needed because the stable 1.23 line cannot run inside a
JDK 25 Gradle daemon. There is no stable detekt 2.0 release yet.


build.gradle.kts
```diff
@@ -1,18 +1,32 @@
...
 
 plugins {
     kotlin("jvm") version "2.3.0" apply false
     id("com.github.gmazzo.buildconfig") version "5.6.5"
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.16.3" apply false
     id("com.gradle.plugin-publish") version "1.3.1" apply false
-    id("io.gitlab.arturbosch.detekt") version "1.23.7"
+    // 2.0 is still in alpha, but 1.23 embeds a Kotlin compiler that rejects a
+    // JDK 25 daemon, and 1.23 is the end of that line.
+    id("dev.detekt") version "2.0.0-alpha.6"
```
sachok42, line 12: Any reasons to use an alpha build?

build.gradle.kts
```diff
@@ -1,18 +1,32 @@
...
 }
 
+// The bytecode target. Independent of the JDK running Gradle, which may be
+// anything from 21 to 25: pinning it here keeps the published class file
+// version from following whichever JDK a developer launched the build with.
+val jdkVersion = 21
```
sachok42, line 18: Even if it can be anything, isn't the overall goal of this PR moving from JDK 21 to JDK 25?

sachok42 (changes requested):

build.gradle.kts
komiputer: The stable 1.23 line embeds a Kotlin compiler whose Java-version parser rejects a JDK 25 daemon. There is no stable detekt 2.0 release yet, so alpha.6 is currently the smallest viable upgrade. I also trimmed the PR to only the JDK 25 migration.

build.gradle.kts
komiputer: Yes. I removed the JDK 21 compatibility/toolchain work; CI and detekt now target JDK 25 directly.

komiputer:
@sachok42 I cut this down to the JDK 25 build changes, kept the compilation target at JDK 21, and addressed both review points. CI is green. Could you take another look?

sachok42 (approved):

PR #271, opened 2026-08-29
Make the uniqueness checker usable from a Gradle build

komiputer:
Gradle compiles with kotlin-compiler-embeddable, which moves the libraries it bundles
under its own package prefix. Some of those appear in the compiler API the plugin builds
on, so a plugin jar compiled against the plain compiler asks the embeddable one for
classes it does not have. This publishes a second jar with those references rewritten and
points the Gradle plugin at it.

The plugin jar also shaded in the whole plain compiler, which turned the mismatch into a
cast failure rather than a missing class. It no longer does; the running compiler supplies
those classes.

Uniqueness checking resolves locality facts, which only the locality extensions provide,
so the compiler plugin now enables both together — as the test harness already does.

Supersedes #261. Relocation list checked against the 2.3.0 package layouts: `com.intellij`
and `kotlinx.collections.immutable` are the only relocated prefixes the plugin's own
sources reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ syntax are not supported.
 This repository consists of three published parts:
 
 - `formver.compiler-plugin`: a K2 compiler plugin that performs formal verification.
+  It is published twice: the plain artifact runs against `kotlinc`, while
+  `formver.compiler-plugin-embeddable` runs against `kotlin-compiler-embeddable`,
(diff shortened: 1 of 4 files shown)


sachok42 (commented):
I requested some changes on the previous iteration of this PR, they weren't made. Later today will copy them here

PR #272, opened 2026-08-29
Make the severity of failed proofs configurable

komiputer:
Adds `formver { verificationErrorSeverity("error") }` and the matching
`verification_error_severity` plugin option, so a failed proof can fail the
build instead of warning.

The Kotlin diagnostic API fixes a severity per factory, so this needs one
factory per severity: `VerificationErrors` (warnings) and
`StrictVerificationErrors` (errors). Both use the same property names, so a
diagnostic's name — and any `@Suppress` or expected test tag — is the same
either way. A `VerificationDiagnostics` holder with required parameters for
every factory keeps the two sets in step.

### DIFF
--- a/formver.compiler-plugin/cli/src/org/jetbrains/kotlin/formver/cli/FormalVerificationCommandLineProcessor.kt
+++ b/formver.compiler-plugin/cli/src/org/jetbrains/kotlin/formver/cli/FormalVerificationCommandLineProcessor.kt
@@ -15,6 +15,7 @@ import org.jetbrains.kotlin.formver.cli.FormalVerificationConfigurationKeys.DUMP
 import org.jetbrains.kotlin.formver.cli.FormalVerificationConfigurationKeys.ERROR_STYLE
 import org.jetbrains.kotlin.formver.cli.FormalVerificationConfigurationKeys.LOG_LEVEL
 import org.jetbrains.kotlin.formver.cli.FormalVerificationConfigurationKeys.UNSUPPORTED_FEATURE_BEHAVIOUR
+import org.jetbrains.kotlin.formver.cli.FormalVerificationConfigurationKeys.VERIFICATION_ERROR_SEVERITY
 import org.jetbrains.kotlin.formver.cli.FormalVerificationConfigurationKeys.VERIFICATION_TARGETS_SELECTION
 import org.jetbrains.kotlin.formver.common.*
 import org.jetbrains.kotlin.formver.common.FormalVerificationPluginNames.CHECK_LOCALITY_OPTION_NAME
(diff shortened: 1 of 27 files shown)


formver.compiler-plugin/test-fixtures/org/jetbrains/kotlin/formver/plugin/services/ExtensionRegistrarConfigurator.kt
```diff
@@ -117,8 +119,24 @@ object FormVerDirectives : SimpleDirectivesContainer() {
...
+    val FAIL_ON_VERIFICATION_ERROR by directive(
+        description = "Report failed proofs as errors rather than warnings"
+    )
 }
 
+/**
+ * The severity failed proofs are reported at in this module.
+ *
+ * Both the plugin's own reporting and the test harness's verification phase read
+ * it, so that a test exercises the option rather than one of its two callers.
```
sachok42, line 132: you sure that this comment paragraph is worth keeping

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
```diff
@@ -27,6 +28,48 @@ object PluginErrors : KtDiagnosticsContainer() {
...
+ *
+ * The Kotlin diagnostic API fixes a severity per factory, so a configurable
+ * severity takes one set of factories per severity. Both sets use the same
+ * property names, and a factory's name is its property name, so which set is in
+ * use is invisible to anything naming a diagnostic: a `@Suppress`, an
+ * `-Xwarning-level`, a test's expected tags.
+ *
+ * Every parameter is required, so a diagnostic added at one severity does not
+ * compile until it is added at the other.
+ */
```
sachok42, line 42: probably too big of a comment

formver.compiler-plugin/cli/src/org/jetbrains/kotlin/formver/cli/FormalVerificationCommandLineProcessor.kt
```diff
@@ -145,6 +157,12 @@ class FormalVerificationCommandLineProcessor : CommandLineProcessor {
                 DUMP_UNIQUENESS_CFG_OPTION ->
                     configuration.put(DUMP_UNIQUENESS_CFG, value.toBooleanStrict())
 
+                VERIFICATION_ERROR_SEVERITY_OPTION ->
+                    configuration.put(
+                        VERIFICATION_ERROR_SEVERITY,
+                        VerificationErrorSeverity.valueOf(value.toUpperCaseAsciiOnly())
+                    )
```
sachok42, line 164: valueOf throws IllegalArgumentException on any invalid value ("warning ", in example). A try/catch structure should be implemented that would create message like "Unknown severity level: 'warning ' ". Not critical but would significantly improve user experience

formver.compiler-plugin/plugin/src/org/jetbrains/kotlin/formver/plugin/compiler/PluginErrors.kt
```diff
@@ -45,3 +88,14 @@ object VerificationErrors : KtDiagnosticsContainer() {
         CONSISTENCY.name
     )
 }
+
+// Not registered as a diagnostic container: its renderers live in the factory
+// VerificationErrors registers, which is enough for them to be found.
```
sachok42, line 93: I think there should be a test that checks that this is not registered, as even though it's not registered right now, someone could accidentally register it, as everything else that looks like this is registered

sachok42 (changes requested):
One UX problem, one test lacking, some minor commentary issues

PR #273, opened 2026-08-29
Pass an argument for every parameter of a call

komiputer:
A call omitting a defaulted argument emitted fewer actuals than the callee has
formals. Silicon leaves the formal unbound, the state after the call is
inconsistent, and everything downstream is proved vacuously — `a.startsWith(b)`
made `verify(false)` succeed.

Omitted parameters are now filled with a value constrained only by the
parameter's type. Whether a callee takes one argument per parameter is a member
of `CallableEmbedding`, so a new implementation must answer it; all arguments of
a call are built in one place, vararg handling included. Named arguments out of
order now reach the parameters they name.

Takes over #234 (rebased, review comments addressed). Tests cover reference,
nullable and function-typed defaults, non-inline and inline callees, and a
vararg parameter alongside an omitted default.

### DIFF
--- a/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
+++ b/formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
@@ -23,6 +23,7 @@ import org.jetbrains.kotlin.formver.common.SnaktInternalException
 import org.jetbrains.kotlin.formver.common.UnsupportedFeatureBehaviour
 import org.jetbrains.kotlin.formver.core.embeddings.LabelLink
 import org.jetbrains.kotlin.formver.core.embeddings.callables.CallableEmbedding
+import org.jetbrains.kotlin.formver.core.embeddings.callables.FunctionSignature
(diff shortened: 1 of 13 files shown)


formver.compiler-plugin/core/src/org/jetbrains/kotlin/formver/core/conversion/StmtConversionVisitor.kt
```diff
@@ -264,34 +265,79 @@ object StmtConversionVisitor : FirVisitor<ExpEmbedding, StmtConversionContext>()
...
+     * Converts [arguments] into the actuals of a call to [callee].
+     *
+     * A `null` entry is a formal argument the call leaves to its default value, which requires
+     * [arguments] to be in the order of the callee's formal arguments. The default value is not
+     * available at the call site, and a Viper call that omits the actual leaves the callee's formal
+     * unbound, which makes the verification state inconsistent and proves anything afterwards. Such
+     * a formal gets a value of its own type that nothing else is assumed about: that loses what the
+     * default would have told us, but only that. The declarations of those values are collected in
+     * [declarations], and belong in front of the call.
+     */
```
sachok42, line 278: giant over-extensive comment that needs to get smaller

sachok42 (changes requested):

PR #275, opened 2026-08-29
Test agent script helper behavior

jesyspa:
## Summary

- exercise `gradle_filter` and golden-assertion classification in the no-build script test suite
- distinguish `.kt` source names from generated test method names
- fix filtering for source files whose names begin with `test`

## Testing

- `agent-scripts/tests/run.sh`
- `./agent-scripts/check-all.sh`

## Friction observed

- The repository has few isolated unit-test targets; most compiler behavior uses generated golden tests.
- `check-all.sh` is comprehensive but took about 3m38s for this shell-only change.
- Running ShellCheck directly against all agent scripts reports pre-existing findings and lacks a configured project command; the repository's actual pre-commit hooks passed.

### DIFF
--- a/agent-scripts/lib.sh
+++ b/agent-scripts/lib.sh
@@ -25,9 +25,12 @@ need_python3() {
 # and --tests is case-sensitive. A name already in method form, and a path to
 # the .kt, are both accepted.
 gradle_filter() {
-    local pattern="${1##*/}"
-    pattern="${pattern%.kt}"
-    if [[ "$pattern" == test* ]]; then
+    local pattern="${1##*/}" source_file=0
+    if [[ "$pattern" == *.kt ]]; then
+        pattern="${pattern%.kt}"
+        source_file=1
+    fi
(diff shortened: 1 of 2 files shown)


agent-scripts/lib.sh
```diff
@@ -25,9 +25,12 @@ need_python3() {
...
 gradle_filter() {
-    local pattern="${1##*/}"
-    pattern="${pattern%.kt}"
-    if [[ "$pattern" == test* ]]; then
+    local pattern="${1##*/}" source_file=0
+    if [[ "$pattern" == *.kt ]]; then
+        pattern="${pattern%.kt}"
+        source_file=1
+    fi
+    if [[ "$source_file" -eq 0 && "$pattern" == test* ]]; then
```
sachok42, line 33: this fix wouldn't work on files without .kt extension. While they aren't used right now, we should either support them or specify in readme that we don't

sachok42 (changes requested):
one issues that I'm not sure on our attitude to, other than that -- a fine fix

PR #278, opened 2026-08-29
Fix loop-carried uniqueness state over recursive structures

komiputer:
Unifies #259 and #262: the test suites and the fixes that make them run.

Two fixes:
- The assignment transfer function projected the RHS substate into the LHS path before moving the paths the RHS read, so `current = current.next` marked the wrong node's `next`. The RHS is now read against the incoming state and moved first.
- The fixed point did not terminate over recursive types. A `PathTrie` path is now collapsed wherever a symbol comes round a second time, giving a summary node whose data is the join of everything below it. Operations at a summary join rather than replace, so the approximation only over-reports.

Also adds linked list and binary tree suites and asserts the uniqueness diagnostics' inline markers, which `tagsToConsider` did not cover. `pivot?.right = root` is still not recognised as restoring `pivot.right` — safe-call desugaring hides the path; documented in the test file.

### DIFF
--- a/formver.common/src/org/jetbrains/kotlin/formver/common/services/TagCollector.kt
+++ b/formver.common/src/org/jetbrains/kotlin/formver/common/services/TagCollector.kt
@@ -120,6 +120,15 @@ abstract class TagCollector(
     }
 
     /**
+     * The expected side of a filtered comparison comes from parsing markers back out of the test file, and a marker
+     * names a tag and a range and nothing else: it cannot say that one diagnostic fired twice at one position. The
+     * reported side is matched to that by collapsing such repeats. The message of each individual report is still
+     * checked against the `.diag.txt` golden.
+     */
+    private fun deduplicatedReportedInfos(): Map<TestFile, List<CodeMetaInfo>> =
+        reportedInfos.mapValues { (_, infos) -> infos.distinctBy { Triple(it.tag, it.start, it.end) } }
+
+    /**
      * Renders the expected output but ignores tags that are not considered for comparison
      */
     private fun expectedFileFilteredForTags(): String {
@@ -137,7 +146,7 @@ abstract class TagCollector(
      */
     fun assertFileEqualFilteredForTags() {
         val expectedOutput = expectedFileFilteredForTags()
-        val actualOutput = renderText(reportedInfos, testServices.sourceFileProvider::getContentOfSourceFile)
+        val actualOutput = renderText(deduplicatedReportedInfos(), testServices.sourceFileProvider::getContentOfSourceFile)
         testServices.assertions.assertEquals(expectedOutput, actualOutput) {
             "Actual tags differ from golden file"
         }
--- a/formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
+++ b/formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
@@ -97,12 +97,24 @@ fun AccessState.append(other: AccessState): AccessState {
  * If any of the intermediate path components is not resolved within [uniquenessState], the uniqueness of those
  * components is automatically inferred to be the join between the declared uniqueness of the component's symbol and the
  * uniqueness of the parent.
+ *
+ * An access that reaches past a summary node lands on the summary itself, which is the only node standing for the path
+ * it names. The transform there applies to the whole region the summary covers rather than to one path in it, so its
+ * result is joined with the value already held rather than replacing it: a transform that raises the value, such as a
+ * move, still takes effect, while one that lowers it, such as restoring a borrow, cannot clear a fact that holds of
+ * some other path in the region.
  */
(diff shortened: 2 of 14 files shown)


formver.common/src/org/jetbrains/kotlin/formver/common/services/TagCollector.kt
```diff
@@ -119,6 +119,15 @@ abstract class TagCollector(
         )
     }
 
+    /**
+     * The expected side of a filtered comparison comes from parsing markers back out of the test file, and a marker
+     * names a tag and a range and nothing else: it cannot say that one diagnostic fired twice at one position. The
+     * reported side is matched to that by collapsing such repeats. The message of each individual report is still
+     * checked against the `.diag.txt` golden.
+     */
+    private fun deduplicatedReportedInfos(): Map<TestFile, List<CodeMetaInfo>> =
```
sachok42, line 128: function names should preferably start with verbs. maybe ЭgetDeduplicatedReportedInfos"?

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/AccessState.kt
```diff
@@ -97,12 +97,24 @@ fun AccessState.append(other: AccessState): AccessState {
  * If any of the intermediate path components is not resolved within [uniquenessState], the uniqueness of those
  * components is automatically inferred to be the join between the declared uniqueness of the component's symbol and the
  * uniqueness of the parent.
+ *
+ * An access that reaches past a summary node lands on the summary itself, which is the only node standing for the path
+ * it names. The transform there applies to the whole region the summary covers rather than to one path in it, so its
+ * result is joined with the value already held rather than replacing it: a transform that raises the value, such as a
+ * move, still takes effect, while one that lowers it, such as restoring a borrow, cannot clear a fact that holds of
+ * some other path in the region.
```
sachok42, line 105: This comment is over-explanatory

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessErrors.kt
```diff
@@ -33,4 +33,16 @@ object UniquenessErrors : KtDiagnosticsContainer() {
     val INVALID_UNIQUENESS_TYPE_TARGET by error0<PsiElement>()
 
     override fun getRendererFactory() = UniquenessErrorMessages
+
+    fun tags() = listOf(
```
sachok42, line 37: why is it a functions, not a variable? Certainly looks like one semantically

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/GraphUniquenessStatesAnalyzer.kt
```diff
@@ -142,20 +151,25 @@ class GraphUniquenessStatesAnalyzer(
...
-                var newUniquenessState = data.getOrInitialize()
+                val uniquenessState = data.getOrInitialize()
                 val leftAccessPaths = leftAccessState.enumeratePaths()
                 val rightAccessState = rightValue.resolveAccessState()
 
+                // The right-hand side is read before the assignment takes effect, so both what it projects and the move
+                // of the paths it reads are computed against the incoming state, and the move is applied first. That
+                // ordering matters when the left-hand side is a prefix of what was read: advancing a cursor with
+                // `current = current.next` moves the field out of the node the cursor is leaving, and overwriting
+                // `current` then drops that mark, because it described a location the cursor no longer reaches.
```
sachok42, line 162: Same over-extensive commentary as was seen above

formver.compiler-plugin/uniqueness/src/org/jetbrains/kotlin/formver/uniqueness/plugin/UniquenessState.kt
```diff
@@ -28,13 +28,53 @@ fun UniquenessState.enumerateInconsistentPaths(): Sequence<Path> =
...
+/**
+ * Collapses every path that revisits a symbol into a summary of the region below the repeat.
+ *
+ * Assignment projects the substate of the right-hand side under the path of the left-hand side, so a loop over a
+ * recursive type can deepen the trie once per iteration - `prev.next`, then `prev.next.next`, and so on - and the
+ * chain of states never stabilizes. Cutting the trie wherever a symbol comes round a second time bounds its depth by
+ * the number of distinct symbols the function mentions, which makes the state space finite and the fixed point
+ * reachable. Any growth that is unbounded has to repeat a symbol, because a loop body can only ever append the finitely
+ * many components its own syntax spells out; paths over non-recursive data are therefore left exactly as they are.
+ */
```
sachok42, line 56: again, a giant over-extensive comment

sachok42 (changes requested):

PR #577, opened 2026-09-18
Diagnose unsupported mutual pure recursion, naming the cycle

### LINKED ISSUE #373: Mutually recursive @Pure functions produce misleading return-type failures
## Minimal reproducer

```kotlin
// FULL_JDK
import org.jetbrains.kotlin.formver.plugin.*

@Pure
fun even(n: Int): Boolean {
    preconditions { n >= 0 }
    return if (n == 0) true else odd(n - 1)
}

@Pure
fun odd(n: Int): Boolean {
    preconditions { n >= 0 }
    return if (n == 0) false else even(n - 1)
}
```

## Expected behavior

The mutually recursive functions decrease their non-negative integer argument on every edge of the cycle. They should either verify, like the equivalent directly recursive `@Pure` countdown does, or receive a specific diagnostic explaining that mutual recursion lacks a supported termination measure.

## Observed behavior

Conversion succeeds and emits both Viper functions with `ensures isSubtype(typeOf(result), boolType())`. Verification then reports the following diagnostic twice for `even` and marks both Kotlin declarations with `VIPER_VERIFICATION_ERROR`:

```text
Postcondition of even might not hold. Assertion isSubtype(typeOf(result), boolType()) might not hold.
```

There is no recursion or well-foundedness diagnostic, and verification completes without a timeout. Direct recursive `@Pure` countdown over the same decreasing integer shape verifies in the same focused test. Mutually recursive non-pure methods with the same shape also verify.

## Environment

- SnaKt revision: `9bac7b389dc5d2e01ffe14b0f92b69f134866c87`
- Base branch: `implementing-air-automations`
- Java: Temurin 21.0.12.1
- Z3: 4.8.7
- Reproduced with: `./agent-scripts/test.sh --verify recursion_variants`
- Assignment: #350

Produced by Air Automations. Name: Testing swarm agent / Run: https://air.jetbrains.cloud/org/05cf1a7f-6ab5-713b-abd3-29d0c8a05e2d/automations/34ccbbdd-3fd2-474e-8f0a-10b4b2d5bda4?run=9e58815c-44c5-42a6-891c-2b3eeed1ca41

komiputer:
Supersedes #568 with two changes on top of its commit.

The diagnostic now names the pure functions that form the recursion cycle, so a function that only calls into the cycle sees which functions are responsible:

    Verification depends on mutually recursive pure functions 'even', 'odd', which is not supported

The explanatory comment now describes the actual limitation. Viper accepts mutually recursive functions; Silicon checks each function's postcondition without assuming the postconditions of the others in its cycle, so even the generated result-type postcondition fails with an error that never mentions recursion. The previous comment attributed this to a consistency check, which is not where the failure comes from.

The test gains a function that only calls the directly recursive countdown, showing it still verifies. Cycle detection now walks the converted pure bodies directly instead of copying them out of the resolver.

Validation: `./agent-scripts/test.sh --update-goldens mutual_recursion` produced the committed golden. `./agent-scripts/check-all.sh` completed every Gradle check task, including the compiler-plugin tests, without a failure, but the host killed the run for memory before it printed its summary, twice. Pre-commit and testData checks passed at commit time.

Fixes #373


sachok42:
Consolidated into #580 — *Report proper exceptions instead of internal errors*, which gathers the airbot/komiputer PRs that replace internal errors with specific diagnostics into a single branch. Both commits from this PR (the mutual-recursion diagnostic and the cycle-naming follow-up) were cherry-picked there verbatim and the golden is unchanged. Closing in favor of #580.

PR #580, opened 2026-09-23
Report proper exceptions instead of internal errors (consolidated)

sachok42:
## Consolidated: report proper exceptions instead of internal errors

This PR gathers the airbot/komiputer bugfix PRs that **do not add any verification feature** — each one simply replaces a plugin *internal error* (a crash, a nondeterministic `INTERNAL_ERROR`, or a silent success) with a specific, source-positioned diagnostic. Consolidated into one branch so the overlapping edits to `ProgramConverter.kt`, `StmtConversionVisitor.kt` and the diagnostic containers are resolved once.

Each source PR is preserved as its own commit(s); a final `Reconcile consolidated exception diagnostics` commit fixes the cross-PR interactions (see below).

### New / changed exceptions & diagnostics

| Source PR | Trigger (previously) | Now reports |
|-----------|----------------------|-------------|
| **#577** (supersedes #568) | Mutually recursive pure functions reached Silicon and failed with a misleading result-type error | New `ConversionErrors.MUTUAL_RECURSION_UNSUPPORTED`: *"Verification depends on mutually recursive pure functions 'even', 'odd', which is not supported"* (names the cycle) |
| **#557** | Impure top-level/member call in a pre/postcondition reached the expression-only pure linearizer → `INTERNAL_ERROR` | `ConversionErrors.PURITY_VIOLATION`: *"Precondition/Postcondition contains an impure function or method call"* |
| **#573** | Impure call stored in a local initializer bypassed purity checking and crashed pure conversion | `ConversionErrors.PURITY_VIOLATION`: *"Impure function body detected in pure function"* (initializer is now traversed) |
| **#553** | FIR recovery nodes in already-rejected source were converted → phase-dependent `INTERNAL_ERROR` | No plugin diagnostic — declarations containing FIR error nodes are skipped so the frontend's real diagnostic stays authoritative |
| **#567** | Unsupported callable references fell through to generic FIR rendering with a per-run JVM identity hash | Stable message via a dedicated visitor: *"Function references are not supported (::increment)"* |
| **#565** | Unmodeled JVM marker supertypes (e.g. `java.io.Serializable` on arrays) failed conversion | Non-`FirRegularClassSymbol` supertypes are skipped during class-hierarchy registration |
| **#559** | Silicon backend aborts/timeouts forwarded no error → phased test & compiler result looked *successful* | New `viper.errors.BackendError`; non-proof/non-consistency failures are reported as `INTERNAL_ERROR` |
| **#430** (exception commit only) | `handleUnimplementedElement` threw `SnaktInternalException` → *"An internal error has occurred… Please report this"* | New `common.SnaktUnsupportedFeatureException` + `PluginErrors.UNSUPPORTED_FEATURE`: *"Formal verification does not support this Kotlin feature: {0}"* (e.g. destructuring / `FirComponentCall`) |

### Cross-PR reconciliation (final commit)
Combining the branches produced interactions that none of them saw in isolation:
- **Callable references now report `UNSUPPORTED_FEATURE`** (#430) instead of `INTERNAL_ERROR` (#567's base) — golden `lambdas` updated.
- **`FirComponentCall` gets a stable diagnostic** (originally from #430's probe commit) so the destructuring `UNSUPPORTED_FEATURE` golden no longer embeds a per-run FIR identity hash — same fix class as #567.
- **`exists_list_get_crash`**: impure list access in a postcondition now also emits the contract-level `PURITY_VIOLATION` from #557.

### Not included
Only #430's `UNSUPPORTED_FEATURE` commit was taken; its data-class *probe* test data was left behind. #568 is dropped in favor of #577 (same base commit). Feature/behavior PRs (#579, #578, #571, #570, #569, #566, #564, #562, #561, #560, #558, #556, #555, #554, #552, #551) are out of scope.

### Validation
- `./agent-scripts/test.sh` — 175 passed, 0 failed
- `./agent-scripts/check-all.sh --rerun` — see CI (the plain `check-all.sh` hits an empty-array `set -u` bug under macOS bash 3.2; `--rerun` sidesteps it)


sachok42:
Full pre-push gate passed locally:

```
Summary:
  gradle check:      passed   (incl. :formver.compiler-plugin:test full verification + viper:test)
  check-testdata.sh: passed
  pre-commit:        passed
BUILD SUCCESSFUL in 8m 43s
```

`./agent-scripts/test.sh` (fast conversion loop): 175 passed, 0 failed.

